PR #2, opened 2017-03-30
EL-2468 - Removing Elements

ashley-hunter:
Removing uses of Elements


ashley-hunter:
Removing uses from codepen example and snippet

PR #6, opened 2017-03-31
S 55/el 2453

donna-marijne:
Added scroll tracking in the side navigation and the URL.
Added height to the side navigation and overflow for scrolling. Could not get the custom scrollbar to work.
Updated the style of the side navigation to match Elements.
Added tracking of snippet rendering to allow navigation scrolling when this is complete.
Added an alternate method of loading snippets including pre-rendered code example snippets via prismjs-loader. This can be used by having a component extend BaseDocumentationSection.
Added BaseDocumentationSection to Contacts section.
Changed fragment IDs to use snake-case derived from the component name.
Updated internal documentation links to use the new format.


donna-marijne:
Note that npm install will be required for new dependencies.

package.json
```diff
@@ -26,6 +27,7 @@
     "jquery-ui-dist": "1.12.1",
     "ng2-bootstrap": "^1.4.2",
     "prismjs": "^1.6.0",
+    "prismjs-loader": "0.0.4",
```
ashley-hunter: Duplicate dependency, should probably only be a devDependency

donna-marijne:
Makes sense - removed.

PR #13, opened 2017-04-06
S 55/el 2381

donna-marijne:
https://jira.autonomy.com/browse/EL-2381

Also made some changes to the side navigation.


docs/app/components/side-navigation/side-navigation.component.less
```diff
@@ -53,4 +54,25 @@
             }
         }
     }
+}
+
+// Just for fun
+::-webkit-scrollbar {
```
ashley-hunter: We need more of this... :+1:

PR #16, opened 2017-04-06
S 55/el 2406

donna-marijne:
Added support for the HPE theme files and ability to load them into the doc site for testing. fonts.less is replaced by the Metric font definitions.

Variables file will be hosted at https://github.hpe.com/caf/ux-aspects-hpe

https://jira.autonomy.com/browse/EL-2406


donna-marijne:
Resolved conflict.

PR #29, opened 2017-04-10
S 55/el 2459

donna-marijne:
Added Getting Started details.
Update version and links on landing page/footer.

https://jira.autonomy.com/browse/EL-2459

### DIFF
--- a/docs/app/pages/getting-started/snippets/angular4.txt
+++ b/docs/app/pages/getting-started/snippets/angular4.txt
@@ -1,0 +1,9 @@
+import { CheckboxComponent } from 'ux-aspects';
+
+@NgModule({
+    imports: [
+        CheckboxComponent,
+        // ...
+    ]
+})
+export class AppModule() {}
--- a/docs/app/pages/getting-started/snippets/angular4.txt
+++ b/docs/app/pages/getting-started/snippets/angular4.txt
@@ -1,8 +1,8 @@
-import { CheckboxComponent } from 'ux-aspects';
+import { CheckboxModule } from 'ux-aspects';
 
 @NgModule({
     imports: [
-        CheckboxComponent,
+        CheckboxModule,
         // ...
     ]
 })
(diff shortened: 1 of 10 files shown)


donna-marijne:
I could use some feedback on the "Using UX Aspects" section. It would be good to have some quick summaries of each scenario, but I'm not sure if that would be too complicated for a getting started page. Also for the case of using UX Aspects in an AngularJS application, do we always recommend setting up typescript and webpack? Would we have to use import statements for both the normal and the ng1 script files? I'm not sure how that will work.

docs/app/pages/landing/landing.component.html
```diff
@@ -10,10 +10,10 @@ <h1 class="landing-page-overview-title">{{ landingPage.overview.title }}</h1>
         <uxd-landing-page-feature *ngFor="let feature of landingPage.overview.features" [header]="feature.title" [description]="feature.description" [image]="feature.image"></uxd-landing-page-feature>
     </uxd-landing-page-feature-list>
 
-    <hr class="landing-page-divider">
+    <!--<hr class="landing-page-divider">
```
ashley-hunter: Can we remove these commented out lines?

docs/app/pages/getting-started/snippets/angular4.txt
```diff
@@ -0,0 +1,9 @@
+import { CheckboxComponent } from 'ux-aspects';
```
ashley-hunter: I have made some changes (currently in review) which export the module of each component so update this to: CheckboxModule rather than CheckboxComponent

ashley-hunter:
Content looks good, in regards to your comments:

`do we always recommend setting up typescript and webpack?`
- Yes, it may be good to mention that these are the recommended options, I dont it it will be possible to give instructions for others as there are just too many alternatives.

`Would we have to use import statements for both the normal and the ng1 script files?`

If using in an Angular 1 only app then they can just reference the Angular 1 script as before (although this will not really be relevant once we migrate to Angular 4 completely)

If using a hybrid app then they will need to import both scripts.

donna-marijne:
Made the above changes. I was thinking that maybe we should have some kind of standard documentation template for Angular 4 components, listing the module name and the selector along with the title.

PR #43, opened 2017-04-12
Documentation Lazy Loading

ashley-hunter:
Scripts for pages are now only loaded when they are required rather than up front. I have added a loading spinner whenever page navigation begins so it doesn't look like the page has just frozen.

Split each page (charts, components & css) into it's own module. Webpack will then split these out into separate JS files at build time. Also create a module containing all the wrappers as well.

There is more work we can do to improve build output size which I will work on later today but I want to get this in so we can detect any issues that may have arisen early.

NPM install will be required.


donna-marijne:
Is the a missing npm dependency?
Loading "Gruntfile.js" tasks...ERROR
>> Error: Cannot find module 'optimize-css-assets-webpack-plugin'

ashley-hunter:
Added in missing dependency

ashley-hunter:
Thats the plan!  So page loading should get a lot quicker after that. It will take a bit longer to make the side navigation work correctly lazy load each category but hopefully that will all be completed as part of EL-2490

PR #45, opened 2017-04-12
EL-2380 - Input controls qa changes

AlastairMcKee:


docs/app/services/codepen/codepen.service.ts
```diff
@@ -167,6 +167,8 @@ export class CodePenService {
         'https://cdnjs.cloudflare.com/ajax/libs/chance/0.8.0/chance.min.js',
         'https://cdnjs.cloudflare.com/ajax/libs/angular.js/1.3.6/angular.js',
         'https://cdnjs.cloudflare.com/ajax/libs/angular-ui-bootstrap/0.13.0/ui-bootstrap-tpls.js',
+        'https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.1/moment-with-locales.min.js',
```
ashley-hunter: Use https://cdnjs.cloudflare.com/ajax/libs/moment.js/2.18.1/moment.min.js instead of with locales. It is significantly smaller and faster

configs/webpack.docs.dev.config.js
```diff
@@ -147,7 +147,8 @@ module.exports = {
         ),
 
         new HtmlWebpackPlugin({
-            template: './docs/index.html'
+            template: './docs/index.html',
```
ashley-hunter: Remove the favicon from index.html as it now gets set here automatically and we have duplicate tags

AlastairMcKee:
made changes

PR #46, opened 2017-04-12
Build Output Optimization

ashley-hunter:
Removing use of prism loader in some of our sections. While we will likely move back to use this once each category has it's own module, for the time removing these from these sections reduces build output ~1.3mb.

Also remove uglify step from grunt task for Angular 4 library as this is no longer needed - the consuming module loader should take care of this now

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -25,7 +25,7 @@ module.exports = function (grunt) {
     grunt.registerTask('scripts', ['execute:iconset']);
     grunt.registerTask('assets', ['copy:fonts', 'copy:images', 'copy:less', 'copy:ng1', 'copy:styles']);
     grunt.registerTask('iconset', ['webfont:iconset']);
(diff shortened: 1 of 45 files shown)


ashley-hunter:
I didn't do this as I haven't removed all occurrences of the use of prism loader. I analysed the webpack bundle and only removed the ones that were having a noticeably negative effect on build output size.

ashley-hunter:
I will look into this to see if disabling on all sections brings much of an further optimization

donna-marijne:
You could see what effect it has on navigation, this is the main reason for precompiling. Since a lot of sections already load asynchronously then it probably won't be much more wonky just to disable it site-wide.

The other option is to fix height on the code sections. Potentially we could do a line count of the snippets during build and generate CSS to fix the height for each.

Funnily enough I noticed that Chrome introduced scroll anchoring recently, which deals with exactly this problem.

ashley-hunter:
Closing, opening another ticket for changes

PR #57, opened 2017-04-13
EL-2487

donna-marijne:
Updated getting started and first changelog entry.

https://jira.autonomy.com/browse/EL-2487

### DIFF
--- a/docs/app/pages/getting-started/getting-started.component.html
+++ b/docs/app/pages/getting-started/getting-started.component.html
@@ -9,12 +9,12 @@
 
     <ol>
         <li>Install <a href="https://nodejs.org/">Node.js</a>, which provides the npm package manager.</li>
-        <li>Install <a href="https://angular.io/">Angular 4</a> and <a href="https://www.typescriptlang.org/">TypeScript</a>:
+        <li>Install <a href="https://www.typescriptlang.org/">TypeScript</a>: <kbd>npm install typescript --save-dev</kbd></li>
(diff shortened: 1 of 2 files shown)


docs/app/pages/getting-started/getting-started.component.html
```diff
@@ -9,6 +9,14 @@
 
     <ol>
         <li>Install <a href="https://nodejs.org/">Node.js</a>, which provides the npm package manager.</li>
+        <li>Install <a href="https://angular.io/">Angular 4</a> and <a href="https://www.typescriptlang.org/">TypeScript</a>:
+            <div>
+                <kbd>npm install @angular/common@latest @angular/compiler@latest @angular/compiler-cli@latest
```
ashley-hunter: We should be able to get away without having to specify "@latest" on each as it should install the latest regardless? Unless there was reason for having to do this?

Also typescript is a developer dependency so should be installed using --save-dev rather than --save.

donna-marijne: No, I just lifted that command from the Angular blog. It sure will be nice to have npm sort out dependencies instead of this!

ashley-hunter:
Changes look good!

docs/app/pages/getting-started/getting-started.component.html
ashley-hunter: Couldn't agree with you more!

PR #71, opened 2017-04-19
EL-2503 - Adding ability to import modules to plunker

AlastairMcKee:


docs/app/interfaces/IPlunk.ts
```diff
@@ -1,3 +1,13 @@
 export interface IPlunk {
-    [key: string]: string;
-}
+    files: {
+        [key: string]: string;
+    };
+    modules?: {
+        imports?: string[];
+        library?: string;
+        map?: {
```
ashley-hunter: We should probably make this an array as we may want to provide multiple mappings

docs/app/services/plunker/templates/main_ts.txt
```diff
@@ -1,11 +1,11 @@
 import { NgModule } from '@angular/core';
 import { BrowserModule } from '@angular/platform-browser';
 import { platformBrowserDynamic } from '@angular/platform-browser-dynamic';
-import { CheckboxModule, ToggleSwitchModule } from 'ux-aspects';
 import { AppComponent } from './app';
+${imports}
 
 @NgModule({
-    imports: [BrowserModule, CheckboxModule, ToggleSwitchModule],
+    imports: [BrowserModule, ${modules}],
```
ashley-hunter: If there are no modules imported then we are going to be left with `[BrowserModule, ]`. The comma should only appear when we add modules as IE can sometimes mistake this for an extra array element

AlastairMcKee:
Made requested changes

docs/app/services/plunker/plunker.service.ts
```diff
@@ -27,18 +30,34 @@ export class PlunkerService {
 
     private initForm(title: string, plunk: IPlunk): HTMLFormElement {
 
+        let modules = (plunk.modules && plunk.modules.imports) ? plunk.modules.imports: '';
```
ashley-hunter: TSLint fails

docs/app/interfaces/IPlunk.ts
```diff
@@ -1,3 +1,13 @@
 export interface IPlunk {
```
ashley-hunter: Looking at this again, we may need to change it to something like this:

```typescript
export interface IPlunk {
    files: {
        [key: string]: string;
    };
    modules?: {
        imports?: string[];
        library?: string;
    }[];
    mappings?: {
        alias: string;
        source: string;
    }[];
}
```

As we may want to import things from multiple libraries and the current interface doesn't allow for this

docs/app/services/plunker/plunker.service.ts
```diff
@@ -27,18 +30,34 @@ export class PlunkerService {
...
+        let library = (plunk.modules && plunk.modules.library) ? plunk.modules.library: '';
+        let imports = (modules && library) ? 'import { ' + modules + ' } from \'' + library + '\';': '';
+
+        let map = ''
+        if(plunk.modules && plunk.modules.map) {
+            for(let i = 0; i<plunk.modules.map.length; i++) {
```
ashley-hunter: Although this will have to change a bit when you make the change to the interface, this whole section can be simplified to something like this:

```typescript
plunk.modules.map.map(mapping => `'${ mapping.alias }': '${ mapping.source}'`).join(',\n\t\t');
```

AlastairMcKee:
Made requested changes, jslint is passing on my machine

AlastairMcKee:
Made review changes to cover additional imports

docs/app/services/plunker/plunker.service.ts
```diff
@@ -36,9 +36,24 @@ export class PlunkerService {
...
-            modules += plunk.modules.map(mapping => `${ mapping.imports }`);
-            modules = ',' + modules;
-            imports = plunk.modules.map(mapping => mapping.library ? `import { ${ mapping.imports } } from '${ mapping.library }';`: '').join('\n');
+            plunk.modules.map(mapping => {
+                if(mapping.imports){ 
+                    modules += `,${ mapping.imports }`;
+                }
+            });
+            plunk.modules.map(mapping => {
+                if(mapping.library){
```
ashley-hunter: Missing whitespace in this code is causing error when running `grunt lint`

AlastairMcKee:
Made those changes, lint seems to be passing now

PR #76, opened 2017-04-21
Fixing TS Lint errors

ashley-hunter:
Fixing ts lint errors that appeared after upgrading ts lint to 5.1.0


AlastairMcKee:
lint is passing, looks good

PR #80, opened 2017-04-24
EL-2477

donna-marijne:
Added $navigationMenu and modified minimizeMenu to make use of it.
Added documentation and a CodePen example using $navigationMenu to
control showing/hiding of the nav menu and enabling/disabling auto-hide
behaviour.

https://jira.autonomy.com/browse/EL-2477


ashley-hunter:
Changes look good, this seems like a much better way of handling things than before. Do you think this would be better documented under the Components page?

donna-marijne:
Yes, I think that would be reasonable. In that case I expect that the whole Side Navigation section should move, since none of that is pure CSS; also, the Page Header > Menu Button section which belongs together with this service.

ashley-hunter:
Yes, I agree, I think that makes much more sense. This feature always did seem a bit out of place on this page to me.

donna-marijne:
Moved relevant sections to Components.

PR #89, opened 2017-04-25
EL-2428 - Moving expression to timeout

ashley-hunter:
Wrapping expression change in a timeout as it occurs in afterViewInit hook


ashley-hunter:
Not sure who is available to do a review so have assigned both of you and whoever is free can take a quick look

donna-marijne:
I see that it's the color patch which causes an issue with sanitization, is there any way to modify that template?

ashley-hunter:
Yes I there are some options to remove the colour patch. I kept it in to match the previous flot version of the chart and can imagine it being common place in a legend. I don't necessarily think it's a bad thing to show an example of using the sanitizer to illustrate some of the security features available in Angular.

PR #101, opened 2017-04-26
EL-2509

donna-marijne:
Removed dist from git.


donna-marijne:
Yes, the delete is versioned, so the files are available in any previous commit/tag.

PR #104, opened 2017-04-27
EL-2512 - Draggable cards dynamically changing edit and remove

AlastairMcKee:
allowEditing & allowRemoving can now be dynamically updated


AlastairMcKee:
Made changes

PR #106, opened 2017-04-27
EL-2529 Complete migration of release build script.

PearseMcMurray:
Script will execute unit and Selenium tests.
Tidied output of unit tests.

### DIFF
--- a/buildscripts/buildRelease.sh
+++ b/buildscripts/buildRelease.sh
@@ -16,7 +16,6 @@ echo http_proxy is $http_proxy
 echo https_proxy is $https_proxy
 echo BuildPackages is $BuildPackages
 echo PrivateArtifactoryURL is $PrivateArtifactoryURL
-echo PrivateArtifactoryCredentials is $PrivateArtifactoryCredentials
 echo BuildDocumentation is $BuildDocumentation
 echo GridHubIPAddress is $GridHubIPAddress
 echo Build number is $BUILD_NUMBER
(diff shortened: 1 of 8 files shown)


buildscripts/buildRelease.sh
```diff
@@ -130,7 +132,9 @@ if [ "$RunTests" == "true" ]; then
 	while read line ; do
 		echo "<p><span class=rvts6>$line</span></p>" >> UXAspectsTestsResults.html
 	done < UnitTestResults.txt
+	sed -i 's/\[1m//g' UXAspectsTestsResults.html
 	sed -i 's/\[4m//g' UXAspectsTestsResults.html
+	sed -i 's/\[22m//g' UXAspectsTestsResults.html
 	sed -i 's/\[24m//g' UXAspectsTestsResults.html
 	sed -i 's/\[31m//g' UXAspectsTestsResults.html
 	sed -i 's/\[32m//g' UXAspectsTestsResults.html
```
donna-marijne: I would suggest that just the summary of the jasmine run be presented here rather than the full output. This should be easier to read since it only presents the number of tests executed and the failures. It also does not include control characters. This can be accomplished by adding `display: 'none'` and `summary: true` to the options in grunt/jasmine.js.

I also suggest that only the output from the line "Testing Jasmine specs via PhantomJS" onwards be included in the HTML.

buildscripts/executeSeleniumTestsReleaseBuild.sh
```diff
@@ -0,0 +1,198 @@
...
+echo Listing the $HOME_FOLDER/ux-aspects folder
+ls -al $HOME_FOLDER/ux-aspects
+cd $HOME_FOLDER/ux-aspects
+
+# Update connect.js to allow building of the documentation. Publish on port 4001.
+echo Updating grunt/connect.js
+sed -i.bak 's/hostname: '"'"'localhost'"'"'/hostname: '"'"'0.0.0.0'"'"'/' grunt/connect.js
+sed -i.bak 's/livereload: true/livereload: false/g' grunt/connect.js
+sed -i.bak 's/4000/4001/g' grunt/connect.js
+
```
donna-marijne: I would recommend creating a "connect:selenium_release" grunt task with these parameters already populated, to avoid possible replace errors.

buildscripts/executeSeleniumTestsReleaseBuild.sh
```diff
@@ -0,0 +1,198 @@
...
+sleep 10
+PID_SELENIUM=`/usr/sbin/fuser -n tcp 4445 2> /dev/null`
+echo After kill, Selenium Grid hub process ID is $PID_SELENIUM
+
+echo Start the new Selenium Grid hub process
+cd $HOME_FOLDER/ux-aspects
+rm -rf target
+cd $HOME_FOLDER/ux-aspects/configuration
+java -jar /home/UXAspectsTestUser/ux-aspects/Selenium/selenium-server-standalone-3.3.1.jar -role hub -hubConfig hub/hubConfigReleaseBuild.json &
+sleep 15
```
donna-marijne: Potentially a loop could be added here, making this part more resilient and possibly reducing the duration of the script. Can a loop function be written so that similar checks can reuse some code?

PearseMcMurray:
Made changes after review.

donna-marijne:
This looks fine. Let's do a trial run build to test when you have merged.

PR #111, opened 2017-05-02
EL-2467 - Added replace true back into help center menu

AlastairMcKee:
Removing replace true caused some issues with the help center menu, I will raise a ticket to add a wrapper for this instead.


AlastairMcKee:
Removed replace true, updated failing unit tests and fixed up styling which was broken by the change

PR #117, opened 2017-05-04
S 57/el 2507

AlastairMcKee:
Added usage section and updated plunker and codepen links to match Rolands design.


docs/app/components/usage-link/usage-link.component.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    @Input() usage: Usage;
+    @Input() pop: string;
+
+    isUsageEnabled() {
+        return this.usage;
+    }
+
+    onShown(popover: any) {
+        let popoverElement = popover._popover._componentRef.location.nativeElement;
```
ashley-hunter: We shouldnt directly modify the DOM, instead inject and use Renderer2

docs/app/components/usage-link/usage-link.component.ts
```diff
@@ -0,0 +1,39 @@
...
+        popoverElement.style.width = '300px';
+        popoverElement.style.maxWidth = '300px';
+        popoverElement.style.borderRadius = '0';
+        popoverElement.style.zIndex = '1';
+    }
+
+    //copy to clipboard button
+    copy(text: string) {
+        let dummy = document.createElement('input');
```
ashley-hunter: Most of this should also be done using the renderer2

AlastairMcKee:
Added keyboard support to both usage popover and plunker/codepen link

docs/app/interfaces/UsageProvider.ts
```diff
@@ -0,0 +1,9 @@
+import { Usage } from './Usage';
+
+export interface UsageProvider {
+    usage: Usage;
+}
+
+export function isUsageProvider(obj: any) {
```
ashley-hunter: We can get rid of this usage provider. Instead just get the information from the json files we already use (eg. charts-page.json) this way we don't have to implement the interface every time etc..

AlastairMcKee:
Changed to use renderer2 and removed the usage provider

docs/app/interfaces/ICategory.ts
```diff
@@ -1,7 +1,9 @@
+import { Usage } from './Usage';
 import { ISection } from './ISection';
 
 export interface ICategory {
     title: string;
     link: string;
+    usage: Usage[];
```
ashley-hunter: Make this an optional property

AlastairMcKee: This actually won't be required as the usage section will be inside "sections"

AlastairMcKee:
Resolved conflicts and added missing usage sections

docs/app/components/usage-link/usage-link.component.ts
```diff
@@ -0,0 +1,44 @@
+import { Usage } from './../../interfaces/Usage';
+import { Input, Renderer2, Component } from '@angular/core';
+
+@Component({
+    selector: 'uxd-usage-link',
+    templateUrl: './usage-link.component.html',
+    styleUrls: ['./usage-link.component.less'],
+    host: {
+        '[class.enabled]': 'isUsageEnabled()'
```
ashley-hunter: Is this needed anymore (and the function isUsageEnabled)? as we use an ngIf not to show this component at all if there is no usage value.

if the off chance that this is needed you should change it to:

```
'[class.enabled]': 'usage'
```

As calling a function is much less performant than just accessing a value as it can trigger change detection

AlastairMcKee:
Removed isUsageEnabled function, ngIf takes care of it.

PR #118, opened 2017-05-04
EL-2434 - Migration of Facets

ashley-hunter:
Migrating the facet components. Facet architecture has been redesigned so while providing essentially the same functionality names and inputs/outputs will differ.

There is now a base facet class that makes it easy to create new facets.

Also adding declaration option to plunker service


ashley-hunter:
Typeahead filtering should now work as expected

ashley-hunter:
Making typeahead search case insensitive. Should now be fixed

PR #121, opened 2017-05-05
Adding package.json file for the npm package

ashley-hunter:


src/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "ux",
+    "aspects",
+    "ui",
+    "framework",
+    "angular",
+    "angularjs",
+    "bootstrap"
+  ],
+  "author": "Ashley Hunter",
+  "license": "Apache-2.0",
```
donna-marijne: Are we not using the MIT license?

src/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "url": "https://github.com/UXAspects/UXAspects/issues"
+  },
+  "homepage": "https://uxaspects.github.io/UXAspects",
+  "dependencies": {
+    "chart.js": "^2.5.0",
+    "ng2-charts": "^1.5.0",
+    "ngx-bootstrap": "^1.6.6"
+  },
+  "devDependencies": {
+    "@types/chart.js": "^2.4.2",
```
donna-marijne: Typescript?

ashley-hunter: Didn't add typescript as technically they can use plain JS or Dart either. Other libs like ngx-bootstrap don't appear to include it as a dependency. I added the type definitions anyway because you can now use the Typescript language service to give hints in regular JS files.

src/package.json
ashley-hunter: I thought so, but wasn't sure. Looked at all other open source CAF services and they were all using Apache 2.0 license. Do you know where we can check this?

donna-marijne: I suppose we would have to ask Andy, but my guess is that we are using the most permissive license (MIT) due to the fact that we currently bundle so much open source code with various licenses.

The MIT license appears here, although slightly modified:
https://uxaspects.github.io/UXAspects/#/licenses

donna-marijne: Although with that said, I think that most of the 3rd party licenses are permissive enough that we can use whichever license we wish. The only one I'm not sure of is the font license.

ashley-hunter:
Changed license to MIT

PR #150, opened 2017-05-11
EL-2514 Continuous rebuild of documentation, using both themes.

PearseMcMurray:


buildscripts/buildTheme.sh
```diff
@@ -0,0 +1,95 @@
...
+    echo ID for $UX_ASPECTS_BUILD_IMAGE_NAME:$UX_ASPECTS_BUILD_IMAGE_TAG_LATEST image is $DOCKER_IMAGE_ID
+    if [ -z "$DOCKER_IMAGE_ID" ] ; then
+        echo Image $UX_ASPECTS_BUILD_IMAGE_NAME:$UX_ASPECTS_BUILD_IMAGE_TAG_LATEST does not exist!
+    else
+        echo Calling docker run ... "$@"
+        docker run --rm --volume "$PWD":/workspace --workdir /workspace --user \
+		    $UID:$GROUPS $UX_ASPECTS_BUILD_IMAGE_NAME:$UX_ASPECTS_BUILD_IMAGE_TAG_LATEST "$@"
+    fi
+}
+
```
donna-marijne: These functions looks like they are common to other files. If so please move them to a common script file as discussed previously, and source the file here. Parameters should be added if anything needs to be adjusted.

buildscripts/executeUnitTests.sh
```diff
@@ -130,17 +132,96 @@ cp index.html $WORKSPACE/reports/index.html
...
 then
     echo Unit tests passed
-    exit 0
 else
     echo "Unit test(s) failed"
     exit 1
 fi
+
+# Clear up previous documentation builds
+rm -rf $WORKSPACE/ux-aspects/KeppelThemeFiles
```
donna-marijne: This file is named executeUnitTests.sh - this part ought to be in a file named after what it does. I'd suggest making a continousIntegration master script which calls scripts for each discrete task that it performs.

buildscripts/updateConfigJSON.sh
```diff
@@ -0,0 +1,19 @@
...
+set -e
+
+theme=$1
+echo "$0" - theme is $theme
+echo "$0" - PWD is $PWD
+
+# Create a modified version of config.json with a new assetsUrl value
+rm -f jqFilter
+echo ".assetsUrl=\"https://pages.github.hpe.com/sepg-docs-qa/UXAspects_"$theme"_Theme_CI/assets\"" > jqFilter
+cat docs/app/data/config.json | jq -f jqFilter > docs/app/data/config_updated.json
```
donna-marijne: You could use an environment variable here I would think, instead of creating and deleting files.

buildscripts/buildTheme.sh
PearseMcMurray: Duplicated functions moved into one file and file is included by the other scripts. Parameters added to the functions where needed.

buildscripts/executeUnitTests.sh
PearseMcMurray: Added executeContinuousIntegrationTasks.sh to execute the unit tests and rebuilt the themed documentation sites.

buildscripts/updateConfigJSON.sh
PearseMcMurray: jq command executed with filter in environment variable instead of text file.

PR #152, opened 2017-05-12
EL-2494

donna-marijne:
Added webpack.dev.config.js which includes webpack configs for docs and each of the 3 assets needed to run examples externally (lib, ng1, css). Since I could not get the grunt-webpack plugin to handle multiple
configs at once, this is currently launched by `npm start`.
Added a grunt task to install the webpack-dev-server localhost certificate on Windows.
Fixed CSS urls to be correctly relative.
Added a config.dev.json which uses the localhost assetsUrl, this will be used when the dev server is run.
Updated the showcase to use css and ng1 from assets, so that can be rebuilt by the dev server.
Removed the connect:plunker task.
Updated the readme.

https://jira.autonomy.com/browse/EL-2494


README.md
```diff
@@ -44,11 +44,15 @@ git clone https://github.com/UXAspects/UXAspects.git
 ```bash
 npm install
 ```
+3. (Optional) Add the webpack-dev-server certificate to the machine's certificate store. This will allow the development server to run using HTTPS without warnings, which is required to test Plunker/CodePen examples locally. On Windows:
```
ashley-hunter: To get this working I had to install openssl on my machine and add it to environment variables. Perhaps this step should be documented?

donna-marijne:
Added some more details to the readme.

PR #153, opened 2017-05-15
S 58/el 2440

AlastairMcKee:
Added a new column sorting component which will replace both single and multiple column sorting. Added plunker and updated Angular 1 components to deprecated.

https://jira.autonomy.com/browse/EL-2440

Other sorting components mentioned in this ticket will not be replaced this sprint.


docs/app/pages/components/sections/tables/column-sorting/column-sorting.component.ts
```diff
@@ -0,0 +1,155 @@
...
+        active: chance.bool()
+    }];
+
+    changeState(columnSortingComponent: ColumnSortingComponent) {
+        this.order = columnSortingComponent.changeState();
+        this.sortByKey(this.sortableTable, this.order);
+    }
+
+    sortByKey(array: object[], order: object) {
+        return array.sort( function(a: any, b: any) {
```
ashley-hunter: All this sorting logic can be simplified:

```typescript
return array.sort((itemOne: TypeHere, itemTwo: TypeHere) => {

    // iterate through each sorter
    for (let sorter of order) {

        let value1 = itemOne[sorter.key];
        let value2 = itemTwo[sorter.key];

        if (value1 < value2) {
            return -1;
        } else if (value1 > value2) {
            return 1;
         }

    }

    return itemOne.id < itemTwo.id ? -1 : 1;
});
```

docs/app/pages/components/sections/tables/column-sorting/column-sorting.component.html
```diff
@@ -0,0 +1,74 @@
...
+    <tbody>
+        <tr class="clickable" *ngFor="let item of sortableTable">
+            <td>
+                <div class="listview-page-icon"></div>
+            </td>
+            <td>{{ item.name }}</td>
+            <td>{{ item.author }}</td>
+            <td>{{ item.date }}</td>
+            <td>
+                <ux-spark trackColor="rgba(139,107,193,0.2)" barColor="#7b63a3" [value]="item.completed" fillHeight="3" [inlineLabel]="item.completed"></ux-spark>
```
ashley-hunter: Use color service for spark colors

src/components/column-sorting/column-sorting.component.ts
```diff
@@ -0,0 +1,57 @@
...
+import { Subject } from 'rxjs/Subject';
+
+@Component({
+    selector: 'ux-column-sorting',
+    templateUrl: './column-sorting.component.html',
+    exportAs: 'ux-column-sorting'
+})
+export class ColumnSortingComponent {
+    
+    @Input() state: ColumnSortingState;
```
ashley-hunter: We should allow state to be a two way binding, so we need to add an output for stateChange that will emit an event whenever the state of this sorter changes

src/components/column-sorting/column-sorting.component.less
```diff
@@ -0,0 +1,36 @@
+@import "../../styles/variables.less";
+ux-column-sorting{
+    .ux-column-sorting {
+        padding-left: 5px;
+        display: inline-block;
+        width: 36px;
+        height: 28px;
```
AlastairMcKee: This is required to keep the number sitting top the top right of the icon as it does in ng1 column sorting.

ashley-hunter: Is that expected? I thought that was a bug in the angular 1 version? Looking at an old version they are inline:

image

AlastairMcKee: Checked with Roland, he wants to retain the slightly raised numbers.

AlastairMcKee:
Made review changes and added some changes involving singleSort which only allows 1 column to be sorted at once

docs/app/pages/components/sections/tables/column-sorting/column-sorting.component.html
```diff
@@ -62,12 +60,25 @@
             <td>Defines the sort key for the column.</td>
         </tr>
 
+        <tr>
```
ashley-hunter: Rename the 'sortKey' property to just 'key'. As part of the migration it's good to try simplifying attribute names, where they were previously more verbose than they needed to be.

Might also be good to add a bit more detail to the description to explain it's purpose as it currently isn't overly clear what it's purpose is.

src/components/column-sorting/column-sorting.component.ts
```diff
@@ -13,6 +11,8 @@ export class ColumnSortingComponent {
     @Input() state: ColumnSortingState;
     @Input() sortKey: string;
     @Input() orderNumber: number;
```
ashley-hunter: Could we run a code tidy on this file as some of the indentation is incorrect. (Alt+Shift+F in VS Code should do it for you)

docs/app/pages/components/sections/tables/column-sorting/column-sorting.component.html
```diff
@@ -62,12 +60,25 @@
...
+            <td><b>@Input</b> singleSort</td>
```
ashley-hunter: I think this should be moved to the Column Sorting directive rather than the component, otherwise they will have to add this on each column sorter they use, rather than just once on the directive.

AlastairMcKee:
Made review changes

PR #154, opened 2017-05-15
EL-2436 - Slider Component

ashley-hunter:
Creating slider component. Moved sliders to it's own page.


docs/app/pages/components/sections/sliders/sliders/snippets/sample4.html
```diff
@@ -0,0 +1,19 @@
+<div class="row">
+
+    <div class="col-md-12">
+        <h4 class="text-black p-t-md p-b-md">Range Slider with Text Inputs</h4>
+    </div>
+
+    <div class="col-md-12">
+        <ux-slider [value]="slider7.value" (valueChange)="updateValue($event)" [options]="slider7.options"></ux-slider>
```
donna-marijne: Can you add linebreaks to this and the TypeScript to avoid side-scrolling?

src/components/slider/slider.component.ts
```diff
@@ -0,0 +1,776 @@
...
+
+    private unionTicks(majorTicks: SliderTick[], minorTicks: SliderTick[]): SliderTick[] {
+
+        // get all ticks combined removing any minor ticks with the same value as major ticks
+        return majorTicks.concat(minorTicks)
+            .filter((tick, index, array) => tick.type === SliderTickType.Major || !array.find(tk => tk.type === SliderTickType.Major && tk.position === tick.position))
+            .sort((t1, t2) => t1.value - t2.value);
+    }
+
+    private deepMerge(destination: any, source: any) {
```
donna-marijne: Would these be worth adding to Object or otherwise factoring out to be globally available?

docs/app/pages/components/sections/sliders/sliders/sliders.component.html
```diff
@@ -0,0 +1,326 @@
...
+
+<div class="demo-attributes table-responsive">
+    <table class="table">
+        <tr>
+            <th>Property</th>
+            <th>Description</th>
+        </tr>
+        <tr>
+            <td class="attribute">type</td>
+            <td>Defines whether the slider should allow selection of a single value or a range of values. The two valid values
```
donna-marijne: Can you list the types of each property in these tables? If you break down the SliderOptions type you could also list e.g. SliderHandleOptions here and then name the type in the subsequent section.

docs/app/data/components-page.json
```diff
@@ -682,6 +672,37 @@
             ]
         },
         {
+            "title": "Sliders",
```
ashley-hunter: The performance of the sliders was a bit slow when added to the input controls section, however I have moved it back as the performance issue will be improved once we remove the angular 1 components eventually.

src/components/slider/slider.component.ts
ashley-hunter: I did consider doing this, however this implementation will not work for certain edge cases such as arrays of objects etc.. which would have required quite a bit more work to cover all possibilities.

ashley-hunter:
Changes made

docs/app/data/components-page.json
donna-marijne: Yes I see, is that from the jscrollpane in Custom Dropdown? Because that doesn't really count as an input control in my opinion and could be moved to the Buttons section.

PR #156, opened 2017-05-16
EL-2547 - Bower Module Fix

ashley-hunter:
Tidying up bower.json file. Added instructions on how to configure TypeScript to find the `ux-aspects` module as typescript currently only searches the node_modules folder by default.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name": "ux-aspects",
   "version": "1.1.0",
   "description": "Responsive UI Framework for Big Data Applications - CSS & Angular modules.",
-  "keywords": [    
+  "keywords": [
     "js",
     "angular",
     "css",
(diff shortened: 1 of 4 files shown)


donna-marijne:
Isn't it something that needs to be configured in webpack or whatever other library implements require in the target application?

Bower sure is a pain these days!

ashley-hunter:
Couldn't agree more, bower really is a nuisance.

Unfortunately not, our current webpack configuration works as expected. Webpack is finding the files correctly, but as it essentially just runs the typescript compiler behind the scenes, it is at this point typescript tries and fails as Typescripts module resolution defaults to node_modules folder unless otherwise specified.

donna-marijne:
Understood - this looks fine and I tested it in the seed project successfully - thanks.

PR #184, opened 2017-05-24
Adding linting for component stylesheets

ashley-hunter:
Adding linting for Angular 4 component less files. Recommended the team install stylelint plugin for vscode, seems very good.

Integrated into grunt lint task.
Will need to run npm install for this to work.

Modified existing stylesheets to make compliant with new rules.


ashley-hunter:
I would really like it if we could, I took at look at linting our ng1 styles, but there we 17,000+ places where it did not conform to the style rules. Im sure there is likely a tool or something that we could use to automatically fix most of these issues, but I think another jira would be required to go through our ng1 styles and make them follow the style rules.

donna-marijne:
Yikes - yes, we'll take some time to clean that up in another ticket. Thanks!

PR #186, opened 2017-05-25
EL-2528 Remove old, and now incorrect, list of items from the 'ignore…

PearseMcMurray:
…' section

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -27,33 +27,5 @@
     "bootstrap": "~3.3.4",
     "angular": "~1.3.16"
   },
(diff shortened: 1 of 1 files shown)


bower.json
```diff
@@ -27,33 +27,5 @@
     "bootstrap": "~3.3.4",
     "angular": "~1.3.16"
   },
-  "ignore": [
```
ashley-hunter: Should we be removing all of these or just the ones that are no longer relevant, eg. emailable-report.html?

PR #193, opened 2017-06-01
EL-2438 - Changes to keyboard controls to keep consistent with Angula…

AlastairMcKee:
…r 1 version

Can now navigate down the table with arrow keys and the panel closes when pressing enter on the close button

### DIFF
--- a/docs/app/pages/components/sections/panels/item-display-panel/snippets/app.ts
+++ b/docs/app/pages/components/sections/panels/item-display-panel/snippets/app.ts
@@ -119,32 +119,32 @@ export class AppComponent {
     }
 
     previous() {
-        let id = this.selectedItem.id - 1;
-        this.selectedItem = this.items[id - 1];
-        this.updatePanel();
+        if (this.previousEnabled) {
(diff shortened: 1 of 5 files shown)


docs/app/pages/components/sections/panels/item-display-panel/snippets/app.ts
```diff
@@ -126,6 +130,24 @@ export class AppComponent {
         this.updatePanel();
     }
 
+    upArrow(event: KeyboardEvent) {
+        if (this.visible) {
+            event.preventDefault();
+            if (this.previousEnabled) {
```
ashley-hunter: We should probably remove this `if (this.previousEnabled)` from this function and instead in the `previous()` function we should have it check if it can move back or not instead (do the same for the next button too) eg,

```typescript
previous() {
     let id = this.selectedItem && this.selectedItem.id > 0 ? this.selectedItem.id - 1 : 0;
     this.selectedItem = this.items[id - 1];
     this.updatePanel();
}
```

AlastairMcKee:
Made changes

PR #195, opened 2017-06-02
EL-2511 - Fixing IE bug in Expanding Text Area

AlastairMcKee:
Fixed the placeholder issue in IE, this has just been added in our documentation, not the sample code or codepen as it works as expected there.


docs/app/pages/components/sections/input-controls/expanding-text-area-ng1/wrapper/expanding-text-area-wrapper.directive.html
```diff
@@ -1 +1 @@
-<textarea class="form-control" rows="1" placeholder="Enter text" expanding-textarea disable-return="true" max-visible-lines="2"></textarea>
\ No newline at end of file
+<textarea class="form-control" rows="1" expanding-textarea disable-return="true" max-visible-lines="2"></textarea>
```
ashley-hunter: Try the following to see if it works correctly:

```html
<textarea class="form-control" rows="1" expanding-textarea ng-attr-placeholder="{{'Enter text'}}" disable-return="true" max-visible-lines="2"></textarea>
```

AlastairMcKee:
Thar worked, checked in all the browsers. Theres a slight size difference between firefox and the other browsers (before the fix too) but its just how firefox handles text areas and there doesn't seem to be anything we can do about it without adding the controller back in.

PR #197, opened 2017-06-05
EL-2417 - Remove hover effect has been removed when removing is disabled

AlastairMcKee:
Requested by Roland: https://jira.autonomy.com/browse/EL-2417


AlastairMcKee:
Good spot, disabling the button didn't remove the pointer effect, made the change with CSS instead.

PR #200, opened 2017-06-06
S 59/el 2435

donna-marijne:
Tags section only.

https://jira.autonomy.com/browse/EL-2435


donna-marijne:
Whoever is most free to do a review please take a look, thanks.

PR #208, opened 2017-06-06
Pass name of file containing Docker container ID to function.

PearseMcMurray:
Sed command was failing to update testng.xml file with correct port values.


donna-marijne:
Selenium is no longer part of the release process.

PR #218, opened 2017-06-14
EL-2517-LabelsSpark

chris-brennan:
Added an example of the spark chart background and bar color to the spark chart table.  Added a table to show an example of the label classes that are available.  Updated the label examples section to use example text rather than the name of the label class, since this is now shown in the label table.


docs/app/pages/css/sections/labels/labels/labels.component.html
```diff
@@ -1,58 +1,126 @@
...
-<code>.label-warning</code> or
-<code>.label-critical</code></p>
+<div class="demo-attributes table-responsive">
+    <table class="table">
+        <tbody>
+            <tr>
+                <th>General Labels</th>
+                <th></th>
+                <th>Outline Labels</th>
+                <th></th>
```
donna-marijne: I think this would be better as two tables in column layout. If you use the col-* classes, you can get them to appear vertically for smaller screens.

See how it looks like this:

````
<div class="row">
  <div class="col-md-6 col-sm-12">
    <!-- General labels -->
  </div>
  <div class="col-md-6 col-sm-12">
    <!-- Outline labels -->
  </div>
</div>
````

See how that looks, it might be good with a small gap between the columns too.

File: docs/app/pages/css/sections/labels/labels/labels.component.html [hunk 2]
```diff
@@ -1,58 +1,126 @@
...
+<span class="label label-outline-vibrant1 m-r-sm">Company</span>
+<span class="label label-outline-alternate1">Location</span>
 
 <p class="m-t-md">Status Labels</p>
-<span class="label label-ok m-r-sm">Label OK</span>
-<span class="label label-warning m-r-sm">Label Warning</span>
-<span class="label label-critical">Label Critical</span>
+<span class="label label-ok m-r-sm">Completed</span>
+<span class="label label-warning m-r-sm">Cancelled</span>
+<span class="label label-critical">Failed</span>
```
donna-marijne: The example code and codepen are not up to date with this change - see the snippets subdirectory.

chris-brennan:
Updated angular 4 section.  Moved spark chart color reference above example code and applied consistent wording with labels section.  Removed extraneous columns from spark chart table (i.e. type, description and optional) and applied `col-md-6`.  Split labels color reference table into two separate tables.  Updated snippets and code pen.

docs/app/pages/charts/sections/spark-charts/spark-chart-ng1/spark-chart-ng1.component.ts
```diff
@@ -6,7 +6,7 @@ import { ICodePen } from '../../../../../interfaces/ICodePen';
 @Component({
     selector: 'uxd-charts-spark-chart-ng1',
     templateUrl: './spark-chart-ng1.component.html',
-    styleUrls: ['./spark-chart-ng1.component.less'],
+    styleUrls: ['./../spark-charts.less'],
```
donna-marijne: I don't think we should do this. Add common documentation styles to the docs/styles.less file.

docs/app/pages/charts/sections/spark-charts/spark-chart-ng1/spark-chart-ng1.component.html
```diff
@@ -128,6 +128,95 @@
...
+            <table class="table">
+                <tbody>
+                    <tr>
+                        <th>Spark Type</th>
+                        <th></th>
+                    </tr>
+                    <tr>
+                        <td class="attribute">spark-chart1</td>
+                        <td class="color-example-cell">
+                            <div class="spark spark-chart1 bg-color-example">
```
donna-marijne: I didn't notice this before - this should be an actual spark directive.

docs/app/pages/charts/sections/spark-charts/spark-charts/spark-charts.component.html
```diff
@@ -115,6 +118,95 @@
     </table>
 </div>
 
+<p>UX Aspects offers a wide range of chart colors. Set the color of the chart by using one of the following classes:</p>
```
donna-marijne: Since the angular 4 component allows arbitrary colors via barColor and trackColor, this description needs to be a bit different. Also the color service for Angular 4 has not yet been documented. I think doing this will take quite a bit more effort, so for now can you just revert the changes in the Angular 4 component? Sorry for the backtrack.

chris-brennan:
Wording of documentation text updated, spark directive used instead of CSS styles, common documentation styles moved to docs/styles.css, Angular 4 spark chart documentation reverted.

PR #221, opened 2017-06-16
S 60/el 2587

donna-marijne:
Added ux-select component and Select > Select documentation section.
Added uxInfiniteScroll directive and Scrollbar > Infinite Scroll documentation section.
Added paging via infinite scroll to the typeahead component.
Fixed some issues in ux-tag-input and refactored some of the key handling code for use in ux-select.
Added a mapping to plunker config to solve the plunker issue. Also added chance for use in examples.
Added styles for the accordion layout to be available in plunker as well.


donna-marijne:
Thanks! Made those changes.

PR #222, opened 2017-06-16
EL-2439 - UI Changes

AlastairMcKee:
UI change and the input box now clears between uses

### DIFF
--- a/src/components/filters/filter-dynamic/filter-dynamic.component.html
+++ b/src/components/filters/filter-dynamic/filter-dynamic.component.html
@@ -21,7 +21,7 @@
         <hr>
 
         <li *ngIf="showTypeahead" class="typeahead-box">
-            <input [(ngModel)]="searchQuery" [typeahead]="typeaheadItems" class="form-control" 
+            <input #inputBox [(ngModel)]="searchQuery" [typeahead]="typeaheadItems" class="form-control" 
             (typeaheadOnSelect)="selectOption($event)" 
             [placeholder]="options?.placeholder || defaultOptions.placeholder"
             [typeaheadMinLength]="options?.minCharacters || defaultOptions.minCharacters"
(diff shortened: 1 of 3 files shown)


src/components/filters/filter-dynamic/filter-dynamic.component.html
```diff
@@ -21,7 +21,7 @@
         <hr>
 
         <li *ngIf="showTypeahead" class="typeahead-box">
-            <input [(ngModel)]="searchQuery" [typeahead]="typeaheadItems" class="form-control" 
+            <input #inputBox [(ngModel)]="searchQuery" [typeahead]="typeaheadItems" class="form-control"
```
ashley-hunter: I don't think we need to do all this element ref stuff. We should just be able to do `this.searchQuery = ""` in the code anywhere we want to clear it?

src/components/filters/filter-dynamic/filter-dynamic.component.ts
```diff
@@ -39,14 +40,16 @@ export class FilterDynamicComponent extends FilterBaseComponent {
...
-
     }
 
     selectOption(typeaheadOption: TypeaheadMatch) { 
         this.removeFilter();
         let idx = this.filters.findIndex(filter => filter.name === typeaheadOption.value);
         this.selected = this.filters[idx];
         this.addFilter(this.selected);
+        if (this.inputBox) {
+            this.inputBox.nativeElement.value = '';
```
ashley-hunter: We should just be able to set `this.searchQuery = ''` here and the other places, however just for future reference, in the rare case where you do need to set the value of an element programmatically we should use the Renderer2 service which has a `setValue` method to do this. It also has many other similar methods which we should always use rather than directly manipulating an element. ([https://angular.io/api/core/Renderer2](https://angular.io/api/core/Renderer2))

AlastairMcKee:
Made those changes, worked as expected. Thanks

PR #224, opened 2017-06-16
S 60/el 2588

AlastairMcKee:
Added the version switcher component to the components and charts pages, it has been left out of the CSS page as there is nothing to be filtered on that page.

### DIFF
--- a/docs/app/components/documentation-category/documentation-category.component.html
+++ b/docs/app/components/documentation-category/documentation-category.component.html
@@ -1,6 +1,6 @@
 <span *ngFor="let section of category.sections">
     <uxd-component-section [id]="section.id" [title]="section.title" [componentName]="section.component"
         [version]="section.version" [deprecated]="section.deprecated" [externalUrl]="section.externalUrl" [usage]="section.usage"
-        *ngIf="!angular && section.version === 'AngularJS' ||
-        angular && !section.deprecated"></uxd-component-section>
+        *ngIf="(versionService.version | async) !== ngVersion.Angular && section.version === 'AngularJS' ||
+        (versionService.version | async) === ngVersion.Angular && !section.deprecated"></uxd-component-section>
 </span>
--- a/docs/app/components/documentation-category/documentation-category.component.html
+++ b/docs/app/components/documentation-category/documentation-category.component.html
@@ -1,3 +1,4 @@
 <uxd-component-section *ngFor="let section of category.sections" [id]="section.id" [title]="section.title" [componentName]="section.component"
     [version]="section.version" [deprecated]="section.deprecated" [externalUrl]="section.externalUrl" [usage]="section.usage"
-    [hidden]="angular4 && section.version === 'AngularJS' && section.deprecated === true"></uxd-component-section>
+    [hidden]="angular && section.version === 'AngularJS' && section.deprecated === true ||
+    !angular && section.version === 'Angular'"></uxd-component-section>
--- a/docs/app/components/documentation-category/documentation-category.component.html
+++ b/docs/app/components/documentation-category/documentation-category.component.html
@@ -1,2 +1,3 @@
 <uxd-component-section *ngFor="let section of category.sections" [id]="section.id" [title]="section.title" [componentName]="section.component"
-    [version]="section.version" [deprecated]="section.deprecated" [externalUrl]="section.externalUrl" [usage]="section.usage"></uxd-component-section>
+    [version]="section.version" [deprecated]="section.deprecated" [externalUrl]="section.externalUrl" [usage]="section.usage"
+    [hidden]="angular4 && section.version === 'AngularJS' && section.deprecated === true"></uxd-component-section>
--- a/docs/app/components/documentation-category/documentation-category.component.html
+++ b/docs/app/components/documentation-category/documentation-category.component.html
@@ -1,4 +1,6 @@
-<uxd-component-section *ngFor="let section of category.sections" [id]="section.id" [title]="section.title" [componentName]="section.component"
-    [version]="section.version" [deprecated]="section.deprecated" [externalUrl]="section.externalUrl" [usage]="section.usage"
-    [hidden]="angular && section.version === 'AngularJS' && section.deprecated === true ||
-    !angular && section.version === 'Angular'"></uxd-component-section>
+<span *ngFor="let section of category.sections">
+    <uxd-component-section [id]="section.id" [title]="section.title" [componentName]="section.component"
+        [version]="section.version" [deprecated]="section.deprecated" [externalUrl]="section.externalUrl" [usage]="section.usage"
+        *ngIf="!angular && section.version === 'AngularJS' ||
+        angular && !section.deprecated"></uxd-component-section>
+</span>
--- a/docs/app/components/documentation-category/documentation-category.component.ts
+++ b/docs/app/components/documentation-category/documentation-category.component.ts
@@ -13,6 +13,8 @@ export class DocumentationCategoryComponent implements OnInit, AfterViewInit {
 
     private category: ICategory;
     private trackScroll: boolean = false;
+
+    angular4: boolean = true;
     
     constructor(private activatedRoute: ActivatedRoute,
         private navigation: NavigationService) { }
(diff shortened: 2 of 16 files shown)


docs/app/components/documentation-category/documentation-category.component.html
```diff
@@ -1,2 +1,4 @@
 <uxd-component-section *ngFor="let section of category.sections" [id]="section.id" [title]="section.title" [componentName]="section.component"
-    [version]="section.version" [deprecated]="section.deprecated" [externalUrl]="section.externalUrl" [usage]="section.usage"></uxd-component-section>
+    [version]="section.version" [deprecated]="section.deprecated" [externalUrl]="section.externalUrl" [usage]="section.usage"
+    [hidden]="angular && section.version === 'AngularJS' && section.deprecated === true ||
```
ashley-hunter: We should use `ngIf` instead of hidden as hidden still keeps everything in the dom and running so we aren't getting any of the performance benefits by hiding them. I realise there is an ngFor here so it will complain about using another structural directive on the same element, however you could move the ngIf to the template of the `uxd-component-section` perhaps

docs/app/services/version/version.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable, Output, EventEmitter } from '@angular/core';
+
+
+@Injectable()
+export class VersionService {
+
+    version: string = 'Angular';
+
+    @Output() versionChange: EventEmitter<string> = new EventEmitter<string>();
```
ashley-hunter: We shouldn't be using `@Output` in a service as this is for components to indicate that they are emitting values. We also shouldn't use and `EventEmitter`, instead you should use a `BehaviorSubject` instead. Currently you have a version variable and an event emitter, a behavior subject is essentially a merge of the two, it stores the value and can be subscribed to so you wont need two separate variables, so you can remove `version`, and rename `versionChange` to `version`.

You should create an enum for the different Angular versions as using a string could cause problems, for example if there is a typo in a string then things may not work as expected.

You should also store the selected version in local storage so if the page is refreshed the version stays the same.

docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
```diff
@@ -45,27 +58,51 @@ export class NavigationBarSearchComponent {
...
         page.categories.forEach((category: ICategory) => {
-            results.push({
-                section: page.title,
-                link: {
-                    title: category.title,
-                    link: category.link
-                }
-            });
+            let addCategory = false;
+            for (let i = 0; i < category.sections.length; i++) {
```
ashley-hunter: Use the `.filter()` function here on the sections array instead

docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
```diff
@@ -45,27 +58,51 @@ export class NavigationBarSearchComponent {
...
                 category.sections.forEach((section: ISection) => {
-                    results.push({
-                        section: page.title,
-                        link: {
-                            title: section.title,
-                            link: category.link,
-                            fragment: section.id
-                        }
-                    });
+                    if (this.angular && !section.deprecated) {
```
ashley-hunter: Could the if and elseif not be merged here to just:

```typescript
if (this.angular && !section.deprecated || !this.angular && section.version === 'AngularJS') {
results.push({
    section: page.title,
    link: {
        title: section.title,
        link: category.link,
        fragment: section.id
        }
    });
}
```

docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
```diff
@@ -6,6 +6,7 @@ import { ISearchResult } from '../../interfaces/ISearch';
 import { ICategory } from '../../interfaces/ICategory';
 import { ISection } from '../../interfaces/ISection';
 import { NavigationService } from '../../services/navigation/navigation.service';
+import { VersionService } from "../../services/version/version.service";
```
ashley-hunter: Grunt lint also fails in this file

AlastairMcKee:
Made requested changes

ashley-hunter:
Conflicts need resolved

docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
```diff
@@ -32,7 +33,21 @@ export class NavigationBarSearchComponent {
     private data: ISearchResult[];
     private history: ISearchResult[] = [];
 
-    constructor(private router: Router, private navigation: NavigationService) {
+    private angular = true;
+
+    private ngVersions = Version;
```
ashley-hunter: Remove this line. We only need to do this if we want to expose the `Version` enum to the template. When using it in a class you can just do `Version.AngularJS` etc..

docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
```diff
@@ -32,7 +33,21 @@ export class NavigationBarSearchComponent {
...
+
+    constructor(private router: Router, private navigation: NavigationService, private versionService: VersionService) {
+
+        // get version
+        this.versionService.version.subscribe((value: Version) => {
+            this.angular = value === this.ngVersions.Angular;
```
ashley-hunter: Remove this, when checking version compare the `this.versionService.version.getValue() === Versions.Angular` instead. Incase we ever wish to extend the number of versions, for example add a CSS version then this wouldn't work just using a boolean

docs/app/components/documentation-category/documentation-category.component.ts
```diff
@@ -13,9 +14,24 @@ export class DocumentationCategoryComponent implements OnInit, AfterViewInit {
 
     private category: ICategory;
     private trackScroll: boolean = false;
+
+    angular: boolean = true;
```
ashley-hunter: As mentioned in the class below, remove this and compare directly to version in service

docs/app/services/version/version.service.ts
```diff
@@ -0,0 +1,30 @@
...
+import { BehaviorSubject } from 'rxjs/BehaviorSubject';
+
+
+@Injectable()
+export class VersionService {
+
+    version: BehaviorSubject<Version> = new BehaviorSubject<Version>(Version.Angular);
+
+    constructor() {
+        this.toggle(this.toVersion(window.localStorage.getItem('version')));
```
ashley-hunter: If there is no previously stored version in local storage this toVersion call breaks the application. Need to add a check in `toVersion` to handle when version is empty

docs/app/services/version/version.service.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    toggle(version: Version): void {
+        if (this.version.getValue() !== version) {
+            window.localStorage.setItem('version', version.toString());
+            this.version.next(version);
+        }
+    }
+
+    toVersion(version: string): Version {
+        return version.toLowerCase() === '0' ? Version.AngularJS : Version.Angular;
```
ashley-hunter: Convert `version` variable passed in to a number then compare to `Version.Angular`. Typescript creates the enum for us, so there is no way to guarantee that 0 will always mean `Angular`

AlastairMcKee:
Resolved conflicts

docs/app/services/version/version.service.ts
```diff
@@ -0,0 +1,30 @@
...
+@Injectable()
+export class VersionService {
+
+    version: BehaviorSubject<Version> = new BehaviorSubject<Version>(Version.Angular);
+
+    constructor() {
+        this.toggle(this.toVersion(window.localStorage.getItem('version')));
+    }
+
+    toggle(version: Version): void {
```
ashley-hunter: Rename this to `setVersion` as `toggle` implies it switches to the next version which isn't the behavior of the method.

AlastairMcKee:
Made changes

docs/app/services/version/version.service.ts
```diff
@@ -8,18 +8,22 @@ export class VersionService {
...
         if (this.version.getValue() !== version) {
             window.localStorage.setItem('version', version.toString());
             this.version.next(version);
         }
     }
 
     toVersion(version: string): Version {
-        return version.toLowerCase() === '0' ? Version.AngularJS : Version.Angular;
+        let value = parseInt(version);
+        if (!value) {
```
ashley-hunter: Change this to `isNaN(value)` in the if statement as parseInt(undefined) will be equal to NaN. Also set the value in this case to the enum rather than a number as we can't be sure that 1 will always represent Angular/AngularJS

AlastairMcKee:
Fixed NaN issue

PR #225, opened 2017-06-16
EL-2591

donna-marijne:
Updated to clear the query when closing the search field.
Added a search history which is displayed when the query is empty.
Added an alt+/ keyboard shortcut.

https://jira.autonomy.com/browse/EL-2591

### DIFF
--- a/docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
+++ b/docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
@@ -9,6 +9,7 @@ import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 
 const QUERY_MIN_CHARS = 3;
 const MAX_HISTORY = 5;
+const LOCAL_STORAGE_KEY = 'uxd-search-history';
 
 @Component({
     selector: 'uxd-navigation-bar-search',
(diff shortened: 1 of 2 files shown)


docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
```diff
@@ -130,28 +146,34 @@ export class NavigationBarSearchComponent {
...
-        }
 
-        // get the latest results
-        this.results = this.data.filter((item: ISearchResult) => {
-            return item.link.title.toLowerCase().indexOf(this.query.toLowerCase()) !== -1;
-        });
+        } else {
+
+            // get the latest results
+            this.results = this.data.filter((item: ISearchResult) => {
```
ashley-hunter: I notice if I type in the name of a section with subsections of the same name eg. dashboard, im getting duplicate results, we should probably filter them out?

image

docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
```diff
@@ -160,4 +182,24 @@ export class NavigationBarSearchComponent {
...
         return this.results.filter((result: ISearchResult) => result.link.title === item.link.title).length > 1;
     }
+
+    /**
+     * Add an item to the bounded history list.
+     */
+    private addToHistory(item: ISearchResult) {
+
+        // Remove the item if it's already there.
+        const historyIndex = this.history.indexOf(item);
```
ashley-hunter: The old site stored the last 5 results in local storage so they persisted across page refreshes. I realise this was required on the old site as it wasnt a single page application, however do you think this feature should be carried across?

donna-marijne: Yes it's a good idea, especially because of the showcase.

donna-marijne:
Made the above changes - thanks.

PR #238, opened 2017-06-20
EL-2588 - QA Fix

AlastairMcKee:
Buttons no longer scroll off the screen


docs/app/components/side-navigation/side-navigation.component.html
```diff
@@ -1,5 +1,6 @@
 <div class="side-navigation-container" #container>
     <div class="side-navigation" [style.width.px]="width" [style.height.px]="height" [style.top.px]="top">
+        <ng-content></ng-content>
```
ashley-hunter: If the buttons are now on all pages with side navigation we should move the buttons and logic into the side navigation component

AlastairMcKee:
buttons are now part of the side navigation

PR #257, opened 2017-06-27
EL-2560 - Facet Refinements

ashley-hunter:
Making some refinements to Facet examples. Also chagin typeahead list facet to allow a "suggestions" list above the typeahead for quick selection


ashley-hunter:
Added focus style to make consistent with tags. The description in jira says 'blue' however the facets should now just match the tags in appearance, eg:

image

PR #260, opened 2017-06-28
EL-2613 Add Protractor tests to the project

PearseMcMurray:
Added Checkbox tests from EL-2582.
Where possible, NPM tasks have been converted to Grunt tasks.
Experimented with the grunt-concurrent and grunt-parallel plugins to replace the 'concurrently' package but neither had the same functionality.
Protractor executed using the grunt-protractor-runner plugin to make use of its webdriverManagerUpdate option which automatically keeps the Selenium binaries up to date.


e2e/.editorconfig
```diff
@@ -0,0 +1,19 @@
+# http://editorconfig.org
```
ashley-hunter: Should probably remove this file. We may want an editorconfig at some point but it will be in the root of the repository and we would need to remove all the options to ensure they comply with our other linting tools.

e2e/pages/app/app.module.ts
```diff
@@ -0,0 +1,41 @@
...
+import { AppComponent } from './app.component';
+import { CheckboxTestPageComponent }  from './checkbox/checkbox.testpage.component';
+import { CheckboxModule } from '../../../dist/lib/index.js';
+
+const ROUTES: Routes = [
+  {
+    path: 'checkboxes',
+    component: CheckboxTestPageComponent
+  },
+  // {
```
ashley-hunter: Can remove this commented out code

e2e/tests/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
```
ashley-hunter: We have two tsconfig files, one in pages and one in tests. They are practically identical (only difference is noImplicitAny). We could move the tsconfig file directly under the e2e folder and use the same one for both. In the merged one check to see if we can set noImplicyAny to `true` which is preferred but if not then set to `false`.

e2e/tslint.json
```diff
@@ -0,0 +1,60 @@
+{
```
ashley-hunter: We can remove this file now as the root folder has a tslint.json file

PR #261, opened 2017-06-29
EL-2627 - Fixing plunker

ashley-hunter:
Changed ngx-bootstrap urls to use a prebuilt bundle instead


PearseMcMurray:
Looks good.

PR #262, opened 2017-06-29
EL-2614 - Number Picker - Making QA changes

ashley-hunter:
1) Fixing typo
2) Using arrows now consistent with scrolling
3) Before disabled had to be set by either the html attribute `disabled="disabled"` or angular input `[disabled]="true"`. Made it more lenient to allow `disabled`, `disabled=""` and `disabled="true"`


PearseMcMurray:
Looks good.

PR #264, opened 2017-06-30
S 61/el 2606 spark chart theme colors

chris-brennan:
Added default behavior to account for user mistakes when specifying barColor and trackColor

### DIFF
--- a/src/components/spark/spark.component.less
+++ b/src/components/spark/spark.component.less
@@ -2,7 +2,7 @@ ux-spark {
     display: block;
 
     .ux-spark {
-        background: rgba(0, 167, 162, 0.2);
+        background-color: fade(@primary, 20%);
         overflow: hidden;
     }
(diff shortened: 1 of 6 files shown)


src/components/spark/spark.component.less
```diff
@@ -2,10 +2,12 @@ ux-spark {
     display: block;
 
     .ux-spark {
+        background: rgba(0, 167, 162, 0.2);
```
ashley-hunter: Change to use `background-color` and use the `@primary` variable here too. To set the alpha value of a variable less has some useful function, I believe the `fade`/`fadeIn`/`fadeOut` functions will be the ones to look at

chris-brennan:
Changed less to use @primary variable and applied fade function.  Thanks Ashley - that's awesome, I didn't know you could do this!

PR #265, opened 2017-06-30
EL-2614 - Adding new facet features

ashley-hunter:
1) Facet typeahead list and facet checklist examples are now sorted alphabetically
2) Typeahead now shows shows more items
3) Typeahead is not aligned with the checkboxes


PearseMcMurray:
Looks good.

PR #266, opened 2017-07-03
EL-2609 - Media Player Component

ashley-hunter:
- Adding media player component
- Added Frame Extraction Service and Audio Service
- Updated webpack to v3 as it sped up our build by ~1 minute,

**Note** - The stock video used doesn't have any audio - the sound isn't broken. We may change clips in future, most public domain stock videos clips don't seem to have audio.


src/components/media-player/media-player.component.less
```diff
@@ -0,0 +1,111 @@
...
+                    transform: scale(1.2);
+                }
+            }
+        }
+    }
+
+    .audio-player {
+        display: flex;
+        flex-direction: column;
+        background-color: @white;
```
donna-marijne: Possibly @media-player-* variables to be set up for these colors?

ashley-hunter:
- Fixed typo
- Moved to use @mediaplayer variables
- Moved the media player api to a service to it can be injected into any extensions

Regarding the SVG in the template, I did it this way to allow us to style it using css as if i put it in a separate file and use an img tag we can't style it with css. Perhaps a future change may be to move it to a replaceable template as you mentioned, however there are some other components such as facets and filters that are doing the same thing so perhaps a future ticket to investigate and implement a consistent approach throughout might be a good idea.

PR #274, opened 2017-07-06
El-2640 - Layout Switcher

ashley-hunter:
Migrating layout switcher component. Simplified the directive quite a lot, now each layout is part of a `group`. A group can have multiple layouts (if multiple ones in the same group, you would define a `minWidth` and `maxWidth` on each that requires it). 

You would then tell the `layoutSwitcher` which group to display and it will show the correct layout based on the group name and size constraints.


ashley-hunter:
Issue should now be fixed

PR #302, opened 2017-07-31
EL-2710 Defined new ux-aspects-quantum-build Docker image, containing N…

PearseMcMurray:
…PM, Ruby and the Sass gem, to be used to build the quantum-ux-aspects project.

Modified the docker_image_build() and docker_image_run() utility functions. Where previously the name and ID of the ux-aspects-build Docker image were hard-coded within the functions, these values are now passed as parameters. This change means there is no need to create duplicates of these functions for use with the new ux-aspects-quantum-build Docker image.

Added functions to build the Quantum projects using the new Docker image and then to update the SEPG Micro Focus repository with the new stylebook.


donna-marijne:
Hi Pearse - I don't think this belongs in the UX Aspects repository. Is it possible to use the quantum repository and the CAF jenkins server? Thanks.

PearseMcMurray:
In principle, yes, the new scripts and Docker image can be run on a different server. However, will we want to link to up-to-date builds of UXAspects rather than the latest NPM release? If so, I can't see how to do this without building UXAspects alongside Quantum on the new server.
Also, this will require a new job on the CAF server.

donna-marijne:
Yes, we will want a current build of UX Aspects. However, updates to this site are to be triggered from commits to the quantum repository, not UX Aspects, so we can clone the open source repo as required.

It makes sense to me that this should be a separate job on the CAF server. Is there any reason other than convenience to build it into the UX Aspects unit test job? Is it difficult to make changes to the CAF jenkins server compared to release engineering?

PearseMcMurray:
I have been talking to Dermot about how to use the CAF server. Jobs can be created by developers but the job must use a POM file to specify the steps (pulling of source code, compilation, writing to the SEPG repository) instead of only the Bash scripts I have been using up to now.

It would take some more time to write the POM file. Other than that, I can see no reason not to do this. If you want to use the code I have a written as a temporary measure until this work is completed I can create a smaller change set, as a lot of the current changes in the branch were made to avoid duplication of functions. If the code to build the Quantum repository is ultimately to be moved to a different repository then those changes become irrelevant.

donna-marijne:
Okay that sounds good to me. I think that ultimately this will be more efficient that waiting for release engineering. (I notice that the NPM fields were never updated in our release build task.)

Unfortunately it's too late in the sprint to make changes to the UX Aspects build scripts, so if it's okay with you I'll close this branch and we can work on a POM file and any other changes to automate this during the next sprint.

EL-2710 can be accepted on the basis of the site that you have manually set up.

PR #318, opened 2017-08-10
EL-2688 - Making changed to allow focus effect on sliders

AlastairMcKee:
Changed required to allow focus on the thumb button


ashley-hunter:
Is this to fix this?

image

If so I don't think these changes are required. I think you should be able to use the :active state on the button to turn the thumb blue when the mouse is dragging it, eg:

AlastairMcKee:
It was yes, I didn't think of that. Thanks

PR #325, opened 2017-08-15
Making changes to fix Quantum issue with Firefox

AlastairMcKee:
Firefox is not working with the :active pseudo class


src/components/slider/slider.component.ts
```diff
@@ -32,6 +32,8 @@ export class SliderComponent implements OnInit, AfterViewInit, DoCheck, OnDestro
     sliderThumb = SliderThumb;
     sliderTickType = SliderTickType;
     sliderThumbEvent = SliderThumbEvent;
+    lowerDragging = false;
```
ashley-hunter: Shouldn't need to add these variables. The dragging status should already be available in the thumbs object. eg. `thumbs.lower.drag` & `thumbs.upper.drag`

AlastairMcKee:
That worked, thanks

PR #329, opened 2017-08-17
Angular 4 Color Service

awrzesinska:
Adding documentation for color service and adding parse method to ThemeColor


docs/app/pages/components/sections/utilities/color-service/color-service.component.html
```diff
@@ -0,0 +1,229 @@
...
+</div>
+
+<blockquote>
+    <p><strong>Note</strong>: Refer to the <a routerLink="/css/color-palette">Color Palette</a> page for details about available
+        colors.</p>
+</blockquote>
+
+<p>Below is an example of how the <code>ColorService</code> can be used:</p>
+
+<uxd-snippet language="javascript" [content]="snippets.compiled.appTs"></uxd-snippet>
```
donna-marijne: Could I suggest a cut-down version of this, for documentation purposes only?

````javascript
import { ColorService, ThemeColor } from 'ux-aspects';
import { Component } from '@angular/core';

@Component({
    selector: 'app',
    templateUrl: './src/app.component.html',
    styleUrls: ['./src/app.component.css']
})
export class AppComponent {

    // ...

    constructor(public colorService: ColorService) {

        this.donutChartColors = [
            {
                backgroundColor: [
                    colorService.getColor('chart1').toRgb(),
                    colorService.getColor('chart2').toRgb(),
                    colorService.getColor('chart3').toRgb(),
                ],
                hoverBackgroundColor: [
                    colorService.getColor('chart1').setAlpha(0.3).toRgba(),
                    colorService.getColor('chart2').setAlpha(0.3).toRgba(),
                    colorService.getColor('chart3').setAlpha(0.3).toRgba(),
                ]
            }
        ];
    }
}
````

I think this would highlight the usage of the color service and the relevant ThemeColor methods.

File: docs/app/data/components-page.json [hunk 2]
```diff
@@ -1229,8 +1229,14 @@
                 },
                 {
                     "title": "Color Service",
+                    "component": "ComponentsColorServiceComponent",
+                    "version": "Angular"
+                },
```
donna-marijne: I forgot to mention one documentation feature that we have added for Angular 4. Could you add a "usage" property here, listing the module required for the color service?

````javascript
                        {
                            "title": "Module",
                            "content": "ColorServiceModule"
                        }
````

**Comment by awrzesinska** [comment]
Sorry for replying so late, I didn't notice the email. I should have these
changes ready in the morning.

Thanks for the feedback :)


Ola

File: docs/app/pages/components/sections/utilities/color-service/snippets/app.ts [hunk 3]
```diff
@@ -0,0 +1,165 @@
...
+                            return previousValue + currentValue;
+                        });
+
+                        // get the value of the current segment
+                        let segmentValue = dataset.data[item.index];
+
+                        // calculate the percentage of the current segment compared to the total
+                        let precentage = Math.round(((segmentValue / total) * 100));
+
+                        return `${ precentage }%, Sales ${ item.index + 1 }`;
```
donna-marijne: Let's cut some of this out, since we don't need to demo donut chart features.

````javascript
label: (item: Chart.ChartTooltipItem, data: any) => {
  return `Sales ${ item.index + 1 }`;
}
````

File: docs/app/pages/components/sections/utilities/color-service/snippets/app.ts [hunk 4]
```diff
@@ -0,0 +1,165 @@
...
+                    }
+                },
+                backgroundColor: tooltipBackgroundColor,
+                cornerRadius: 0,
+                displayColors: false
+            } as any
+        };
+
+        this.donutChartColors = [
```
donna-marijne: The example will update the colors dynamically, so it won't be necessary to initialise them here.

docs/app/pages/components/sections/utilities/color-service/snippets/app.ts
```diff
@@ -0,0 +1,165 @@
...
+                    index = 2;
+                    if (customized) {
+                        this.sales3 = '';
+                    }
+              }
+              
+              let newColor = ThemeColor.parse(color);
+              this.donutChartColors[0].backgroundColor[index] = color;
+              this.donutChartColors[0].hoverBackgroundColor[index] = newColor.setAlpha(0.3).toRgba();
+              this.donutChartColors = this.donutChartColors.slice();
```
donna-marijne: I appreciate that you have taken the time to update the colors individually. When it comes to efficiency in javascript/typescript, there is another thing to consider: the amount of code that the client has to download. Most browsers support array functions like map, reduce, find and so on, which we can use to reduce the amount of code written. (We use polyfills, which inject these functions for browsers which do not support them natively.)

Here are some example functions that might be of use:
````javascript
    colorNamesUpdated() {
        this.salesColorValues = this.salesColorNames.map(colorName => this.colorService.getColor(colorName).toHex());
        this.colorValuesUpdated();
    }
    
    colorValuesUpdated() {
        const chartColors = this.salesColorValues.slice();
        const chartHoverColors = this.salesColorValues.map(color => ThemeColor.parse(color).setAlpha(0.3).toRgba());
        this.donutChartColors = [{
            backgroundColor: chartColors,
            hoverBackgroundColor: chartHoverColors
        }];
    }
````

For these I used two new arrays, which are 2-way bound to the ux-select and input fields. Since the data is stored in a manner which is easily bound to the UI, it is a lot easier to keep everything in sync. In this case, the valueChange/ngModelChange events are used to trigger colorNamesUpdated/colorValuesUpdated, and parameters are not needed because the value is already updated by the Angular binding.

PR #331, opened 2017-08-21
Changes for custom class in integrated date picker

AlastairMcKee:
I have not added any examples to this to UX Aspects


AlastairMcKee:
Thanks Ashley, made those changes and tested it still works

PR #332, opened 2017-08-22
EL-2747 (e2e transpiled files should not be output alongside source f…

PearseMcMurray:
…iles)

.js and .js.map files now created in e2e/dist.
Unneeded lines removed from .gitignore.


PearseMcMurray:
Change made.

PR #338, opened 2017-08-29
Making changes to package.json for docs package

ashley-hunter:
Modified package.json file for docs package. 

Ideally this would probably have its own package.json file in the docs folder, however our docs are currently referencing our component scripts in the root/src folder so if i did this the src folder wouldn't be included and cause errors. Likely something we could implement later.

Also included the config/loaders folder to package to include our custom snippet loader. I know there was some discussion about making this a separate package but I don't think this is done yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "ux-aspects",
+  "name": "ux-aspects-docs",
   "version": "1.3.3",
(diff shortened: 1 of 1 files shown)


donna-marijne:
One other thing I just remembered, let's use the @ux-aspects scope for this.

ashley-hunter:
Adding scope to package name

PR #342, opened 2017-08-31
Service for data JSON files

donna-marijne:
Added methods to AppConfiguration to return the contents of the data/*.json files, and update them. This is used in the new documentation site to inject custom site layout.

https://jira.autonomy.com/browse/EL-2759

### DIFF
--- a/docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
+++ b/docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
@@ -8,6 +8,7 @@ import { ISearchResult } from '../../interfaces/ISearch';
 import { ISection } from '../../interfaces/ISection';
 import { NavigationService } from '../../services/navigation/navigation.service';
 import { Version, VersionService } from '../../services/version/version.service';
+import { AppConfiguration } from '../../services/app-configuration/app-configuration.service';
 
 const QUERY_MIN_CHARS = 3;
 const MAX_HISTORY = 5;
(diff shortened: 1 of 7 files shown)


docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
```diff
@@ -34,46 +35,35 @@ export class NavigationBarSearchComponent {
...
-    constructor(private router: Router, private navigation: NavigationService, private versionService: VersionService) {
-
-        // get version
-        this.versionService.version.subscribe((value: Version) => {
-            this.data = [];
-            this.data = this.data.concat(this.getSearchResults(componentsData));
-            this.data = this.data.concat(this.getSearchResults(cssData));
-            this.data = this.data.concat(this.getSearchResults(chartsData));
-        });
+    constructor(private router: Router, private navigation: NavigationService, private versionService: VersionService, private _appConfig: AppConfiguration) {
```
chris-brennan: Should private variables be prefixed with an underscore. i.e. `private _router: Router` ?

docs/app/pages/customize/customize.component.ts
```diff
@@ -11,13 +12,13 @@ export class CustomizePageComponent {
     sections: ICustomisePageSection[];
     variables: any;
 
-    constructor(private lessService: LessService) {
+    constructor(private lessService: LessService, private _appConfig: AppConfiguration) {
```
chris-brennan: underscore prefix?

docs/app/services/app-configuration/app-configuration.service.ts
```diff
@@ -3,19 +3,39 @@ import { Injectable } from '@angular/core';
...
     }
 
     private getConfig() {
         switch (process.env.ENV) {
             case 'development':
-                return require('../../data/config.dev.json');
+                return this._data['config.dev'];
         }
-        return require('../../data/config.json');
+        return this._data['config'];
```
chris-brennan: I'm not sure but could this use dot notation? Or is bracket notation preferred here?

docs/app/components/navigation-bar-search/navigation-bar-search.component.ts
donna-marijne: Yes. This is what I used. `private _appConfig: AppConfiguration`

docs/app/services/app-configuration/app-configuration.service.ts
donna-marijne: It's a key-value table, so dot is not preferred. Most keys include a period or hyphen so using property syntax would be impossible.

chris-brennan: OK I didn't know that, thanks.

chris-brennan:
looks good to me

PR #345, opened 2017-09-06
S 66/el 2772 plunker code pen examples

chris-brennan:
Changes made to the Plunker interface and service to allow the templates property to supply index.html, config.js and main.ts files to the Plunker example.  CodePen service updated to push a reference of the quantum-ux-aspects stylesheet into the array when 'secondaryAssetsUrl' is provided in config.json and config-dev.json.


chris-brennan:
This can be closed since it was easier to start again than to fix the issues outlined here.

PR #353, opened 2017-09-15
S 67/el 2777

AlastairMcKee:
Updated the color services to allow new themes to be used, the Micro Focus theme has been exported to allow it to be easily used. New themes will require the css to be provided. 

A default theme is provided but not exported as this will always be included and is included along with any custom themes to ensure legacy support. 

Also made a small change to  no longer require case sensitivity when specifying a color.


src/ng1/services/colorService/colorService.service.js
```diff
@@ -2,46 +2,61 @@ export default function $colorService() {
```
donna-marijne: Please reformat this file to consistently use spaces for indentation rather than tabs.

src/ng1/services/colorService/colorService.service.js
```diff
@@ -2,46 +2,61 @@ export default function $colorService() {
...
-				'<div class="partition14-color"></div>'+
-				'<div class="social-chart-node-color"></div>'+
-				'<div class="social-chart-edge-color"></div>';
-
-	var element = document.createElement('div');
-	element.className = 'color-chart';
-	element.innerHTML = html;
-
-	document.body.appendChild(element);
+	const defaultTheme = [
```
donna-marijne: I was thinking about this, and it would be good to separate it out of the code. Could we add `keppel-colors.json` and `micro-focus-colors.json` to the data directory? Then expose them in an object which is exported:

````
const ColorSets = {
  keppel: require('path-to-json-file'), //etc...
}
````

File: src/ng1/services/colorService/colorService.service.js [hunk 3]
```diff
@@ -59,47 +74,27 @@ export default function $colorService() {
...
-		partition13: getColorValue('partition13'),
-		partition14: getColorValue('partition14'),
-		'social-chart-node': getColorValue('social-chart-node'),
-		'social-chart-edge': getColorValue('social-chart-edge')
-   	};
-
-   	element.parentNode.removeChild(element);
+	var colors = {};
+
+	function getColors() {
```
donna-marijne: Any reason that getColors and setColors are separate functions?

src/services/color/color.service.ts
```diff
@@ -8,69 +8,61 @@ import {
...
-    '<div class="chart1-color"></div>' +
-    '<div class="chart2-color"></div>' +
-    '<div class="chart3-color"></div>' +
-    '<div class="chart4-color"></div>' +
-    '<div class="chart5-color"></div>' +
-    '<div class="chart6-color"></div>' +
-    '<div class="ok-color"></div>' +
-    '<div class="warning-color"></div>' +
-    '<div class="critical-color"></div>';
+    private _defaultTheme = [
```
donna-marijne: Same as the AngularJS version, let's pull these from a common file and make sure that the objects are exposed so that developers can assign whichever color set that they prefer.

(Also I was thinking that "theme" might not be a good term to use here, since we use theme to describe the full set of styling changes, rather than just colors.)

src/services/color/color.service.ts
```diff
@@ -8,69 +8,61 @@ import {
...
+        'chart4',
+        'chart5',
+        'chart6',
+        'ok',
+        'warning',
+        'critical'
+    ];
+
+
+    private _html: string = '';
```
donna-marijne: I don't think this needs to be an instance variable.

src/services/color/color.service.ts
```diff
@@ -8,69 +8,61 @@ import {
...
 
     private _element: HTMLElement;
-    private _colors: any;
+    private _colors: any = {};
```
donna-marijne: Specify ThemeColor type here?

src/services/color/color.service.ts
```diff
@@ -8,69 +8,61 @@ import {
...
+
+    private _theme: string[] = this._defaultTheme;
```
donna-marijne: Let's establish a proper type for this, even if it is currently just a list of strings. We might want to add more stuff later, I'm thinking that it would be more efficient to precompile the colors in by parsing SCSS, for example.

src/services/color/color.service.ts
```diff
@@ -91,7 +83,12 @@ export class ColorService {
     }
 
     getColor(color: ColorIdentifier): ThemeColor {
-        return this._colors[color];
+        return this._colors[color.toLowerCase()];
+    }
+
+    setTheme(theme: string[]) {
```
donna-marijne: It would be good to use get/set syntax here, so that the theme/color set can be read as well as assigned.

src/services/color/color.service.ts
```diff
@@ -91,7 +83,12 @@ export class ColorService {
     }
 
     getColor(color: ColorIdentifier): ThemeColor {
```
donna-marijne: Change ColorIdentifier to string and remove the definition below.

src/services/color/color.service.ts
```diff
@@ -91,7 +83,12 @@ export class ColorService {
...
-        return this._colors[color];
+        return this._colors[color.toLowerCase()];
+    }
+
+    setTheme(theme: string[]) {
+        this._theme.push.apply(this._theme, theme);
```
donna-marijne: I think you can just assign this.

AlastairMcKee: The reason I did it this way was so that the default colors would still be included when using another theme, for example in preexisting applications which switch to the Micro Focus theme we don't want to break uses of old colors throughout the application

src/services/color/color.service.ts
AlastairMcKee: This has been documented in UX-Aspects so could cause a breaking change if we remove it

donna-marijne: You can change the type to string without breaking anything. String is a superset of ColorIdentifier.

AlastairMcKee: Yes I changed the type, we just cant remove the definition

src/services/color/color.service.ts
donna-marijne: We don't want the old theme colors to appear after we have installed the Quantum stylesheet and set up the new color set though. Then we would have HPE green appearing wherever 'primary' was specified.

How about extending the color set data to be a map of [color name] -> [class name]? Then you can add in all of the UX Aspects color names and specify their rebranded color, while still allowing the new color names to be used in new code.

````
  "primary": "brand-blue",
  "chart1": "emerald",
````

Or let me know if you have any other ideas.

Comment (AlastairMcKee) [reply src/services/color/color.service.ts]: Yes that way sounds good, my thinking was that the variables file would be updated and mapped to the Micro Focus colors therefore primary would be the equivalent of brand blue, in the documentation we would encourage the use of the Micro Focus naming convention with the variables file there as a fallback. I think your approach seem cleaner however.

**Comment by AlastairMcKee** [comment]
Thanks Donna, I have made the requested changes, I will updated the documentation to reflect the new changes when this has been approved

File: src/ng1/services/colorService/colorService.service.js [hunk 11]
```diff
@@ -1,51 +1,39 @@
...
-	element.innerHTML = html;
-
-	document.body.appendChild(element);
+	function setColors() {
+
+		html = '';
+
+		for (let key in colorSet) {
+            html += '<div class="' + colorSet[key] + '-color"></div>';
+        }
```
donna-marijne: Indentation is still a bit wonky here, as you can see on GitHub. This is because the file incorrectly uses tabs instead of spaces.

AlastairMcKee:
Thanks Donna, the indentation should now be correct and i have corrected the spelling of coper to match the quantum change

AlastairMcKee:
Made changes to support primary-bg style class names rather than primary-color

AlastairMcKee:
Made changes to allow hex values to be added to fix the issue mentioned in the quantum PR

src/styles/colors.less
```diff
@@ -1,137 +1,136 @@
-.primary-color {
+.primary-bg {
```
donna-marijne: Ah sorry, I meant this change for Quantum since we were creating new classes there. These might be used by other applications, so let's just stick with the current suffix.

AlastairMcKee:
Thanks, those changes have reverted now

PR #366, opened 2017-09-28
S 68/el 2793 ui updates

chris-brennan:
- Fixed page load column error for the features
- Updated search so that it functions on Micro Focus site as well
- Updated footer logo to handle the text using CSS


docs/app/components/side-navigation/side-navigation.component.html
```diff
@@ -5,7 +5,7 @@
                 [ngModel]="versionService.version | async"
                 (ngModelChange)="versionService.setVersion($event)"
                 [btnRadio]="Version.AngularJS">AngularJS</button>
-            <button type="button" class="btn button-toggle-accent"
+            <button type="button" class="btn button-toggle-accent primary-override"
```
ashley-hunter: Dont think this class is needed here as you should be able to override styles by doing the following:

```css
.side-navigation-container .side-navigation .btn-group .button-toggle-accent {
    // styles go here
}
```

docs/app/data/footer-navigation.json
```diff
@@ -9,7 +9,8 @@
             "link": "/changelog"
         }]
     }],
-    "logo": "assets/img/UX-Aspects-Logo.png",
+    "logo": "assets/img/UX-graphic-small.png",
```
ashley-hunter: Perhaps change this to:

```json
"logo": {
    "image": "assets/img/UX-graphic-small.png",
    "title": "UX Aspects"
}
```

This way the two logo properties are grouped rather than naming one logo and one logoTitle. It also means in future if we were to, for example, make it go to a link when clicked we could just add a link property under logo rather than creating a new `logoLink` property,

docs/app/components/page-footer/page-footer.component.less
```diff
@@ -40,9 +40,20 @@
...
+    }
+
     .page-footer-brand-logo {
-        display: block;
-        margin: 0 auto;
+        margin-right: 5px
+    }
+
+    .page-footer-brand-logo-title {
+        color: #ffffff;
```
donna-marijne: Please use the variables as much as possible - it will make it easier to override for the Micro Focus site (which will be black rather than white here, for example).

docs/app/components/side-navigation/side-navigation.component.html
donna-marijne: Sorry, this wasn't what I meant for this change. Please bind the button class (.btn-toggle-accent) to an input property on this component. E.g. `[ngClass]="angularButtonClass"`. For the sake of consistency, do this on both buttons. The reason for this is that there are a lot of states and properties defined for buttons in the MF style, and overriding the style on the existing class would involve duplicating a lot style definitions from the quantum stylesheet.

If we used Sass then potentially we could mix in the quantum definitions, but for now I believe that this is the cleanest approach.

chris-brennan: Sorry I meant to remove this.  I have overridden the styles in ux-aspects-microfocus like you suggested.

chris-brennan:
- Removed unused class
- Updated to group logo properties
- Updated to use variables
- Button classes controlled by input property for version toggle

docs/app/components/side-navigation/side-navigation.component.html
```diff
@@ -1,11 +1,13 @@
 <div class="side-navigation-container" #container>
     <div class="side-navigation" [style.width.px]="width" [style.height.px]="height" [style.top.px]="top">
          <div class="btn-group version-toggle p-b-sm">
-            <button type="button" class="btn button-toggle-primary"
+            <button type="button" class="btn"
+                [ngClass]="angularButtonClass"
```
donna-marijne: Shouldn't this be `angularJsButtonClass`?

chris-brennan:
updated to point to angularJsButtonClass

chris-brennan:
Resolved conflicts

configs/webpack.docs.dev.config.js
```diff
@@ -219,7 +219,12 @@ var docsConfig = {
             'process.env': {
                 'ENV': '"development"'
             }
+<<<<<<< HEAD
```
donna-marijne: This is a merge file - please resolve. If you have any trouble with merges, let me know and I can describe the workflow.

chris-brennan:
resolved conflicts

donna-marijne:
@ashley-hunter can you review this? Chris cannot reproduce the above error.

ashley-hunter:
Just spoke to him. This branch is outdated and does not have the latest webpack configs, he is rebasing it now as these issues have already been fixed in develop.

donna-marijne:
Thanks Ashley.

chris-brennan:
not fixed

chris-brennan:
That should be it fixed now, thanks.

PR #368, opened 2017-10-02
EL-2803 - Small documentation change for DTD fix

AlastairMcKee:
Unnecessary divs were messing up the Micro Focus documentation site after the DTD change. Removing them has no impact on UX Aspects


AlastairMcKee:
Sorry Donna, these appeared to be causing some additional height to the multiple select dropdowns which would shift the text down slightly messing up alignment with the single select dropdowns, on further investigation this was actually down to the form-group div. This PR can be closed

donna-marijne:
No problem, thanks for double checking.

PR #369, opened 2017-10-02
EL-2814 - Build Improvements

ashley-hunter:
Library build output now follows the Angular Package Format. This will allow it to work with Angular cli and AOT applications. This should also fix a lot of the issue with VS Code not giving intellisense for interfaces etc...

The output structure has now changed quite a bit regarding the library, providing a range of different bundle formats.


src/services/color/color.service.ts
```diff
@@ -202,8 +202,124 @@ export class ThemeColor {
 }
 
 export const colorSets = {
-    keppel: require('../../data/keppel-colors.json'),
-    microFocus: require('../../data/micro-focus-colors.json')
+    keppel: {
```
ashley-hunter: Note - had to change this as we can no longer use requires in our components as this is not supported by angular compiler. Requires will still work elsewhere in our docs as they are built by webpack.

ashley-hunter:
I fixed that problem yesterday as part of [https://github.com/UXAspects/UXAspects/pull/371](https://github.com/UXAspects/UXAspects/pull/371). I have rebased the branch on to develop so it should now have this fix included.

donna-marijne:
I see that it was merged but I'm not a fan of pasting 3rd party code into our library. Can we reconsider that change?

ashley-hunter:
Made a change - added the import to ux-aspects.less. Have to give a relative path as using tilde only works when building with webpack which our build process does not.

donna-marijne:
I think that you need to use `@import (inline)` for the tree CSS. Otherwise we introduce a new dependency for our CSS file.

ashley-hunter:
We have a grunt task that calls `inline-less.js` which automatically makes all our imports inline.

donna-marijne:
That's true, I changed that to separate variables.less, so I'll make the change.

PR #372, opened 2017-10-03
S 68/el 2794 less variables

donna-marijne:
Modified the Less build process to keep variables.less separate in assets. (This turned out not to be necessary for this ticket, but could be useful for other projects, e.g. parsing out colors like the customize page.)
Added variables for all library images, allowing them to be selectively replaced.
Removed workarounds for opacity in CSS.

https://jira.autonomy.com/browse/EL-2794


donna-marijne:
Closing since further changes are required.

PR #380, opened 2017-10-10
EL-2814 - Maintaining backwards compatibility with file path changes

ashley-hunter:
Added a script to the grunt build task to place files in the locations before the package structure changed to import the new version of the file maintaining backwards compatibility


ashley-hunter:
Ah good point, forgot about those. They should be added in now. Thanks

PR #390, opened 2017-10-13
EL-2827 - Wizard component

ashley-hunter:
Jira: https://jira.autonomy.com/browse/EL-2827

Created Angular 4 wizard control


ashley-hunter:
Example code no longer uses ngFor. Thanks

PR #394, opened 2017-10-17
Fixing issue with tooltip being left after element is removed

AlastairMcKee:
This seems to be the only way we can reliably remove a specific tooltip. When the tooltip is applied to the body it gets an ID which matches the aria-describedBy attribute on the element so we can use this to remove the specific tooltip if it gets left behind.

### DIFF
--- a/src/ng1/directives/overflowTooltip/singleLineOverflowTooltip.directive.js
+++ b/src/ng1/directives/overflowTooltip/singleLineOverflowTooltip.directive.js
@@ -40,7 +40,7 @@ export default function singleLineOverflowTooltip(safeTimeout, $window, $resize)
             //watch for changes to window size
             $window.addEventListener('resize', updateTooltipFn);
 
-            scope.$on('$destroy', function () {              
+            scope.$on('$destroy', function () {
                 element.tooltip('destroy');
                 $window.removeEventListener('resize', updateTooltipFn);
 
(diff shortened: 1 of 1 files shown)


src/ng1/directives/overflowTooltip/singleLineOverflowTooltip.directive.js
```diff
@@ -43,6 +43,16 @@ export default function singleLineOverflowTooltip(safeTimeout, $window, $resize)
             scope.$on('$destroy', function () {
```
ashley-hunter: Instead of using scope on $destroy bind to jquery's $destroy, this will fire when the element data is still available so `element.tooltip('destroy)` should work as expected. That should be the only change required to fix this issue.

AlastairMcKee:
That works great, thanks

src/ng1/directives/overflowTooltip/singleLineOverflowTooltip.directive.js
```diff
@@ -40,7 +40,7 @@ export default function singleLineOverflowTooltip(safeTimeout, $window, $resize)
             //watch for changes to window size
             $window.addEventListener('resize', updateTooltipFn);
 
-            scope.$on('$destroy', function () {
+            element.bind('$destroy', function () {
```
ashley-hunter: Great, just one minor change. Can you change this to use `.on` as `.bind` is a deprecated function

AlastairMcKee:
Made that change

PR #397, opened 2017-10-18
S 69/el 2810

AlastairMcKee:
jscrollpane should now resize correctly when popovers and dropdowns are toggled. See the search builder, popovers would be cut off when added close to the bottom of the modal.

Jira: https://jira.autonomy.com/browse/EL-2810

### DIFF
--- a/docs/app/data/components-page.json
+++ b/docs/app/data/components-page.json
@@ -836,16 +836,6 @@
                     "title": "Search Builder",
                     "component": "ComponentsSearchBuilderNg1Component",
                     "version": "AngularJS"
-                },
-                {
-                    "title": "Search History",
-                    "component": "ComponentsSearchHistoryNg1Component",
-                    "version": "AngularJS"
-                },
-                {
-                    "title": "Search Toolbar",
-                    "component": "ComponentsSearchToolbarNg1Component",
-                    "version": "AngularJS"
                 }
             ]
         },
--- a/docs/app/data/components-page.json
+++ b/docs/app/data/components-page.json
@@ -836,6 +836,16 @@
                     "title": "Search Builder",
                     "component": "ComponentsSearchBuilderNg1Component",
                     "version": "AngularJS"
+                },
+                {
+                    "title": "Search History",
+-                   "component": "ComponentsSearchHistoryNg1Component",
+                    "version": "AngularJS"
+-               },
+-               {
+                    "title": "Search Toolbar",
+-                   "component": "ComponentsSearchToolbarNg1Component",
+-                   "version": "AngularJS"
                 }
             ]
         },
--- a/docs/app/data/components-page.json
+++ b/docs/app/data/components-page.json
@@ -839,13 +839,13 @@
                 },
                 {
                     "title": "Search History",
--                   "component": "ComponentsSearchHistoryNg1Component",
+                    "component": "ComponentsSearchHistoryNg1Component",
                     "version": "AngularJS"
--               },
--               {
+                },
+                {
                     "title": "Search Toolbar",
--                   "component": "ComponentsSearchToolbarNg1Component",
--                   "version": "AngularJS"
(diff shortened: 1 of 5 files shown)


src/ng1/plugins/customscroll/jquery.jscrollpane.js
```diff
@@ -184,13 +199,13 @@ LICENSE-END
 					maintainAtBottom = settings.stickToBottom && isCloseToBottom();
 					maintainAtRight  = settings.stickToRight  && isCloseToRight();
 
-					hasContainingSpaceChanged = elem.innerWidth() + originalPaddingTotalWidth != paneWidth || elem.outerHeight() != paneHeight;
+					hasContainingSpaceChanged = elem.innerWidth() + originalPaddingTotalWidth != paneWidth || elem[0].outerHeight != paneHeight;
```
ashley-hunter: Unfortunately we won't be able to use this solution. `outerHeight` is not a valid property on an element, it is only present on the window object:

image

With this change here the `outerHeight` value will always be different than the paneHeight value as it will always be undefined.

AlastairMcKee:
Made change to update the sizes when a mutation has been observed, also replaced timeout with the delay time

docs/app/data/components-page.json
```diff
@@ -836,16 +836,6 @@
                     "title": "Search Builder",
                     "component": "ComponentsSearchBuilderNg1Component",
                     "version": "AngularJS"
-                },
-                {
```
ashley-hunter: Is this change expected? Should these sections be removed?

src/ng1/plugins/customscroll/jquery.jscrollpane.js
```diff
@@ -136,7 +136,7 @@ LICENSE-END
 											(parseInt(elem.css('paddingRight'), 10) || 0);
 			}
 
-			function initialise(s)
+			function initialise(s, mutationObserved)
```
ashley-hunter: Rename this variable to something like `forceResize` or something as the value of this variable will force it to recalculate the size and may have other uses than only when a mutation is observed.

AlastairMcKee:
Added sections back in and updated name. Thanks

src/ng1/plugins/customscroll/jquery.jscrollpane.js
```diff
@@ -108,6 +108,21 @@ LICENSE-END
...
+					return;
+				}
+				timer = setTimeout(function() {
+					if(settings) {
+						initialise(settings, true);
+					}
+					timer = null;	
+				}, settings.autoReinitialiseDelay);
+			});
+			observer.observe(elem[0], {attributes: true, childList: true, subtree: true});
```
ashley-hunter: We need to disconnect the mutation observer and cancel any pending timeout when the scrollpane is destroyed.

AlastairMcKee:
Added teardown code, observer should now disconnect and timeout cleared

src/ng1/plugins/customscroll/jquery.jscrollpane.js
```diff
@@ -108,6 +108,21 @@ LICENSE-END
 				}
 			};
 
+			// UX Aspects modification to reinitialise when popovers are shown	
+			var timer = null;
+			var observer = new MutationObserver(function() {
```
ashley-hunter: Apologies for the numerous comments, I have been doing a bit of performance investigating around this as it's such a widely used component.

I think we should introduce two new settings, one specifically for the debounce time (i realise i suggested we use the autoReinitialiseDelay, but upon further investigation this is often set to 0 which would make this quite performance heavy). The time should probably be defaulted to something like 500ms.

We should also add one to enable/disabled this feature all together for extra performance where this wont be needed at all.

Apart from that it looks good.

AlastairMcKee:
Thanks Ashley, added 2 new variables for this and documented them under the custom scrollbar documentation. This should improve performance

PR #398, opened 2017-10-18
El 2798 code pen plunker bugs2

chris-brennan:
Fixed QA issues


chris-brennan:
Jira: https://jira.autonomy.com/browse/EL-2798

docs/app/pages/components/sections/select/select/snippets/app.snippet.ts
```diff
@@ -48,18 +48,19 @@ export class AppComponent implements OnInit {
         this.pagingEnabled.subscribe((value) => {
             this.selected = null;
             this.dropdownOpen = false;
-            this.options = this.pagingEnabled.getValue() ?
-                this.loadOptionsCallback :
-                this.selectedDataSet();
+            this.options = this.pagingEnabled.getValue() ? this.loadOptionsCallback : this.selectedDataSet();
```
donna-marijne: What does this change accomplish?

docs/app/pages/components/sections/select/select/snippets/app.ts
```diff
@@ -49,24 +49,25 @@ export class AppComponent implements OnInit {
...
+        this.dataSets.strings = ['United States', 'United Kingdom', 'Afghanistan' /*...*/];
```
donna-marijne: Confused about this as well. You've removed all but 3 data items from the plunker example.

docs/app/pages/components/sections/select/select/snippets/app.html
```diff
@@ -1,5 +1,5 @@
-<div class="row">
-    <div class="col-sm-6 col-xs-12">
+<div class="row m-t-xl">
+    <div class="col-sm-6 col-xs-12 m-t-xl">
```
donna-marijne: Presumably this is to allow dropDirection="up", in which case the appropriate change is to add a CSS class with margin-top of 250px, matching the dropdown height.

docs/app/pages/components/sections/select/select/snippets/app.snippet.ts
chris-brennan: I don't think this accomplishes anything other than consistency in the layout of the code.  What I think has happened here is that I made changes in the flat select and rather than make them again individually, I copied the contents of the component over.  Sorry about that.

docs/app/pages/components/sections/select/select/snippets/app.ts
chris-brennan: The flat select plunker example only has 3 data items in it (for whatever reason that may be).  I thought that since it is the newest version that there may have been some discussion on this and I updated the rest to keep them consistent.  I'm guessing now that the full set is preferred - sorry I shouldn't have assumed this.

chris-brennan: Alastair has just explained that having three items in the dataset is only for the layout in the snippets.  I'll updated all plunker examples with the full dataset.  Thanks

docs/app/pages/components/sections/select/select/snippets/app.snippet.ts
donna-marijne: In most cases, wrapping is to avoid a horizontal scrollbar in the code snippet window. I noticed that this did add a scrollbar, although it was only for the sake of a few pixels.

chris-brennan:
Added class to allow for down direction up and added full dataset

PR #400, opened 2017-10-19
EL-2818 - Adding persistent data service and documentation

AlastairMcKee:
https://jira.autonomy.com/browse/EL-2818

I have added a service which will attempt to use localStorage before falling back to cookies, you can also choose to save between localStorage, cookies and sessionStorage (this will still fallback to cookies if localStorage fails). 
I have also updated our site to use this where localStorage was being used previously and have added documentation for this service and a note to search builder explaining localStorage issues.

### DIFF
--- a/docs/app/pages/components/sections/search/search-history-ng1/snippets/controller.js
+++ b/docs/app/pages/components/sections/search/search-history-ng1/snippets/controller.js
@@ -1,8 +1,8 @@
 angular.module('app').controller('SearchHistoryDemoCtrl', SearchHistoryDemoCtrl);
 
-SearchHistoryDemoCtrl.$inject = ['$scope', '$modal', '$persistentDataService'];
+SearchHistoryDemoCtrl.$inject = ['$scope', '$modal'];
 
-function SearchHistoryDemoCtrl($scope, $modal, $persistentDataService) {
+function SearchHistoryDemoCtrl($scope, $modal) {
     var vm = this;
 
     vm.scrollConfig = {
@@ -64,7 +64,7 @@ function SearchHistoryDemoCtrl($scope, $modal, $persistentDataService) {
     storeSearches();
 
     vm.getHistory = function () {
-        vm.searches = JSON.parse($persistentDataService.getData('mySearches'));
+        vm.searches = JSON.parse(localStorage.getData('mySearches'));
     };
 
     vm.openModal = function () {
@@ -81,7 +81,7 @@ function SearchHistoryDemoCtrl($scope, $modal, $persistentDataService) {
         modalInstance.result.then(function (result) {
             if (result === "cancel") return;
 
-            vm.searches = JSON.parse($persistentDataService.getData('mySearches'));
+            vm.searches = JSON.parse(localStorage.getData('mySearches'));
             vm.searches.unshift(result);
             vm.searches.pop();
 
@@ -96,6 +96,6 @@ function SearchHistoryDemoCtrl($scope, $modal, $persistentDataService) {
     };
 
     function storeSearches() {
-        $persistentDataService.setData('mySearches', JSON.stringify(vm.searches));
+        localStorage.setData('mySearches', JSON.stringify(vm.searches));
     }
 }
--- a/docs/app/pages/components/sections/search/search-history-ng1/snippets/controller.js
+++ b/docs/app/pages/components/sections/search/search-history-ng1/snippets/controller.js
@@ -1,8 +1,8 @@
 angular.module('app').controller('SearchHistoryDemoCtrl', SearchHistoryDemoCtrl);
 
-SearchHistoryDemoCtrl.$inject = ['$scope', '$modal'];
+SearchHistoryDemoCtrl.$inject = ['$scope', '$modal', '$persistentDataService'];
 
-function SearchHistoryDemoCtrl($scope, $modal) {
+function SearchHistoryDemoCtrl($scope, $modal, $persistentDataService) {
     var vm = this;
 
     vm.scrollConfig = {
@@ -64,7 +64,7 @@ function SearchHistoryDemoCtrl($scope, $modal) {
     storeSearches();
 
     vm.getHistory = function () {
-        vm.searches = JSON.parse(localStorage.getItem('mySearches'));
+        vm.searches = JSON.parse($persistentDataService.getData('mySearches'));
     };
 
     vm.openModal = function () {
@@ -81,7 +81,7 @@ function SearchHistoryDemoCtrl($scope, $modal) {
         modalInstance.result.then(function (result) {
             if (result === "cancel") return;
 
-            vm.searches = JSON.parse(localStorage.getItem('mySearches'));
+            vm.searches = JSON.parse($persistentDataService.getData('mySearches'));
             vm.searches.unshift(result);
             vm.searches.pop();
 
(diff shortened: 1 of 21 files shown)


src/services/persistent-data/persistent-data.service.ts
```diff
@@ -0,0 +1,48 @@
...
+            }
+            return null;
+        } else {
+            return sessionStorage.getItem(key);
+        }
+    }
+
+}
+
+export enum PersistentDataStorageType {
```
ashley-hunter: Currently this class isn't very extensible, for example we may in future want to add additional storage options (such as IndexedDB or FileAPI support) which could prove a bit tricky with all the if statements etc.. We also should be using a try..catch to most accurately determine whether or not local/session storage works or not.

Consider the following code which should be a bit more extensible if we want to add more storage types in future:

```typescript
import { Injectable } from '@angular/core';

@Injectable()
export class PersistentDataService {

    /**
     * Save the item in some form of persistent storage
     */
    setItem(key: string, value: string, storageType: PersistentDataStorageType = PersistentDataStorageType.LocalStorage): void {

        // check if storage is supported - otherwise fallback
        storageType = this.getStorageType(storageType);

        // peform the save on the desire storage target
        switch (storageType) {

            case PersistentDataStorageType.LocalStorage:
                this.setItemStorage(localStorage, key, value);
                break;

            case PersistentDataStorageType.SessionStorage:
                this.setItemStorage(sessionStorage, key, value);
                break;

            case PersistentDataStorageType.Cookie:
                this.setItemCookie(key, value);
                break;
        }
    }

    /**
     * Get a stored value from persistent storage
     */
    getItem(key: string, storageType: PersistentDataStorageType = PersistentDataStorageType.LocalStorage): string {

        // get the supported storage type
        storageType = this.getStorageType(storageType);

        switch (storageType) {

            case PersistentDataStorageType.LocalStorage:
                return this.getItemStorage(localStorage, key);

            case PersistentDataStorageType.SessionStorage:
                return this.getItemStorage(sessionStorage, key);

            case PersistentDataStorageType.Cookie:
                return this.getItemCookie(key);
        }
    }

    /**
     * Store value in either local or session storage
     */
    private setItemStorage(storage: Storage, key: string, value: string): void {
        storage.setItem(key, value);
    }

    /**
     * Save the value in a cookie 
     */
    private setItemCookie(key: string, value: string): void {
        document.cookie = key + '=' + value + '; path=/';
    }

    /**
     * Retrieve value in either local or session storage
     */
    private getItemStorage(storage: Storage, key: string): string {
        return storage.getItem(key);
    }

    /**
     * Retrieve the value in a cookie 
     */
    private getItemCookie(key: string): string {

        // get all the cookies for this site
        const cookies = document.cookie.split(';');

        // process the cookies into a from we can easily manage
        const match = cookies
            .map(cookie => ({ key: cookie.split('=')[0].trim(), value: cookie.split('=')[1].trim() }))
            .find(cookie => cookie.key === key);

        return match ? match.value : null;
    }

    /**
     * Determine which storage method can be used by the browser
     */
    private getStorageType(desiredType: PersistentDataStorageType): PersistentDataStorageType {

        // If local storage is desired then make sure we can use it - or fallback to cookies
        if (desiredType === PersistentDataStorageType.LocalStorage) {
            return this.getStorageSupported(localStorage) ? PersistentDataStorageType.LocalStorage : PersistentDataStorageType.Cookie;
        }

        // If session storage is desired then make sure we can use it - or fallback to cookies
        if (desiredType === PersistentDataStorageType.SessionStorage) {
            return this.getStorageSupported(sessionStorage) ? PersistentDataStorageType.LocalStorage : PersistentDataStorageType.Cookie;
        }

        return desiredType;
    }

    /**
     * Perform a test to see if the storage option is supported
     * @param storage Either local storage or session storage
     */
    private getStorageSupported(storage: Storage): boolean {

        // if the storage type is null or undefined then fall back to cookies
        if (!storage) {
            return false;
        }

        try {
            storage.setItem('ux-persistent-data-service', 'ux-persistent-data-service');
            storage.removeItem('ux-persistent-data-service');
            return true;
        } catch (err) {
            return false;
        }
    }

}

export enum PersistentDataStorageType {
    LocalStorage,
    Cookie,
    SessionStorage
}
```

> Note, I havent tested this so you would need to check that it works ok

AlastairMcKee:
Thanks Ashley, Made the requested changes and added removeItem and clear functions

docs/app/pages/components/sections/utilities/persistent-data-service/snippets/app.ts
```diff
@@ -9,15 +9,22 @@ export class AppComponent {
 
     searches = ['checkbox', 'toggle switch', 'tree grid'];
 
-    constructor(public persistentDataService: PersistentDataService) {
-        this.persistentDataService.setData(
+    constructor(public _persistentDataService: PersistentDataService) {
```
ashley-hunter: Make this variable private private

docs/app/pages/components/sections/search/search-history-ng1/snippets/controller.js
```diff
@@ -1,8 +1,8 @@
 angular.module('app').controller('SearchHistoryDemoCtrl', SearchHistoryDemoCtrl);
 
-SearchHistoryDemoCtrl.$inject = ['$scope', '$modal'];
+SearchHistoryDemoCtrl.$inject = ['$scope', '$modal', '$persistentDataService'];
```
ashley-hunter: Might be best to keep this example using localStorage as these changes break the codepen and for people only using Angular 1 and not a hybrid this code will not work.

docs/app/pages/components/sections/utilities/persistent-data-service/snippets/app.ts
```diff
@@ -0,0 +1,30 @@
...
+
+@Component({
+    selector: 'app',
+    templateUrl: './src/app.component.html'
+})
+export class AppComponent {
+
+    searches = ['checkbox', 'toggle switch', 'tree grid'];
+
+    constructor(public _persistentDataService: PersistentDataService) {
```
ashley-hunter: Make private

src/services/persistent-data/persistent-data.service.ts
```diff
@@ -0,0 +1,230 @@
+import { Injectable } from '@angular/core';
```
ashley-hunter: Changes are good. There are a few console.logs in there to remove. I have sent you an email as this would be a perfect use case for the Adapter design pattern here. Would just involve slight reorganizing the code (see my email)

AlastairMcKee:
Thanks Ashley, have added in the restructured code and tested, everything seems to be working as expected

docs/app/pages/components/sections/utilities/persistent-data-service/persistent-data-service.component.html
```diff
@@ -0,0 +1,54 @@
...
+            <th>Returns</th>
+            <th>Description</th>
+        </tr>
+        <tr>
+            <td>setItem</td>
+            <td>string, string, PersistentDataStorageType?</td>
+            <td></td>
+            <td>This method expects a key and value parameter as strings and can also accept a 
+                PersistentDataStorageType if you wish to set the storage type, 
+                otherwise it will follow defauly behaviour.</td>
```
ashley-hunter: Typo - 'default' might be good to mention what the default value is also

AlastairMcKee: Default behavior is mentioned above this

AlastairMcKee:
Thanks Ashley, that seems to be working correctly now and typos have been fixed

PR #415, opened 2017-10-26
EL-2830 - Dashboard Fixes

ashley-hunter:
https://jira.autonomy.com/browse/EL-2839 & https://jira.autonomy.com/browse/EL-2830

Fixing several issues:

- When widgets are dynamically added eg. using ngFor the height of the dashboard isnt recalculated
- When using ngFor to add widgets, they did not get their padding set
- Adding types to resize service subject
- Changing directive selector to follow naming conventions (kept old selector as well to prevent any breaking changes)


ashley-hunter:
Good spot, issue fixed!

PR #418, opened 2017-10-27
EL-2813 - Adding custom class to AngularJS TreeView

AlastairMcKee:
https://jira.autonomy.com/browse/EL-2831

Custom class can now be added to Angular JS tree view along with a demo of how it can be added dynamically.

Angular 4 version gives users full control of the templates so custom classes can already be added if required


src/ng1/directives/treeView/treeView.partial.html
```diff
@@ -1,4 +1,4 @@
-<div class="tree-node tree-node-content" data-nodrag ng-click="tv.select(this)" ng-class="{highlight : tv.isSelected(this)}" >
+<div class="tree-node tree-node-content {{node.class}}" data-nodrag ng-click="tv.select(this)" ng-class="{highlight : tv.isSelected(this)}" >
```
ashley-hunter: Change this to something like `className` rather that `class` as it is a registered javascript keyword.

docs/app/pages/components/sections/tree-view/tree-view-ng1/snippets/sample.html
```diff
@@ -1 +1,15 @@
-<tree-view data="vm.data" selected="vm.selected" add-item="vm.addItem" icons="vm.iconConfig" delete="vm.deleteFn" options="vm.treeOptions"></tree-view>
\ No newline at end of file
+<tree-view data="vm.data" selected="vm.selected" add-item="vm.addItem" icons="vm.iconConfig" delete="vm.deleteFn" options="vm.treeOptions"></tree-view>
+
+<div class="row uxd-customize-example">
```
ashley-hunter: I think we could probably remove this as a custom class won't be a feature that is used very often. Just might allow us to keep the sample code more simple and concise making it easier for teams to understand.

AlastairMcKee:
Thanks Ashley, made those changes

PR #425, opened 2017-11-01
MF-Theme-Feedback-09-10 - Updated spacing between buttons

chris-brennan:
- Jira: https://jira.autonomy.com/browse/EL-2821
- Updated buttons to use the same spacing as on EverythingUX, and to use a flex layout to address the additional white space between inline elements.

### DIFF
--- a/docs/app/pages/components/sections/buttons/dropdown-ng1/snippets/dropdown.html
+++ b/docs/app/pages/components/sections/buttons/dropdown-ng1/snippets/dropdown.html
@@ -1,6 +1,6 @@
 <p>Button dropdown open: <em ng-bind="vm.buttonOpen" class="ng-binding">false</em>. Split dropdown open: <em ng-bind="vm.splitOpen" class="ng-binding">false</em>.</p>
 
-<div class="m-t m-b">
+<div class="btn-container">
     <div class="btn-group" dropdown is-open="vm.buttonOpen">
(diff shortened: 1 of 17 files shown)


chris-brennan:
- Removed styles from documentation stylesheet.
- Added `.btn-container` class to UX Aspects stylesheet and included it in relevant (examples that have more than one button) code snippets, codepen examples, and html files

chris-brennan:
Hi Donna, I don't seem to be able to remove the `.btn-group` class. I think it is needed for positioning the dropdown menu relative to the button. If I replace it with the `btn-container` class, the dropdown will position relative to the container that contains the buttons.

donna-marijne:
Oh, I see. In that case, no changes are needed to this example, so you can revert the addition of .btn-container.

chris-brennan:
OK no probs.  But if I remove the 'btn-container' class then the buttons will display as inline-block and the spacing between each button will contain additional white space.  The 'btn-container' class fixes this by applying a flex layout to it as the parent.  What are your thoughts on this? Thanks

donna-marijne:
No, it's good as it is, I misread the code.

chris-brennan:
ah Ok, No probs at all.

chris-brennan:
yeah sure - I have updated dropdown-ng1 now

ashley-hunter:
Can this be merged, or is it waiting on further changes?

PR #427, opened 2017-11-02
EL-2846 Complete migration of Selenium tests to Protractor and remove…

PearseMcMurray:
… unneeded files and build steps


e2e/pages/app/app.module.ts
```diff
@@ -122,17 +143,32 @@ const ROUTES: Routes = [
...
+  }, {
     path: 'tags',
     component: TagsTestPageComponent
   }, {
     path: 'toggle-button',
     component: ToggleButtonTestPageComponent
   }, {
-    path: 'toggleswitches',
-    component: ToggleSwitchesTestPageComponent
+    path: 'toggle-button',
```
ashley-hunter: This route appears to be duplicated - i think this should be toggle switches instead

PR #441, opened 2017-11-08
Added polyfill library to codepen references

donna-marijne:
https://jira.autonomy.com/browse/EL-2838

### DIFF
--- a/docs/app/services/codepen/codepen.service.ts
+++ b/docs/app/services/codepen/codepen.service.ts
@@ -161,6 +161,7 @@ export class CodePenService {
 
     // Script files for CodePen to reference
     private CODEPEN_SCRIPTS = [
+        'https://unpkg.com/core-js@2.5.1/client/core.js',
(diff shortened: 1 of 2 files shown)


docs/app/services/codepen/codepen.service.ts
```diff
@@ -161,6 +161,7 @@ export class CodePenService {
 
     // Script files for CodePen to reference
     private CODEPEN_SCRIPTS = [
+        'https://unpkg.com/core-js@2.5.1/client/core.js',
```
ashley-hunter: This will fix the codepen, however teams consuming only the ux-aspects-ng1.js script file will still get this error in IE if they try to use this feature I think?

donna-marijne: Okay, I thought corejs was one of our dependencies

ashley-hunter: It is, but its only included in our Angular 4 stuff, so it may not be included if they are only using our ng1 script

donna-marijne:
Removed Object.assign instead, it's not really accomplishing anything.

PR #449, opened 2017-11-13
EL-2813 - Api Properties

AlastairMcKee:
Added components for api properties, have used this on the dashboards page for now, the rest will be added when this has been approved so I wont have to redo everything if something needs changed.

https://jira.autonomy.com/browse/EL-2813

### DIFF
--- a/docs/app/components/api-properties/api-properties.component.html
+++ b/docs/app/components/api-properties/api-properties.component.html
@@ -1,0 +1,6 @@
+<div class="table-responsive">
+    <h3 *ngIf="title">{{title}}</h3>
+    <table class="api-properties-table">
+        <ng-content></ng-content>
+    </table>
+</div>
--- a/docs/app/components/api-properties/api-properties.component.html
+++ b/docs/app/components/api-properties/api-properties.component.html
@@ -1,5 +1,5 @@
 <div class="table-responsive">
-    <h3 *ngIf="title">{{title}}</h3>
+    <h4 *ngIf="title">{{title}}</h4>
     <table class="api-properties-table">
         <ng-content></ng-content>
     </table>
--- a/docs/app/components/api-properties/api-properties.component.html
+++ b/docs/app/components/api-properties/api-properties.component.html
@@ -1,5 +1,5 @@
 <div class="table-responsive">
-    <h4 *ngIf="title">{{title}}</h4>
+    <h4 *ngIf="tableTitle">{{tableTitle}}</h4>
     <table class="api-properties-table">
         <ng-content></ng-content>
     </table>
--- a/docs/app/components/api-properties/api-properties.component.ts
+++ b/docs/app/components/api-properties/api-properties.component.ts
@@ -6,5 +6,5 @@ import { Component, Input } from '@angular/core';
     styleUrls: ['./api-properties.component.less']
 })
 export class ApiPropertiesComponent {
-    @Input() title: string;
+    @Input() tableTitle: string;
 }
--- a/docs/app/components/api-properties/api-properties.component.ts
+++ b/docs/app/components/api-properties/api-properties.component.ts
@@ -1,0 +1,10 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+    selector: 'uxd-api-properties',
+    templateUrl: './api-properties.component.html',
+    styleUrls: ['./api-properties.component.less']
+})
+export class ApiPropertiesComponent {
+    @Input() title: string;
+}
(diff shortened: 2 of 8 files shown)


docs/app/components/api-properties/api-properties.component.html
```diff
@@ -0,0 +1,6 @@
+<div class="table-responsive">
+    <h3 *ngIf="title">{{title}}</h3>
```
donna-marijne: This should be h4.

docs/app/components/api-property/api-property.component.html
```diff
@@ -0,0 +1,12 @@
+<td class="api-property col-md-3">
+    <span class="highlight">{{ name }}</span>
+</td>
+<td class="api-property col-md-9">
+    <span *ngIf="required" class="italic">Required<br /></span>
```
donna-marijne: A couple of things about this:
1. Classes named `italic` and `highlight` are not acceptable. I suggest something like `api-property-metadata-name` and `api-property-metadata-value`.
2. span + br is not the right way to vertically lay these out. Please contain each line within a div that has an appropriate class name. The content should also have a parent div, in case it need particular styling in the future.

docs/app/components/api-property/api-property.component.html
```diff
@@ -0,0 +1,12 @@
+<td class="api-property col-md-3">
+    <span class="highlight">{{ name }}</span>
+</td>
+<td class="api-property col-md-9">
```
donna-marijne: Both of these tds have the same class name. It would be helpful if they were different, for purposes of targeting with CSS, e.g. `api-property-name` and `api-property-details`.

docs/app/pages/components/sections/dashboard/dashboard/dashboard.component.html
```diff
@@ -105,123 +105,37 @@ <h3 class="widget-title">Alert</h3>
...
+        the positions of the widgets within the dashboard. This is a two way binding that will be 
+        updated with the current layout when it changes.
+    </tr>
+    <tr uxd-api-property name="options" type="DashboardOptions">Configures the options for the dashboard, if an option is not 
+        specified the default value will be used.</tr>
+</uxd-api-properties>
+
+<uxd-api-properties title="Outputs">
+    <tr uxd-api-property name="layoutChange" type="DashboardOptions">Emits when layout has been changed.</tr>
+</uxd-api-properties>
```
donna-marijne: It's nice and compact - good job.

docs/app/components/api-properties/api-properties.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component, Input } from '@angular/core';
+
+@Component({
+    selector: 'uxd-api-properties',
+    templateUrl: './api-properties.component.html',
+    styleUrls: ['./api-properties.component.less']
+})
+export class ApiPropertiesComponent {
+    @Input() title: string;
```
donna-marijne: One more thing - could you renamed this property? Having an attribute called "title" on the component causes the text to appear as a tooltip when hovering over any part of the table.

AlastairMcKee:
Thanks Donna, I have made those changes. I think it might be good to merge these changes ahead of updating all of the other tables as it will allow other members of the team use the component if they are adding any new components while I update the existing ones.

docs/app/components/api-property/api-property.component.less
```diff
@@ -0,0 +1,17 @@
...
+
+.api-property-name,
+.api-property-details {
+    border: 1px solid @table-grid-border;
+    padding: 8px;
+    vertical-align: initial;
+
+    .api-property-metadata-value {
+        color: #c7254e;
+        font-style: normal;
```
donna-marijne: I just noticed this. If you arrange the spans side-by-side, you won't have to override the italic style here.

docs/app/components/api-properties/api-properties.component.ts
donna-marijne: This still needs to be addressed.

AlastairMcKee:
Sorry missed the title comment, Have made that change and updated the spans. Thanks

docs/app/pages/components/sections/dashboard/dashboard/dashboard.component.html
```diff
@@ -105,123 +105,37 @@ <h3 class="widget-title">Alert</h3>
...
-
-        <tr>
-            <td><b>@Input</b> options</td>
-            <td><b>DashboardOptions</b></td>
-            <td>Configures the options for the dashboard, if an option is not specified the default value will be used.</td>
-        </tr>
-
-    </tbody>
-</table>
+<uxd-api-properties title="Inputs">
```
donna-marijne: Change title attribute here and below.

AlastairMcKee:
Made the changes

PR #450, opened 2017-11-13
EL-2780 Investigate how to test code coverage of Protractor tests

PearseMcMurray:
See EL-2780 for detailed description of changes.


PearseMcMurray:
Only that calculating coverage seems to take a lot longer - about 90 mins with the coverage tests against a few minutes without - on my machine. Would we want to introduce a delay like that in a CI  (or release) build? If you like, I could include my changes in the SEPG job I created in the last sprint to demonstrate the integration of Protractor tests into the build to see how long calculation takes on an SEPG system.

donna-marijne:
A 90 minute run is not acceptable. Do you know why it takes so long? How does it compare with other projects using instanbul?

PearseMcMurray:
Added a further change to fix some linting errors.

PR #453, opened 2017-11-17
EL-2863 - Angular 5

ashley-hunter:
Upgraded dependencies to migrate docs to angular 5 and have our components support Angular 5.

Also now using the newer upgrade tool rather than the old one. This allows us to compile our docs with AOT for our production build for much better performing docs.

Also switched to use @ngtools as our typescript loader as this handles inlining templates & styles lazy route loading and AOT compilation (in production build).

Updated test application to Angular 5 too.

Removed any references to Angular 4 which are now just Angular in docs.

Tidied webpack configs and added some additional plugins to better report progress and to skip writing files when an error occurs.

Had to restructure our folders a little. Each page had a folder called sections and when lazy loading we had some sections with the same name eg, under component we had components/sections/buttons/buttons.module and we also had under css -> css/sections/buttons/buttons.module. As a result this caused problems with the new lazy loading as they had the same relative paths. Now they are under folders called components-sections and css-sections to fix this.

Upgraded typescript which actually revealed a few type errors in our docs which were incorrect, fixed these issue and in the code snippets associated with them.

We also now have type checking in our views with the new loader, so any errors that were detected were also fixed.

### Instructions

- Run npm install - lots of new dependencies
- run npm start site should build and work as expected.
- run grunt build. Site should now be built using AOT (should be nice and fast)
- run grunt e2e, the site should build as expected.


donna-marijne:
Did you happen to notice that loading times are quite a bit longer with Angular 5? I'm just running it with `npm start`.

ashley-hunter:
I did notice this, however this isn't a bug with anything angular related, it seems to be a bug in Webpack Dev Server and only when running https. You'll notice if you are to comment out the https line in the webpack dev server config the site will run as expected. There are large number of issues [reported online about it](https://www.google.de/search?ei=ABkPWtW0MKeR0gLwjbjwBQ&q=%5BWDS%5D+disconnected+https&oq=%5BWDS%5D+disconnected+https&gs_l=psy-ab.3...510.1242.0.1392.6.6.0.0.0.0.241.241.2-1.1.0....0...1.1.64.psy-ab..5.1.239...0j0i67k1.0.dspMHukSNaY), however I havent been able to find a solution that works yet.

PR #465, opened 2017-11-21
EL-2861 - Adding Angular search builder component

ashley-hunter:
Jira: https://jira.autonomy.com/browse/EL-2861

Migrating search builder component. Currently this is a bare bones form of the search builder, covering all the basic functionality.

Currently we provide only one default component which is a simple text input. In future I would like to add pre-defined components such as date pickers, date ranges, number inputs, selects/tags etc.. to make it super easy for teams to use - just didn't have time in this jira to create them.

I would also like to extend the search builder example to be similar to the ng1 component where you can display a side panel to select a field. While this is all currently possible with this implementation of the search builder, we will need to migrate the modal inset panel component which is currently preventing me from adding this to the example

Protractor tests will come in separate PR, want to get the component reviewed with higher priority


donna-marijne:
Is it intended to add date range components etc in a future ticket?

ashley-hunter:
Yes, ticket covering the future additions can be found here: https://jira.autonomy.com/browse/EL-2887

ashley-hunter:
Also raised ticket to migrate the inset panel which is required to provide and similar experience to the AngularJS example:
https://jira.autonomy.com/browse/EL-2888

PR #472, opened 2017-11-22
EL-2842 - Adding focus on keyboard only buttons

AlastairMcKee:
UX Aspects version, still need to do the quantum mappings for this


docs/app/pages/css/css-sections/buttons/colored-buttons/colored-buttons.component.html
```diff
@@ -25,4 +26,40 @@
 
 <p>Buttons are <code>inline-block</code> level elements so whitespace in the HTML of the page is significant and will affect their layout.</p>
 
-<uxd-snippet [content]="snippets.compiled.sampleHtml"></uxd-snippet>
\ No newline at end of file
+<uxd-snippet [content]="snippets.compiled.sampleHtml"></uxd-snippet>
+
+<br>
+
+<h3>Focus On Keyboard Control Only</h3>
```
donna-marijne: Can you make this a separate documentation section which we can link directly to?

src/styles/buttons.less
```diff
@@ -1179,4 +1179,272 @@ fieldset[disabled] .btn-toggle.active {
       color: @thumbnail-icon-color;
     }
   }
+}
+
+// focus on tab;
+
+.button-keyboard-focus:not(.disabled):not([disabled]) {
+  &.btn {
```
donna-marijne: Please explain this.

AlastairMcKee: For this css only approach the button itself does not contain the styling, we then add a span inside it which is styled like the buttons and given a tab index of -1, this is why styling has been repeated for the most part. This allows us to set up styling so that on click we will will not trigger any focus styles so focus will only occur when using keyboard controls.

donna-marijne: Can't you add this to the existing button style?

AlastairMcKee: The existing styles will be styling the .btn class, these styles are overwriting those styles and styling the .button_container class instead so there wont be any overlap. I thought keeping them separate from the existing buttons was tidier and easier to read but I can move them if you think thats preferable.

donna-marijne: What differences are there? It looks to be mostly copy pasted

donna-marijne: Okay, I'm wrong on that, don't worry about it.

AlastairMcKee:
Moved to a new section

AlastairMcKee:
misnamed pages, please wait for a fix before review

PR #481, opened 2017-11-23
Fixing incorrect imports

ashley-hunter:
Importing from specific part of ngx-bootstrap rather than root package


ashley-hunter:
Seems to be working for me? Pearse had an issue earlier where he had to clear the cache in UX Aspects as it was still carrying over the changes.

Also in the plunker main.ts the following line should be present which fixes it:

```typescript
import { PaginationModule } from 'ngx-bootstrap/pagination';
```

ashley-hunter:
I am now seeing an issue where if i open a second plunker the main.ts file stays the same as the one in the first plunker, causing errors. I believe that specific issue is related to the change chris made in his rebranding ticket. Unfortunately i'm not able to stay any later tonight to fix that as part of this PR

PR #488, opened 2017-11-30
Making changes to btn-container

AlastairMcKee:
https://jira.autonomy.com/browse/EL-2880

Updated btn-container to use margin left


donna-marijne:
Fixing the issue with split dropdowns within a flex .btn-container.

PR #489, opened 2017-12-01
S 72/el 2824 spark charts

chris-brennan:
- Jira: https://jira.autonomy.com/browse/EL-2824
- Added support to allow bar, track and theme to be set by specifying colors from the color palette.
- Tidied up some of the spark chart documentation

### DIFF
--- a/src/components/spark/spark.component.less
+++ b/src/components/spark/spark.component.less
@@ -101,10 +101,9 @@ ux-spark {
 }
 
 .ux-spark-theme(@themeName, @color) {
-    
     &.ux-spark-theme-@{themeName} {
         background-color: fade(@color, 20%);
-    
+
         .ux-spark-bar {
             background-color: @color;
         }
--- a/src/components/spark/spark.component.less
+++ b/src/components/spark/spark.component.less
@@ -8,78 +8,103 @@ ux-spark {
         &.ux-spark-theme-primary {
             .ux-spark-theme(@primary);
         }
+
         &.ux-spark-theme-accent {
             .ux-spark-theme(@accent);
         }
+
         &.ux-spark-theme-secondary {
             .ux-spark-theme(@secondary);
         }
+
         &.ux-spark-theme-alternate1 {
             .ux-spark-theme(@alternate1);
         }
+
         &.ux-spark-theme-alternate2 {
             .ux-spark-theme(@alternate2);
         }
+
         &.ux-spark-theme-alternate3 {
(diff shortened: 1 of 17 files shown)


src/components/spark/spark.component.less
```diff
@@ -71,4 +175,12 @@ ux-spark {
     .ux-spark-multi-value {
         display: flex;
     }
+}
+
+.ux-spark-theme(@color) {
```
ashley-hunter: Try and add the class name to mixin to simplify this further, so you could pass in, `.ux-spark-theme('primary', @primary)` and it would generate:

```less
&.ux-spark-theme-primary {
    background-color: fade(@primary, 20%);

    .ux-spark-bar {
        background-color: @primary;
    }
}
```

chris-brennan:
Ready for another review

src/ng1/services/colorService/colorService.service.js
```diff
@@ -80,6 +85,20 @@ export default function $colorService() {
     }
   };
 
+  $colorService.resolve = function resolve(value) {
+    if (!value) {
+        return;
+    }
+
+    for (let color in colors) {
+        if (value.toLowerCase() === color.toLowerCase()) {
```
ashley-hunter: Were we going to add a regex replace here to convert space to hyphens? Or am i thinking of somewhere else?

src/services/color/color.service.ts
```diff
@@ -92,6 +92,20 @@ export class ColorService {
             }
         }
     }
+
+    resolve(value: string): string {
+        if (!value) {
+            return;
+        }
+
+        for (let color in this._colors) {
```
ashley-hunter: Regex replace here also?

chris-brennan:
updated to use regex to convert from spaces to hyphens

src/components/spark/spark.component.ts
```diff
@@ -37,16 +66,16 @@ export class SparkComponent {
...
 
     get value() {
         return this.values;
     }
 
     constructor(private _colorService: ColorService) { 
-
+        
+        // ensure 'barColor' is an array
+        this.barColor = Array.isArray(this.barColor) ? this.barColor : [this.barColor];
```
ashley-hunter: Is this needed as it is done in the value setter?

chris-brennan:
Removed not needed code

donna-marijne:
Thank you Chris for continuing work on this.

PR #503, opened 2017-12-11
EL-2895 - Adding fixed header table directives and documentation

ashley-hunter:
Jira: https://jira.autonomy.com/browse/EL-2895

Add new AngularJS & Angular Fixed Header Table component. They now use flexbox for much better performance and added paging functionality.

Renamed the old fixed header table section to scrollable table to match the selector that it used. Added a link to that section to inform users they should be using the new Fixed Header Table directive instead.

Note: Request to add custom scrollbars doesn't seem really possible as we can't add divs to a table - perhaps css styling would suffice.


docs/app/pages/components/components-sections/tables/fixed-header-table-ng1/fixed-header-table-ng1.component.html
```diff
@@ -3,39 +3,33 @@
...
+    content is scrolled.
 </p>
 
 <p>The attributes of this directive are:</p>
 
 <uxd-api-properties>
-    <tr uxd-api-property name="watch" type="object" binding="variable" required="true">
-        Defines the values of the rows.
+    <tr uxd-api-property name="tableHeight" required="true" type="number" binding="variable">
+        This defines the maximum height of the table.
```
donna-marijne: Should mention that this is in pixels. Also is it possible to use other CSS units such as vh or % for the max height? I can imagine that would be useful.

ashley-hunter:
Review changes made.

Now, if the user supplies a number for the table height, pixels will be used by default. However if a string is supplied then the user can provide their own unit of choice.

Also added gave Id column in the example a width to show how this is done and added a comment in the description about how to do it. I tried a few different ways to do try and do it automatically but they proved to be quite problematic.

Also fixed a few inconsistencies I had overlooked between the Angular and AngularJS examples

docs/app/pages/components/components-sections/tables/fixed-header-table-ng1/fixed-header-table-ng1.component.html
```diff
@@ -3,39 +3,40 @@
...
-    </tr>
-    <tr uxd-api-property name="container-height" type="number" binding="literal" required="true">
-        Defines the height of the scrollable container.
-    </tr>
-    <tr uxd-api-property name="header-height" type="number" binding="literal" required="true">
-        Defines the height of the fixed header.
+    <tr uxd-api-property name="tableHeight" required="true" type="number | string" binding="variable">
+        This defines the maximum height of the table.
+        When a number is provided the unit will default to pixels. 
+        If a string is provided then the unit will need to be specifed as part of the string.
```
donna-marijne: Sorry, could you specify that it's CSS units, and give an example here? Other than that, this all looks fine to me!

ashley-hunter: No problem, will do that. I notice in IE it looks all wrong. The reason being is because of quirks mode in IE. Adding the doctype fixed the issue. I was wondering if perhaps I could add https://jira.autonomy.com/browse/EL-2548 to this sprint (the ticket to bring back doctype)? Apart from the review comments you have made I have only a ticket to do the contributing guide left to do, so am quite a bit ahead of schedule. If i bring this ticket into this sprint it will resolve the IE issue I am having.

donna-marijne: Yes, absolutely. I have a feeling that it will be a can of worms but sure if you have most of the rest of the sprint free, it would be great to get it done!

ashley-hunter: Yeah, I would like to get it done, I think the longer it is left the more we are going to have to fix. Great, thanks!

ashley-hunter:
Changes made - included HTML Doctype changes in this PR.

PR #510, opened 2017-12-22
El 2883 (Integrate protractor tests into the continuous integration build)

PearseMcMurray:
Execute Protractor tests and return number of failures.
Rename installed version of chromedriver so that it does not include a version number.
Record failure if any unit or Protractor tests fail.
Convert results XML data to HTML.
Don't build documentation if Protractor tests fail.
Commented out some tests not working with headless Chrome. To be fixed in another JIRA.


grunt/clean.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
     images: 'dist/img',
     less: 'dist/less',
     licenses: 'dist/Licenses',
-    e2e: ['e2e/dist', 'e2e/_test-output', 'e2e/coverage'],
+    e2e: ['e2e/dist', 'e2e/_test-output', 'e2e/coverage', 'e2e/html', 'e2e/xml'],
     shim: 'dist/dist',
-    bower: 'target/bower'
+    bower: 'target/bower',
+    webdriver: 'node_modules/protractor/node_modules/webdriver-manager/selenium/chromedriver_*.zip'
```
donna-marijne: Can we just copy the driver into e2e/dist and use it from there? And have it cleaned up with everything else? I don't really like the idea of adding/removing files in node_modules apart from via npm.

grunt/clean.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
     images: 'dist/img',
     less: 'dist/less',
     licenses: 'dist/Licenses',
-    e2e: ['e2e/dist', 'e2e/_test-output', 'e2e/coverage'],
+    e2e: ['e2e/dist', 'e2e/_test-output', 'e2e/coverage', 'e2e/html', 'e2e/xml', 'e2e/chromedriver.exe'],
     shim: 'dist/dist',
-    bower: 'target/bower'
+    bower: 'target/bower',
+    webdriver: 'node_modules/protractor/node_modules/webdriver-manager/selenium'
```
donna-marijne: Why is this still being deleted? I don't understand.

PearseMcMurray: In the context of the continuous integration build this cleanup isn't needed as the repository is recloned for each build. My concern is what could happen when the 'e2e' task is run on development machines without this cleanup. On one day the chromedriver_2.34 binary could be downloaded by the 'run:webdriver_manager_update' task; on the next the chromedriver_2.35 could appear. The 'copy:webdriver' task would have to recognise the newer version before copying it beside the protractor configuration file. The simpler option seemed to be to delete the entire folder each time.

donna-marijne: Does it make sense to always use the latest version as soon as it comes out? Perhaps we should be sticking with a particular version until we can verify that a newer version does not introduce some unexpected behaviors?

PearseMcMurray: We can do that . To avoid storing the binary in the repository, a new task would be needed to download it. There are curl and wget plugins for grunt.

donna-marijne: Could you perhaps use npm to get the webdriver? E.g. https://www.npmjs.com/package/chromedriver. I suggest this because we can then use package.json to manage the version.

PearseMcMurray: Yes, that works. I will make that change.

PR #515, opened 2018-01-02
EL-2904 - Infinite Scroll Refresh

ashley-hunter:
Reworked the infinite scroll directive. Rather than manually create all the elements, I have now changed it to use ng-repeat for the list of items and ng-if for showing/hiding items. This allows us to easily update data by letting angular handle all the diffing - and this also reduced the amount of code required significantly.

Added two new broadcast events. One allows the user to reload all currently loaded pages - essentially if there are 5 pages loaded this will call the paging function for pages 1-5 and update the data when it is returned. I also added a reloadPage event that can be passed a specific page number - this will only request the data for that page and update that page only.

Also split out scrollbar handling. We supported scrolling on an element, on a scroll pane and on the window and this previously involved lots of duplicated checks to see which scrollbar was being used. Now extracted these into separate classes with the same surface API. Now we do a check at the start to determine which type of scrollbar is being used and then use the appropriate class which will handle all future scroll events, interactions and cleanup

### DIFF
--- a/docs/app/pages/components/components-sections/scrollbar/infinite-scroll-ng1/infinite-scroll-ng1.component.html
+++ b/docs/app/pages/components/components-sections/scrollbar/infinite-scroll-ng1/infinite-scroll-ng1.component.html
@@ -59,9 +59,19 @@
     </tr>
 </uxd-api-properties>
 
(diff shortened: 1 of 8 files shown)


donna-marijne:
Do we need a ticket to add something similar to the Angular equivalent?

ashley-hunter:
I can probably do it as part of this ticket. I will raise a separate PR for it.

ashley-hunter:
Codepen issue should now be resolved

PR #522, opened 2018-01-12
EL-2897 Hierarchical Navigation - Angular Migration

PearseMcMurray:
Protractor tests to follow.


docs/app/pages/components/components-sections/side-navigation/navigation/snippets/app.html
```diff
@@ -0,0 +1,34 @@
+<ux-side-navigation style="width: 750px">
```
donna-marijne: Please use the hierarchical example from http://m.everythingux.net/styleguide/ui-components/side-menus/ here. This is implemented in the MF documentation, for reference.

src/components/side-navigation/side-navigation-item.component.html
```diff
@@ -0,0 +1,11 @@
+<div class="menu-item" (click)="select()">
+
+  <i class="hpe-icon menu-item-icon" *ngIf="icon" [ngClass]="icon"></i>
+  <span class="menu-item-title">{{ header }}</span>
+  <span class="menu-item-state-icon hpe-icon hpe-previous" [class.expanded]="expanded" *ngIf="children.length !== 1"></span>
```
donna-marijne: I think the chevron should be on the left of the nav item, as per MF guidelines. Although this differs from the Keppel theme, that is no longer our priority. This will make it easier to style.

src/components/side-navigation/side-navigation-item.component.less
```diff
@@ -0,0 +1,77 @@
...
+
+    &.active {
+        & > .menu-item {
+            & > .menu-item-title {
+                font-weight: bold;
+            }
+        }
+    }
+
+    &.fourth-level {
```
donna-marijne: Could you use classes with digits, e.g. `level-5`, which will make it easier to assign classes based on level.

src/components/side-navigation/side-navigation.component.html
```diff
@@ -0,0 +1,33 @@
+<ux-side-navigation style="width: 750px">
```
donna-marijne: I presume this example code is included here by mistake?

src/components/side-navigation/side-navigation.component.less
```diff
@@ -0,0 +1,6 @@
+ux-side-navigation {
+    display: block;
+    border-left: 4px solid #05aba5;
```
donna-marijne: Variable for this color?

src/components/side-navigation/side-navigation-item.component.less
```diff
@@ -0,0 +1,77 @@
...
+    font-weight: 400;
+    font-size: 0.9375rem;
+
+    &.no-children {
+        &.active {
+            font-weight: 600;
+        }
+    }
+
+    &.collapse[style] {
```
donna-marijne: What is this for? I don't think we should ever make a selector dependent on the style attribute.

src/components/side-navigation/side-navigation-item.component.ts
```diff
@@ -0,0 +1,87 @@
+import { Component, Input, ViewEncapsulation, ContentChildren, QueryList, Renderer2, ElementRef, Optional, SkipSelf } from '@angular/core';
+import { SideNavigationService } from './side-navigation.service';
+
+@Component({
+  selector: 'ux-side-navigation-item',
+  templateUrl: './side-navigation-item.component.html',
+  host: {
+    '[class.active]': 'active',
```
donna-marijne: If I understand correctly, this needs to change. We need to distinguish between the node which is active (i.e. the current navigation route) and ancestors of the node which is active. As I said before, only the active node will be highlighted in the MF theme, so this is important to consider.

src/components/side-navigation/side-navigation.service.ts
```diff
@@ -0,0 +1,10 @@
+import { Injectable } from '@angular/core';
+import { Subject } from 'rxjs/Subject';
+
+@Injectable()
+export class SideNavigationService {
+
+    deselect: Subject<void> = new Subject<void>();
```
donna-marijne: I think that the Angular router should provide the necessary observables to implement selection/deselection, so I don't think that this service is required. I think what is needed is some kind of provider interface to allow us to plug in to the router, and also use other sources of navigation as required, e.g. a dummy router for the purposes of example code. 

I'll add Ashley as a reviewer, since he has implemented something similar to this recently, and will hopefully be able to give you some explicit guidance.

ashley-hunter: This service likely can be removed now. In a real application the user would be using Angular's routerLink & routerLinkActive directives, which essentially just adds a specified class to an element when it's route is active, eg:

```html
 <ux-side-navigation-item [expanded]="true" header="Menu Levels" icon="hpe-cluster">

    <ux-side-navigation-item header="Second Level">

        <ux-side-navigation-item header="Third Level Item" routerLink="item1" routerLinkActive="active"></ux-side-navigation-item>

    </ux-side-navigation-item>

</ux-side-navigation>
```

We wouldn't want to tie this component directly to the router as they may not always want a click event to perform a navigation, instead they may want to call a function or show a popover for example.

It would probably be best to start by creating a plunker for the section that actually uses the Angular router and use the routerLink and routerLinkActive directives as it would most commonly in a real application. The code to do that should be documented. 

In the example we show in our docs site we can't use the router, but it is easy enough to fake this by simply adding a, `active` class to one of the menu items when they click on it.

donna-marijne: `routerLinkActive` would also need to control whether the nav item and its ancestors are expanded. So it might be advisable to provide an integration.

ashley-hunter: Yes, it can be injected into a constructor or the menu item and marked as optional, so it should work both with and without the router being used.

```
constructor(@Optional() routerLinkActive: RouterLinkActive) {
}
```

docs/app/interfaces/IPlunk.ts
```diff
@@ -14,6 +14,11 @@ export interface IPlunk {
         alias: string;
         source: string;
     }[];
+    routes?: {
+        path: string;
+        pathMatch?: string;
+        component: string
+    }[];
```
donna-marijne: Hello Pearse. Please don't add this, it's only useful for your one example. All you need is to allow a custom module template. Can you do that instead? And check things like this before spending time on them?

donna-marijne:
Can this PR be closed?

PearseMcMurray:
Yes. Will I close it or you?

PR #533, opened 2018-01-23
EL-2937 - Fixing slider bug

ashley-hunter:
JIRA: https://jira.autonomy.com/browse/EL-2937
Related PRs: https://github.houston.softwaregrp.net/caf/quantum/pull/339 & https://github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/pull/122

Normally a minor tick will not appear when a major tick is in the same position - this is expected, however if major ticks are set to be hidden, minor ticks in the same position are still hidden which is incorrect.


ashley-hunter:
Added additional safety checks to snapping functionality - error should no longer occur

PR #539, opened 2018-01-30
In the Plunker service the default 'configJs' configuration template …

PearseMcMurray:
…may now be overridden by the ux-aspects-micro-focus copy.


donna-marijne:
What was the purpose of this change? The config.js file is customizable.

PearseMcMurray:
A new version of config_js.txt has been added in ux-aspects-micro-focus. It contains the line

        '@micro-focus/ux-aspects': window.uxdAssetsUrl + '/lib/ux-aspects.js'

This is needed in the Micro Focus project to access the new NavigationModule in Plunker. Adding this line to the UXAspects copy of config_js.txt would cause an error when a Plunker is loaded.

donna-marijne:
1. The file is customizable. Look at the following statement, right next to where you made your changes:
````
            this.configJs = require('./templates/config_js.txt')
            .replace(MAPPINGS_PLACEHOLDER, mappings.join(',\n\t\t\t\t'));
````
2. Why is there a NavigationModule in `@micro-focus/ux-aspects`? Who did you agree this change with?

PearseMcMurray:
Have I misunderstood what I was to do here?

Originally, the plan was to add the Angular component in UXAspects and import it into Micro Focus. I had thought we had agreed to instead create the new component in Micro Focus and return to the UXAspects version later (if I recall correctly, this was due to differences in the styling between the UXAspects and Micro Focus Ng1 versions). That is why I created the new component in ux-aspects-micro-focus.

donna-marijne:
Yes, it's a misunderstanding. I asked you to write documentation in the Micro Focus site only, since you were struggling with the Keppel themed version. If there is doubt, just ask.

PR #541, opened 2018-01-30
EL-2897 Moved NavigationModule from ux-aspects-micro-focus to UXAspects.

PearseMcMurray:


src/components/navigation/navigation-item/navigation-item.component.html
```diff
@@ -0,0 +1,7 @@
+<a *ngIf="link" (click)="toggle(); $event.stopPropagation()" [class.has-arrow]="this.children.length > 0" [class.no-arrow]="this.indentWithoutArrow()" [routerLink]="link">
```
donna-marijne: You don't toggle if you have a link - NavigationEnd should handle the expansion of parents.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+  QueryList,
+  Renderer2,
+  ElementRef,
+  Optional,
+  SkipSelf,
+  AfterContentInit,
+  ContentChild,
+  Output,
+  AfterViewInit,
+  HostListener
```
donna-marijne: Make indentation consistent. We use 4 spaces for typescript files in this repository.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+    encapsulation: ViewEncapsulation.None
+})
+export class NavigationItemComponent implements AfterViewInit {
+    @Input() header: string;
+    @Input() icon: string;
+    @Input() expanded: boolean = false;
+    @Input() link?: string;
+
+    @Input()
+    set level(value: number) {
```
donna-marijne: What does it mean to set `level` as an input property? The level should be calculated from the hierarchy.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+    get level(): number {
+      return this._level;
+    }
+
+    @Output()
+    get active() {
+      if (this.link) {
+          return this._router.isActive(this.link, true);
+      } else {
+          return this._active;
```
donna-marijne: I feel like we had this discussion before - what is the use case for a nav item being active other than being the active route?

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+      }
+    }
+
+    private _level: number = 2;
```
donna-marijne: Why is this 2 by default?

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+        private _elementRef: ElementRef,
+        private _renderer: Renderer2,
+        @Optional()
+        @SkipSelf()
+        private _parent: NavigationItemComponent,
+        @Optional() private _router: Router
+    ) {
+        this.level = _parent ? _parent.level + 1 : 2;
+
+        if (this._router) {
```
donna-marijne: I don't understand this. Router is a part of Angular. Why do we need to check for it?

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+
+      return false;
+    }
+
+    private getLevelClass(): string {
+      var levelClass: string = '';
+      switch (this._level) {
+        case 2:
+          levelClass = 'nav-second-level';
+          break;
```
donna-marijne: Make this more compact by doing e.g. `return 'nav-second-level'`.

src/components/navigation/navigation.component.html
```diff
@@ -0,0 +1,12 @@
+<nav class="navbar-default navbar-static-side-dark tree">
+    <ol class="nav">
+        <ng-content></ng-content>
+    </ol>
+</nav>
+
+<br>
+<br>
+<hr>
+<br>
```
donna-marijne: Remove the above 4 lines.

src/components/navigation/navigation-item/navigation-item.component.html
ashley-hunter: Also, remove the use of `this` in the view, it is not needed as the component class is the context used for rendering the view

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+  Component,
+  Input,
+  ViewEncapsulation,
+  ContentChildren,
+  QueryList,
+  Renderer2,
+  ElementRef,
+  Optional,
+  SkipSelf,
+  AfterContentInit,
```
ashley-hunter: Review imports, some of these are never used

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+    private hasActiveLink(link: string | UrlTree): boolean {
+      if (link) {
+        return this._router.isActive(link, true) || (this._parent && this._parent.hasActiveLink(link));
+      }
+
+      return false;
+    }
+
+    private getLevelClass(): string {
+      var levelClass: string = '';
```
ashley-hunter: Should never use `var` any more. If the value never changes use `const` otherwise use `let`

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    ngAfterViewInit() {
+      let ol = this._elementRef.nativeElement.querySelector('.nav');
+      if (ol) {
+        this._renderer.addClass(ol, this.getLevelClass());
+      }
+    }
+
+    toggle() {
```
ashley-hunter: Add return types to functions, both for consistency and also we potentially get better error detection if changes are made in future

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+        '[class.active]': 'active',
+        '[class.selected]': 'expanded',
+    },
+    encapsulation: ViewEncapsulation.None
+})
+export class NavigationItemComponent implements AfterViewInit {
+    @Input() header: string;
+    @Input() icon: string;
+    @Input() expanded: boolean = false;
+    @Input() link?: string;
```
ashley-hunter: Remove the `?`. By default the value will be undefined unless the user specifies it.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,151 @@
...
+                this.select();
+            }
+
+            this.expanded = this.hasActiveLink(this.link);
+          });
+        }
+    }
+
+    ngAfterViewInit() {
+      let ol = this._elementRef.nativeElement.querySelector('.nav');
```
ashley-hunter: Use `const` and also possibly rename this variable to give better indication about what we are searching for

donna-marijne:
Pending resolution to https://github.houston.softwaregrp.net/caf/quantum/pull/358.

src/components/navigation/navigation-item/navigation-item.component.html
PearseMcMurray: toggle() removed from anchor element with router link.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Changed indentation to four spaces in all files.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Unneeded set function for _level variable removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: get active() modified - only active routes can return true.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Default value of _level changed to 1. Along with a change to ngAfterViewInit() this will allow the assigment of classes like nav-second-level, nav-third-level, etc to match their distribution in the ng1 example.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Removed check for existence of router. This was a remnant of checks I was making earlier in development concerning which router objects (Router, ActivatedRoute, ...) were available in the documentation example and the Plunker example.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Reworked getLevelClass().

src/components/navigation/navigation.component.html
PearseMcMurray: Unneeded lines removed.

src/components/navigation/navigation-item/navigation-item.component.html
PearseMcMurray: Removed unnecessary 'this' in template.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Removed unused imports.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: '?' removed.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    },
+    encapsulation: ViewEncapsulation.None
+})
+export class NavigationItemComponent implements AfterViewInit {
+    @Input() header: string;
+    @Input() icon: string;
+    @Input() expanded: boolean = false;
+    @Input() link: string;
+
+    @Output()
```
ashley-hunter: Can remove @Output as this is not an eventemitter. Also likely dont need _level any more, just make this:

```typescript
level: number = 1;
```

Remove the @output from active as well. Also active getter never uses _active so this can be removed

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Renamed variable in ngAfterViewInit() and declared as const.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Added return types to functions.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Replaced var with const or let.

src/components/navigation/navigation.component.ts
```diff
@@ -0,0 +1,10 @@
+import { Component, Input, ViewEncapsulation } from '@angular/core';
+
+@Component({
+    selector: 'ux-navigation',
+    templateUrl: './navigation.component.html',
+    encapsulation: ViewEncapsulation.None
```
ashley-hunter: We can remove this line and the associated import. This is only required if we specify a styleUrls property here

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,138 @@
...
+        const parentListElement = this._elementRef.nativeElement.parentElement;
+        if (parentListElement) {
+            let levelClass: string = this.getLevelClass();
+            if (levelClass.length > 0) {
+                this._renderer.addClass(parentListElement, levelClass);
+            }
+        }
+    }
+
+    toggle(): void {
```
ashley-hunter: Remove this function - this can be done in the view which is more performant, eg:

```
(click)="expanded = !expanded"
```

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,138 @@
...
+        private _elementRef: ElementRef,
+        private _renderer: Renderer2,
+        @Optional()
+        @SkipSelf()
+        private _parent: NavigationItemComponent,
+        @Optional() private _router: Router
+    ) {
+        this._level = _parent ? _parent.level + 1 : 1;
+
+        _router.events.filter(event => event instanceof NavigationEnd).subscribe(event => {
```
ashley-hunter: You need to store the subscription of this and dispose it when the component is destroyed - look into ngOnDestroy lifecycle hook

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,138 @@
...
+import 'rxjs/add/operator/filter';
+
+@Component({
+    selector: '[ux-navigation-item]',
+    templateUrl: './navigation-item.component.html',
+    host: {
+        '[class.active]': 'active',
+        '[class.selected]': 'expanded',
+    },
+    encapsulation: ViewEncapsulation.None
```
ashley-hunter: Can remove this and associated import

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    ElementRef,
+    Optional,
+    SkipSelf,
+    ContentChild,
+    Output,
+    AfterViewInit,
+  HostListener
+} from '@angular/core';
+import { Router, RouterLinkActive, NavigationEnd, RouterState, ActivatedRoute, UrlTree } from '@angular/router';
+import 'rxjs/add/operator/filter';
```
ashley-hunter: Instead of using this rxjs import here, it is not recommended to use lettable operators instead. [Read this for an intro](https://blog.hackages.io/rxjs-5-5-piping-all-the-things-9d469d1b3f44).

This would change the import to something like:

```typescript
import { filter } from 'rxjs/operators/filter';

...

_router.events.pipe(filter(event => event instanceof NavigationEnd)).subscribe...
```

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    Component,
+    Input,
+    ViewEncapsulation,
+    ContentChildren,
+    QueryList,
+    Renderer2,
+    ElementRef,
+    Optional,
+    SkipSelf,
+    ContentChild,
```
ashley-hunter: Import never used

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,138 @@
...
+    Output,
+    AfterViewInit,
+  HostListener
```
ashley-hunter: Import never used

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,138 @@
...
+} from '@angular/core';
+import { Router, RouterLinkActive, NavigationEnd, RouterState, ActivatedRoute, UrlTree } from '@angular/router';
```
ashley-hunter: RouterLinkActive, RouterState and ActivatedRoute can be removed as they are never used

src/components/navigation/navigation-item/navigation-item.component.html
```diff
@@ -0,0 +1,7 @@
+<a *ngIf="link" [class.has-arrow]="children.length > 0" [class.no-arrow]="indentWithoutArrow()" [routerLink]="link">
```
ashley-hunter: Is it possible to determine in code if it should have no arrow and store the result in a variable instead of calling `indentWithoutArrow()` in the view? This would result in this function being called every time change detection occurs which is quite inefficient.

src/components/navigation/navigation.component.html
```diff
@@ -0,0 +1,7 @@
+<nav class="navbar-default navbar-static-side-dark tree" role="navigation">
```
donna-marijne: Remove these outdated classes, as previously requested. (We do not support a dark theme, apart from anything.)

src/components/navigation/navigation-item/navigation-item.component.html
donna-marijne: Perhaps an observable based on `this._children.changes`?

PearseMcMurray: Information now stored in variable which is assigned a value in ngAfterContentInit().

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Import removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Import removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Imports removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Lettable operator used as suggested.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: encapsulation and import removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: @Output removed from active().
level now a public member. level() removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Subscription now stored and disposed of in ngOnDestroy().

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Function removed and view updated as suggested.

src/components/navigation/navigation.component.ts
PearseMcMurray: encapsulation and associated import removed.

src/components/navigation/navigation.component.html
PearseMcMurray: navbar-default and navbar-static-side-dark removed.
Tree class still needed to allow correct styling in MF. Without it, top-level classes will have height 56px instead of the correct 40px.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -61,9 +54,9 @@ export class NavigationItemComponent implements AfterViewInit {
         private _parent: NavigationItemComponent,
         @Optional() private _router: Router
     ) {
-        this._level = _parent ? _parent.level + 1 : 1;
+        this.level = _parent ? _parent.level + 1 : 1;
 
-        _router.events.filter(event => event instanceof NavigationEnd).subscribe(event => {              
+        this._navigationEnd = _router.events.pipe(filter(event => event instanceof NavigationEnd)).subscribe(event => {
```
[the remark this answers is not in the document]
ashley-hunter: There are quite a few benefits outlined in the link below - pretty much the way we were doing before is not tree-shakable. Plus I believe the way we were doing before will be removed in rxjs v6 which will be used by Angular v6 by March/April

https://github.com/ReactiveX/rxjs/blob/master/doc/pipeable-operators.md#why

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -73,20 +66,25 @@ export class NavigationItemComponent implements AfterViewInit {
...
+                this._renderer.addClass(parentListElement, 'nav');
                 this._renderer.addClass(parentListElement, levelClass);
             }
         }
     }
 
-    toggle(): void {
-        this.expanded = !this.expanded;
+    ngAfterContentInit(): void {
+        this.setIndentWithoutArrow();
```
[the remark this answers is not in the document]
PearseMcMurray: Subscription added. Sorry, I missed the significance of this change.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,142 @@
...
+        return this._children.filter(item => item !== this);
+    }    
+
+    constructor(
+        private _elementRef: ElementRef,
+        private _renderer: Renderer2,
+        @Optional()
+        @SkipSelf()
+        private _parent: NavigationItemComponent,
+        @Optional() private _router: Router
```
donna-marijne: Remove Optional here.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,142 @@
...
+        // Update 'indentWithoutArrow' in response to changes to children
+        this._childrenChanges = this._children.changes.subscribe(() => this.setIndentWithoutArrow());
+    }
+
+    ngOnDestroy () {
+        this._navigationEnd.unsubscribe();
+        this._childrenChanges.unsubscribe();
+    }
+
+    select(): void {
```
donna-marijne: This function is not needed as far as I can tell. Expanding parents should be taken care of by the code above:

````
this.expanded = this.hasActiveLink(this.link);
````

File: src/components/navigation/navigation-item/navigation-item.component.ts [hunk 29]
```diff
@@ -0,0 +1,142 @@
...
+        private _renderer: Renderer2,
+        @Optional()
+        @SkipSelf()
+        private _parent: NavigationItemComponent,
+        @Optional() private _router: Router
+    ) {
+        this.level = _parent ? _parent.level + 1 : 1;
+
+        this._navigationEnd = _router.events.pipe(filter(event => event instanceof NavigationEnd)).subscribe(event => {              
+            if (this.link && this._router.isActive(this.link, true)) {
```
donna-marijne: Remove this. `this.expanded = this.hasActiveLink(this.link);` takes care of it.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,142 @@
...
+        '[class.selected]': 'expanded',
+    }
+})
+export class NavigationItemComponent implements AfterViewInit, AfterContentInit, OnDestroy {
+    @Input() header: string;
+    @Input() icon: string;
+    @Input() expanded: boolean = false;
+    @Input() link: string;
+
+    get active() {
```
donna-marijne: Specify return type: boolean.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: @Optional decorator removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: select() function removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: select() function removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Return type added to active() getter.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,135 @@
...
+        @SkipSelf()
+        private _parent: NavigationItemComponent,
+        private _router: Router
+    ) {
+        this.level = _parent ? _parent.level + 1 : 1;
+
+        this._navigationEnd = _router.events.pipe(filter(event => event instanceof NavigationEnd)).subscribe(event => {              
+            this.expanded = this.hasActiveLink(this.link);
+            if (this.expanded) {
+                // Expand ancestors
```
donna-marijne: I don't think this is needed. All navigation items subscribe to this event, and expand as required, so this would be duplicated effort.

src/components/navigation/navigation.component.html
```diff
@@ -0,0 +1,7 @@
+<nav class="tree" role="navigation">
+    <ol class="nav">
+        <ng-content></ng-content>
+    </ol>
+</nav>
+
+<router-outlet></router-outlet>
```
donna-marijne: Isn't it up to the application to determine where the router outlet goes? Why was this added here?

donna-marijne:
Any update on this PR coming soon?

PearseMcMurray:
I haven't had much time to work on this today so I won't be able to submit this today. I have been having problems with the expansion of parent folders when lowest-level items are selected. Can this be moved to the next sprint? After today's planning session I have free time.

donna-marijne:
Done

PearseMcMurray:
Moved router-outlet directive to documentation.
hasActiveLink() now sets the 'expanded' property. Then calls hasActiveLink() in ancestors to set the 'expanded' property in them also. Previously it was not being set in the ancestors.

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -0,0 +1,133 @@
...
+
+    ngOnDestroy () {
+        this._navigationEnd.unsubscribe();
+        this._childrenChanges.unsubscribe();
+    }
+
+    private hasActiveLink(link: string | UrlTree): void {
+        if (link) {
+            this.expanded = this._router.isActive(link, true);            
+            if (this.expanded && this._parent) {
```
donna-marijne: This is still wrong. Revert to the original version and please address the review comment. Contact me for clarification if you do not understand.

PearseMcMurray: I have reverted to the original code but expansion of parents of a selected item does not occur, whether they themselves have a link or not. I have experimented with addition and removal of router links in the example HTML and with changes to the link definitions in the router module.

The call to hasActiveLink in the NavigationEnd handler sets the value of the 'expanded' flag according to the component's link. For parents of selected items this is always false. The recursive call within hasActiveLink will return 'true' but it doesn't set the expanded flag in the component.

What am I missing here?

donna-marijne: Something like this?

````
private hasActiveLink(link: string | UrlTree): boolean {
    if (link) {
        return this._router.isActive(link, true) || this.children.any((item) => item.hasActiveLink(item.link));
    }

    return false;
}
````

Explanation: Every component receives the NavigationEnd event. We use this to determine the state of the `expanded` property for each node, ensuring that the entire tree is always up to date with every navigation event. This avoids the need to go back and collapse siblings explicitly. This makes use of recursion, which is okay since the maximum depth is 5, and keeps the code tidy.

Let me know if there is anything I've overlooked with this, or anything that you don't understand. It's not a problem to ask for help, as previously agreed.

**Comment by PearseMcMurray** [comment]
'expanded' property set if component is the active link or if any of its descendants are the active link.

File: src/components/navigation/navigation-item/navigation-item.component.ts [hunk 34]
```diff
@@ -90,17 +90,22 @@ export class NavigationItemComponent implements AfterViewInit, AfterContentInit,
...
 
-            return;
+        // If this component has children, check if any of them, or their descendants, are active.
+        let hasActiveChildren = false;
+        if (this.children.length > 0) {
+            hasActiveChildren = this.children.some((item) => item.hasActiveLink(item.link));
         }
 
-        this.expanded = false;
+        return hasActiveChildren;
```
donna-marijne: It's fine if you want to write this out for your own understanding, but try to condense it and remove redundant variables, since the browser ultimately has to download the javascript version of this.

This would be more readable and concise:

````
if (link && this._router.isActive(link, true)) {
  return true;
}

return this.children.some((item) => item.hasActiveLink(item.link));
````

**Comment by donna-marijne** [comment]
@ashley-hunter Could you verify these changes and merge when all comments have been addressed to the appropriate standard? Let's try get this done today.

**Comment by ashley-hunter** [comment]
@donna-marijne sure! No problem!

File: src/components/navigation/navigation-item/navigation-item.component.ts (hunk 34 above)
Comment (PearseMcMurray) [reply src/components/navigation/navigation-item/navigation-item.component.ts]: Condensed code as suggested.

File: src/components/navigation/navigation-item/navigation-item.component.ts [hunk 35]
```diff
@@ -0,0 +1,130 @@
...
+    AfterViewInit,
+    AfterContentInit,
+    OnDestroy
+} from '@angular/core';
+import { Router, NavigationEnd, UrlTree } from '@angular/router';
+import { filter } from 'rxjs/operators/filter';
+import { Subscription } from 'rxjs/Subscription';
+
+@Component({
+    selector: '[ux-navigation-item]',
```
ashley-hunter: Attribute selectors should follow upper camel case naming convention, eg: `uxNavigationItem`

ashley-hunter: Leave this as is for now, this is a bit of a grey area as components generally don't use an attribute selector. We can change this at a later point if we want to

PR #543, opened 2018-01-31
EL-2961 - Adding wrapper for Sliders with Charts

ashley-hunter:
Adding a wrapper for sliders with charts. Initially also intended to add a wrapper for modal inset panel but upon closer inspection discovered that it uses the transclude function to perform manual transclusion which is not supported, raised a separate ticket to cover the migration of it.


donna-marijne:
Is this used in the documentation example?

ashley-hunter:
Not currently, however i can change it to if youd like

donna-marijne:
I'll re-merge this after the release

donna-marijne:
Merged now.

PR #551, opened 2018-02-01
EL-2557 - Marquee Wizard Proposal

ashley-hunter:
Added documentation proposal for the marquee wizard migration.

We can simply extend the WizardComponent and WizardStepComponent classes and add any additional functionality required avoiding a lot of code duplication. Because of this the inputs and outputs are largely identical.

### DIFF
--- a/docs/app/data/components-page.json
+++ b/docs/app/data/components-page.json
@@ -1424,7 +1424,23 @@
                 {
                     "title": "Marquee Wizard",
                     "component": "ComponentsMarqueeWizardNg1Component",
-                    "version": "AngularJS"
+                    "version": "AngularJS",
+                    "deprecated": true
+                },
+                {
+                    "title": "Marquee Wizard",
+                    "component": "ComponentsMarqueeWizardComponent",
+                    "version": "Angular",
+                    "usage": [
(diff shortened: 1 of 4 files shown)


donna-marijne:
Also the step needs some extra properties, notably `hidden`. I don't think `header` is still needed since it can be added as part of the step template. There might still be a use case for setting `completed` programmatically.

ashley-hunter:
I have removed the "header" in the way that we had it in the ng1 version - before this added a header at the top of the content which is no longer the case. This instead refers to the text displayed in the list of steps on the left. Perhaps you can think of a better name for it?

I don't think hidden is required, as doing a simple ngIf works?

donna-marijne:
I agree that the header option is no longer required.

ashley-hunter:
Updated the outputs to add stepChanging event. Also added a description input which can take a string if they just wish to add some simple text, or a templateRef allowing custom html in the content to add custom html and icons etc..

Hidden property not required as ngIf works for hiding steps.

Header property now refers to the text in the tab, rather than text at the top of the step content, eg.:

image

Added input and output for the completed state of each step

PR #560, opened 2018-02-08
EL-2557 - Marquee Wizard Implementation

ashley-hunter:
Component implementation complete. Additional PR for code snippets, plunker, tests and MF theme will be created.

- Created Marquee Wizard component that extends existing wizard component classes and only adds additional functionality and a new template.

- Added onFinishing event to the existing wizard component as stepChanging does not get fired on final step and makes sense that it shouldn't.

- Removed duplicate less imports

- Renamed hybrid component classes as there was a class name collision when i added the MarqueeWizardComponent class.

- Added type="button" to AngularJS marquee wizard as it closed if you pressed enter when the modal was open


src/components/marquee-wizard/marquee-wizard.component.html
```diff
@@ -0,0 +1,43 @@
...
+
+    </ul>
+</div>
+
+<div class="marquee-wizard-content-panel">
+    <div class="marquee-wizard-content">
+        <ng-content></ng-content>
+    </div>
+
+    <div class="marquee-wizard-footer">
```
donna-marijne: I think if you use `.btn-container` here, it give correct flex layout and spacing between buttons. Also I think `.modal-footer` is similar.

PR #562, opened 2018-02-12
EL-2962 - wrapper docs

donna-marijne:
Added `hybrid` property to the page json definition for all sections related to HybridModule components.
Added "Angular Compatible" clickable label to the header of hybrid sections with some info in a popover.
Removed the doc wrapper from the floating action button section as a test.
Tooltips positioned below the labels to match the popover.

https://jira.autonomy.com/browse/EL-2962


docs/app/components/component-section/component-section.component.html
```diff
@@ -12,19 +12,58 @@ <h3 class="uxd-component-section-title">{{ title }}</h3>
...
     <ux-ebox-content class="uxd-component-section-content">
         <div #container></div>
     </ux-ebox-content>
-</ux-ebox>
\ No newline at end of file
+</ux-ebox>
+
+<ng-template #hybridHelpContent>
+    <p>
+        To use this AngularJS component in an Angular 5 application, import <code>HybridModule</code> into the module
```
ashley-hunter: Possibly removed the 5 from Angular 5 so we don't have to keep updating this?

docs/app/interfaces/ISection.ts
```diff
@@ -4,6 +4,7 @@ export interface ISection {
     component: string;
     codepen: boolean;
     version: 'AngularJS' | 'Angular';
+    hybrid: boolean;
```
ashley-hunter: Possibly make this an optional property as it doesn't seem to be used in every section.

docs/app/components/component-section/component-section.component.html
donna-marijne: Yeah true. Would the hybrid components work in a site based on Angular 4 or lower?

ashley-hunter: They should work fine in an Angular 4 or 5 application. I don't believe they will work in Angular 2 applications, however Angular releases are only designed to be backwards compatible with the previous major release so we can only ever support the 2 most recent major versions of Angular anyway so none of our library is guaranteed to work with Angular 2 anymore.

donna-marijne:
Made those changes; also added the property to sections in the Components page.

PR #575, opened 2018-02-16
EL-3015 - Dashboard Improvements

ashley-hunter:
- Simplified a lot of the dashboard code
- Separated drag functionality into a separate directive as we were duplicating this code quite a lot, and there are also many other future uses for this functionality
- Updated sample code - it was using attribute before rather than binding to the property value
- No longer using the doCheck lifecycle hook, much more efficient to use setters/getters
- Now using observables in many places. Before there were lots of values that were somewhat dependant on each other, this now ensures values are kept in sync. This allow allows us to use OnPush change detection for better performance, and also prevents the 'Expression has changes after check' error.
- Also fixed bug when resizing downwards past the lower bound of the dashboard
- Updated resize service to no longer require renderer to be injected.


ashley-hunter:
I will do that! Thanks

PR #586, opened 2018-02-25
EL-2941 (Timeline - Angular Migration)

PearseMcMurray:
Inserted 'li' element between 'ul' element and 'ux-timeline-event' elements.
Removed unused 'li.timeline-inverted' element which originated in the ng1 example.
Now generating a new date and a unique ticker number for each event.
Changed badge colours.
Vertically centered the panel arrow.
Updated the Protractor tests.

https://jira.autonomy.com/browse/EL-2941

### DIFF
--- a/docs/app/pages/components/components-sections/timeline/timeline/snippets/app.html
+++ b/docs/app/pages/components/components-sections/timeline/timeline/snippets/app.html
@@ -2,21 +2,23 @@
     <div class="col-md-10">
         <button class="btn button-primary" (click)="addEvent()">Add New Event</button>
         <ux-timeline>
-            <ux-timeline-event *ngFor="let event of events" [badgeColor]="event.color"
-                [badgeTitle]="event.date | date:'EEE LLL d'">
+            <li *ngFor="let event of events">
+                <ux-timeline-event [badgeColor]="event.color"
+                    [badgeTitle]="event.date | date:'EEE LLL d'">
 
-                <div class="m-b-sm">
-                    <span>
-                        <i class="hpe-icon hpe-document-time"></i>
-                        <span>{{event.date | date:'EEEE, MMMM d, y, h:mm:ss a'}}</span>        
-                    </span>
-                </div>
-                <p class="m-b-nil">Ticket
-                    <a [href]="event.url">{{event.id}}</a>
-                    was {{event.action}} by {{event.assignee}}.
-                </p>
+                    <div class="m-b-sm">
+                        <span>
+                            <i class="hpe-icon hpe-document-time"></i>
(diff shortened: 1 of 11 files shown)


donna-marijne:
Assign your PRs to me please Pearse, I'll determine who the reviewer is.

docs/app/pages/components/components-sections/timeline/timeline/snippets/app.html
```diff
@@ -2,21 +2,23 @@
     <div class="col-md-10">
         <button class="btn button-primary" (click)="addEvent()">Add New Event</button>
         <ux-timeline>
-            <ux-timeline-event *ngFor="let event of events" [badgeColor]="event.color"
-                [badgeTitle]="event.date | date:'EEE LLL d'">
+            <li *ngFor="let event of events">
```
ashley-hunter: Why is this li element here? Can the ngFor not just be on the ux-timeline-event element?

ashley-hunter: It seems quite messy and an unnecessary complexity to have to have them insert an li around each event. 

If we take a look at all other libraries like ngx-bootstrap where they have a similar structure eg. accordion, tabs etc.. they don't require anything like this. If the UX timeline component is currently using a UL then perhaps it should just be changed to use a div or something instead removing the requirement for the LI.

docs/app/pages/components/components-sections/timeline/timeline/snippets/app.ts
```diff
@@ -53,7 +59,7 @@ interface TimelineEvent {
     color: string;
     date: Date;
     url: string;
-    id: string;
+    id: number;
```
andyreidz: Why the change to number?

PearseMcMurray: I changed the code to use 'chance.integer()' to return a random number instead of using the same ID for each ticket. It returns an integer so I changed the type of 'id'.

docs/app/pages/components/components-sections/timeline/timeline/snippets/app.html
PearseMcMurray: The 'li' element was introduced becuae of the existing 'ul' element. I will look into replacing the 'ul' as you suggested.

PearseMcMurray:
Replaced 'ul.timeline' with 'div.timeline' and removed the new 'li' element.
Updated Protractor tests.

docs/app/pages/components/components-sections/timeline/timeline/snippets/app.ts
andyreidz: ok, I missed this interface was only part of the doc.

PR #590, opened 2018-02-28
Creating UrlTree to determine if link is active.

PearseMcMurray:
https://jira.autonomy.com/browse/EL-2897

### DIFF
--- a/src/components/navigation/navigation-item/navigation-item.component.ts
+++ b/src/components/navigation/navigation-item/navigation-item.component.ts
@@ -1,7 +1,6 @@
 import { ActivatedRoute, NavigationEnd, Router, RouterState, UrlTree } from '@angular/router';
 import { AfterContentInit, AfterViewInit, Component, ContentChildren, ElementRef, Input, OnDestroy, Optional, QueryList, Renderer2, SkipSelf, HostBinding } from '@angular/core';
 import { filter } from 'rxjs/operators/filter';
-import { Observable } from 'rxjs/Observable';
 import { Subscription } from 'rxjs/Subscription';
 
 @Component({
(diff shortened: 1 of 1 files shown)


src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -1,35 +1,21 @@
...
-    Output,
-    AfterViewInit,
-    AfterContentInit,
-    OnDestroy
-} from '@angular/core';
-import { Router, NavigationEnd, UrlTree } from '@angular/router';
+import { ActivatedRoute, NavigationEnd, Router, RouterState, UrlTree } from '@angular/router';
+import { AfterContentInit, AfterViewInit, Component, ContentChildren, ElementRef, Input, OnDestroy, Optional, QueryList, Renderer2, SkipSelf, HostBinding } from '@angular/core';
 import { filter } from 'rxjs/operators/filter';
+import { Observable } from 'rxjs/Observable';
```
ashley-hunter: I don't think this is used so can probably be removed

src/components/navigation/navigation-item/navigation-item.component.ts
```diff
@@ -85,14 +68,20 @@ export class NavigationItemComponent implements AfterViewInit, AfterContentInit,
         this._childrenChanges = this._children.changes.subscribe(() => this.setIndentWithoutArrow());
     }
 
-    ngOnDestroy () {
+    ngOnDestroy() {
```
ashley-hunter: Add return type for consistency

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Import removed.

src/components/navigation/navigation-item/navigation-item.component.ts
PearseMcMurray: Return type added.

PR #594, opened 2018-03-02
EL-3029 - Removing dark styling

ashley-hunter:
- Removing Dark Styling (anything under .body-dark)
- Making existing styles follow stylelint rules - fixing formatting, missing semicolons, removing some < IE 10 vendor specific prefixes
- Adding grunt task to lint the stylesheets now (with the exception of the icons as that stylesheet is autogenerated)


ashley-hunter:
Issues should now be addressed - comments inline:

New stylelint:stylesheets task not called from 'grunt lint'.
**Added to lint task**

buttons.less line 313
Two lines merged into one.
**I have merged this with the previous statement to prevent duplication - but this is expected**

navigation.less line 1713
Two lines merged into one.
**Expected**

Some deleted lines. Just checking they are deliberate changes:
**These were duplicated rules remove is expected**

Some other occurrences of .body-dark:
**Removed these**

PR #599, opened 2018-03-07
EL-3000 Change colour of icon if 'iconColor' value is defined.

PearseMcMurray:
Remove unused 'band' element.
Allow title of 'Background Color' field to be changed.

https://jira.autonomy.com/browse/EL-3000


PearseMcMurray:
Labels in Customize Example panel changed to the names of the properties being set.
Components rearranged so that labels and components are on the same line.

PR #600, opened 2018-03-07
Removing fixed ngx-bootstrap version to allow newer versions

ashley-hunter:
In response to the issue Thomas was having on slack I have relaxed the ngx-bootstrap version requirement to prevent multiple versions of it being installed.

His app depended on ngx-bootstrap "^2.0.0-beta.8" which was installing version 2.0.2. The ux-aspects package depended on "2.0.0-beta.8" exact - so it was installing multiple versions causing StaticInjectorErrors in a production build

### DIFF
--- a/src/package.json
+++ b/src/package.json
@@ -32,7 +32,7 @@
     "chart.js": "~2.6.0",
     "ng2-charts": "^1.6.0",
     "ng2-file-upload": "^1.3.0",
-    "ngx-bootstrap": "^2.0.0-beta.8",
+    "ngx-bootstrap": "^2.0.0",
     "angular-tree-component": "^5.2.0",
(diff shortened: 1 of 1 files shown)


donna-marijne:
Can we switch to the 2.0 release version of bootstrap? Or does that cause problems?

ashley-hunter:
Updated to 2.0.0 - the only time this could cause an issue is if someone explicitly had 2.0.0-beta.8 exact listed in their package.json, in which case they would get the same issue as Thomas. 

However anyone who used our seed projects should not have those issues as ngx-bootstrap isn't listed as a dependency explicitly so I don't see this causing any issues and worst case they simply update ngx-bootstrap

PR #601, opened 2018-03-08
EL-3028 - Floating Action Button Tooltips

ashley-hunter:
Had to rename the directive property, it was previously using tooltip and tooltip placement but as they are attributes also used by ui-bootstrap, tooltips were being created and shown twice.


ashley-hunter:
Tooltips added to example

ashley-hunter:
Line breaks added

PR #604, opened 2018-03-08
Removed 'accordion-chevron' classes from places where it had incorrec…

PearseMcMurray:
…tly been added.


PearseMcMurray:
The class was added, incorrectly, in the previous PR (https://github.com/UXAspects/UXAspects/pull/603/files). As I understand it, the only place the chevron should appear in the Angular version of the component is in the Accordion section itself. The chevron in the AngularJS version is added through a template. This hasn't been changed.

donna-marijne:
Could you add it to the correct component instead?

PR #605, opened 2018-03-09
EL-2994 - Lerna

donna-marijne:
* theme-hpe - has an ~~optional~~ dependency on the private package. Compiles less with the imported variables and resources. Creates the bower .tgz file for Artifactory.
* Updated docs to build into dist/keppel when the `build` command is used, and dist/hpe when the `build:hpe` command is used. Stylesheet used is controlled by a Less global variable defined in webpack. Also added a compress step which creates the documentation site tarball for the specified theme under target/gh-pages.
* Added version scripts to bump the version of all packages (including legacy bower packages).
* Fixed the library build to avoid the dist root directory. There is a patch step to remove the `files` property from the copied package.json. A postpack script does `npm pack` in the dist directory and copies the tgz file into the main directory, ensuring that `npm pack` command can still be used.
* Added jasmine report step
* `grunt test` no longer fails if any tests fail, so that the HTML report can be produced.
* Added `gather` script to collect all release artifacts.
* ux-aspects-hpe is no longer an optional dependency, since lerna does not support it.

Build script changes will need to be merged before testing, so this will require more iterations.

Dashboard example is not working correctly - need some advice on this.


donna-marijne:
Any idea what's up with the dashboard?

ashley-hunter:
Yeah, it doesn't seem to be getting any of its styling, for example the ux-dashboard-widget when I compare it is missing the following styles:

```css
ux-dashboard-widget {
    position: absolute;
    display: inline-block;
    overflow: hidden;
}```

Although I'm not sure why this would be happening, the code and styles seem fine

PR #606, opened 2018-03-12
EL-3034 - Floating action buttons

ashley-hunter:
- Migrating Floating Action Button
- Reworked floating action button to prevent closing when mouse leaves (also greatly simplified code to allow style sharing with migrated version)
- Updated ng1 protractor tests
- Added Angular FAB protractor tests
- Added BrowserAnimationsmodule to all plunkers as it is beginning to get wider use in components


src/components/floating-action-buttons/floating-action-buttons.component.ts
```diff
@@ -0,0 +1,41 @@
...
+import { FloatingActionButtonsService } from './floating-action-buttons.service';
+
+@Component({
+    selector: 'ux-floating-action-buttons',
+    templateUrl: './floating-action-buttons.component.html',
+    providers: [FloatingActionButtonsService],
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    preserveWhitespaces: false,
+    animations: [
+        trigger('fabAnimation', [
```
donna-marijne: The main concern I have with this is whether we can override the animations in the quantum stylesheet for Micro Focus (my understanding is that we cannot).

src/ng1/directives/floatingActionButton/floatingActionButton.controller.js
```diff
@@ -1,9 +1,45 @@
-FloatingActionButton.$inject = ["$scope"];
+export default class FloatingActionButton {
 
-export default function FloatingActionButton() {
-	var vm = this;
+	constructor($scope, $document, $element) {
```
donna-marijne: Can you check the indentation in this file and ensure that it is consistent with the rest of the project? (Spaces instead of tabs.)

docs/app/pages/components/components-sections/buttons/floating-action-button/floating-action-button.component.html
```diff
@@ -0,0 +1,85 @@
...
+
+<hr>
+
+<div class="row uxd-customize-example">
+    <div class="col-md-12">
+        <accordion>
+            <accordion-panel heading="Customize Example...">
+                <div class="row">
+                    <div class="col-md-3">
+                        <ux-radio-button [(ngModel)]="direction" option="top">Top</ux-radio-button>
```
donna-marijne: Can you change this to refer to the API values, so that authors can directly associate it with the code that needs to be written?

````
direction    ( ) top   (x) right   ( ) bottom   ( ) left
````

File: docs/app/pages/components/components-sections/buttons/floating-action-button/floating-action-button.component.html [hunk 4]
```diff
@@ -0,0 +1,85 @@
...
+                    </div>
+                </div>
+            </accordion-panel>
+        </accordion>
+    </div>
+</div>
+
+<p>
+    Floating Action Buttons are contained in a <code>ux-floating-action-buttons</code> component. 
+    Each action button should be added as child using the <code>ux-floating-action-button</code> component.
```
donna-marijne: Since they aren't explicitly `button`s, and the example doesn't show it, maybe mention that the action should be attached via a `(click)` attribute.

docs/app/pages/components/components-sections/buttons/floating-action-button/floating-action-button.component.html
```diff
@@ -0,0 +1,85 @@
...
+<uxd-api-properties tableTitle="Inputs">
+    <tr uxd-api-property name="direction" type="string" defaultValue="top">
+        Defines direction items should be displayed. Possible values are <code>top</code>, <code>right</code>, <code>bottom</code> or <code>left</code>.
+    </tr>
+</uxd-api-properties>
+
+<p>The following input can be used to customize the appearance of the <code>ux-floating-action-button</code> component:</p>
+
+<uxd-api-properties tableTitle="Inputs">
+    <tr uxd-api-property name="icon" type="string">
```
donna-marijne: This should be `required="false"`.

ashley-hunter: Adding this doesn't change the UI, a "Required" message is only shown when the value is true, by default it is false

src/components/floating-action-buttons/floating-action-buttons.component.ts
ashley-hunter: You are correct, we cannot override it by using the stylesheets. I have proposed a solution in my latest changes, if you like it I can create a jira to update our docs/seed project with how to use it, if not I can revert to CSS animations but we will lose the stagger in the button animations. 

There are cases where it would be good for our components to know what the active theme is, this would allow us to show and hide certain elements based on the theme or use different animations etc..

Currently teams set the theme by injecting the colorService eg:

```
constructor(colorService: ColorService) {
   colorService.setColorSet(colorSets.microFocus);
}
```

I have created a ThemeService that would allow the teams to do something similar:

```
constructor(themeService: ThemeService) {
   themService.setTheme(Theme.MicroFocus);
}
```

Upon setting the theme this would automatically update color palette and allow any components to subscribe to an active theme and update anything based on the active theme and use the appropriate animations etc..

(I also have it so even if they are setting the theme only by the colorService it will update the active theme based on the color set provided to prevent any issues with teams not updating to use the theme service)

ashley-hunter:
Review changes addressed

ashley-hunter:
The colorService changes:
- Missing injectable decorator which was a bug so I fixed it
- Added some logic to ensure it stays in sync with the ThemeService
- Left in debugger (woops) now removed it.

Regarding the shadow, that was an oversight, I will add it in again

donna-marijne:
I don't understand the color/theme service changes. Floating action button doesn't use this as far as I can tell. If it's an unrelated bug can you raise a ticket and make another PR?

ashley-hunter:
- Added in missing shadows
- Reverted theme/color changes - they weren't used by FAB, I added it to provide a way to handle different animations etc.. if we needed them in future, but we can added something like that in future if and when we need it.

src/components/floating-action-buttons/floating-action-buttons.component.ts
donna-marijne: I totally missed this explanation, sorry.

donna-marijne: Let's make a ticket on this topic and discuss it during storytime. In the meantime we can keep these animations so that it is consistent with the ng1 buttons.

ashley-hunter:
No problem! Will do! Thanks

PR #611, opened 2018-03-14
Store icon colour in NotificationRef object. Not overloading the exis…

PearseMcMurray:
…ting 'backgroundColor' property as the MF documentation refers to the 'iconColor' property in the Notification Configuration object.

https://jira.autonomy.com/browse/EL-3000


PearseMcMurray:
Replace hardcoded colour values with use of Color Service.

PR #614, opened 2018-03-14
Release 1.5.7

donna-marijne:
https://jira.autonomy.com/browse/EL-3030


donna-marijne:
Yes - thank you for noticing that

PR #618, opened 2018-03-16
Adding ability to specify a template for dynamic facet options

victor-ac:
## Context
- Custom Facets provide the great ability to customize facet contents to application needs, but they don't come with the functionality already embedded in Dynamic Facets (e.g. ability to limit the number of facet options shown, show/hide autocomplete textbox depending on this limit, show an overflow tooltip for long names, highlight value when it's hovered). In the application I'm working, we want to keep this functionality while having the ability to customize the markup used for each facet option. This PR addresses this need.

## Changes
- Added attribute 'facet-option-template' to 'facet-dynamic' directive. If the template is not informed, the existing facet option layout is used as default
- Removed 'name' and 'count' attributes from 'dynamic-facet-option' directive, since 'data' is now being passed and contains the entire option object
- Updated documentation to be consistent with changes made
- Added unit tests to exercise scenarios with and without the facet-option-template attribute


src/ng1/ux-aspects-ng1.module.js
```diff
@@ -123,6 +123,7 @@ import './services/windowCommunicationService/windowCommunication.module.js';
 
 import './templates/angular-breadcrumb/angular-breadcrumb.html';
 import './templates/ui-bootstrap/pagination/pagination.html';
+import './directives/facets/facetOption/facetOptionDefault.html';
```
victor-ac: I didn't want to create this file under /templates because it relates to the ng1 facets component, but I'm placing the import along with other templates because the import for the ng1 facets module doesn't include templates.

victor-ac:
@donna-marijne or @ashley-hunter, let me know if you have any questions.

donna-marijne:
Thank you for your PR! I will review this later today.

docs/app/pages/components/components-sections/facets/dynamic-facets-ng1/dynamic-facets-ng1.components.html
```diff
@@ -60,6 +69,11 @@
 <p>Set the <code>placeholder</code> attribute to a string property to set the placeholder text that will be shown in the text
     search input control.</p>
 
+<h4>Customizing the Dynamic Facet Option Template</h4>
```
donna-marijne: Can you remove this heading? It gives the appearance that the sample code only applies to the template subsection. The current documentation standard is to list attributes in a tabular format for clarity, but since this is a deprecated component I don't think it's worth making that change now.

docs/app/pages/components/components-sections/facets/dynamic-facets-ng1/snippets/dynamic-facets.snippet.html
```diff
@@ -1,7 +1,15 @@
...
-                   placeholder=" 'Find Author' " update-callback="vm.longListCallback" max-length="100" max-lines="3" disable-return="'false'" >
+                   placeholder=" 'Find Author' " update-callback="vm.longListCallback" max-length="100" max-lines="3" disable-return="'false'"
+                   facet-option-template="'options-with-star'" >
     </facet-dynamic>
     <!--...-->
-</facet-container>
\ No newline at end of file
+</facet-container>
+
+<script type="text/ng-template" id="options-with-star">
```
donna-marijne: We usually document templates in a separate tab, partly because the syntax highlighter doesn't handle HTML in script tags properly.

docs/app/pages/components/components-sections/facets/dynamic-facets-ng1/dynamic-facets-ng1.components.html
```diff
@@ -8,6 +8,15 @@
     To use the Dynamic Facet, use the <code>facet-dynamic</code> directive instead of the <code>facet</code> directive. Set
     the
     <code>facet-options</code> attribute with an object containing a <code>title</code> attribute, and an <code>options</code>    property containing collection of objects that follow the interface for the <code>facet-option</code> defined above.
+</p>
+
+<blockquote>
+    <p><strong>Note</strong>: For this type of facet, however, each one of these objects is
+    accessible via the <code>data</code> attribute. Since both <code>name</code> and <code>count</code> are properties within that object,
+    they should be accessed as <code>data.name</code> and <code>data.count</code>, respectively.</p>
```
donna-marijne: This refers to a `data` attribute. Is this actually an attribute somewhere, or does this refer to the scope variable available to the custom template? If the latter, this piece of information can be relocated to the paragraph mentioning the custom template.

victor-ac: This maps to the `data` attribute of the `dynamic-facet-option` directive, but it's not exposed to `dynamic-facet`. So, yes, it will be available as a scope variable in the custom template, but I wanted to place this note here to warn the user that following the `facet-option` interface outlined above won't work in this case.

docs/app/pages/components/components-sections/facets/dynamic-facets-ng1/dynamic-facets-ng1.components.html
victor-ac: Done

docs/app/pages/components/components-sections/facets/dynamic-facets-ng1/snippets/dynamic-facets.snippet.html
victor-ac: Done

victor-ac:
@donna-marijne I just realized that I need to also provide the option to customize the typeahead template. Without that, the entry will be incomplete:
image

Let me work on that and push another patch later, but let me know if you agree with my comment regarding the `data` attribute.

docs/app/pages/components/components-sections/facets/dynamic-facets-ng1/dynamic-facets-ng1.components.html
donna-marijne: Thanks - it's not very clear, as worded. Could you rewrite the preceding paragraph to provide correct information and remove the note? Also care should be taken to distinguish between HTML attributes and object properties (I appreciate that the original documentation makes this mistake too).

victor-ac:
@donna-marijne I addressed your comment, added the ability to customize the typeahead template, and also the selected facet option.

donna-marijne:
For reference, this is the codepen that I'm using to validate these changes. Feel free to amend this if I have overlooked something.

https://codepen.io/donnamarijne/pen/NYzZmK?editors=1010

donna-marijne:
Hi Victor. Go ahead and push up your code with the single template and I can help to resolve the problems with it. I may have some time during the current sprint, but if not I will reserve time during the next (starting 10 May).

victor-ac:
Hi @donna-marijne, let me know if you have any questions. With the recent changes, I'm still having problems keeping the checkbox state when switching to a different view (which, in other words, changes the template being loaded for the facets)

donna-marijne:
Currently we are very busy, so I'll schedule this in the next sprint (beginning thursday). In the meantime, a codepen example showing how you are going to use this feature would help it along (you could use the one above as a starting point).

donna-marijne:
Superseded by https://github.com/UXAspects/UXAspects/pull/723.

PR #619, opened 2018-03-20
EL-2974 - Reorderable Table

ashley-hunter:
JIRA: https://jira.autonomy.com/browse/EL-2974

- Added new uxSortable directive (can be used on any element not just tables, works with ngFor and can specify custom drag handles

- Added new dependency - dragula to handle some of the complex drag logic (along with type definitions)

- Added documentation example of reorderable table and plunker

- A few of our Usage popovers used "Module name" instead of "Module" so I renamed them all to say "Module" for consistency

- Set all our AngularJS table sections to use OnPush change detection as they aren't doing anything Angular related so no point running change detection on them, give us a slight performance improvement as its a very large docs section and can be a bit slow

- Protractor tests to come along with MF docs section and Quantum styles


docs/app/pages/components/components-sections/tables/reorderable-table/reorderable-table.component.html
```diff
@@ -0,0 +1,97 @@
...
+        This can be used to bind a dataset to the list of items that can be reordered. This can be used along with <code>ngFor</code> to keep the 
+        model in sync. If specifed each reorderable item must have a <code>uxSortableModel</code> attribute with the value that item represents.
+    </tr>
+</uxd-api-properties>
+
+<uxd-api-properties tableTitle="Outputs">
+
+    <tr uxd-api-property name="sortableModelChange" type="any[]">
+        This event will be triggered when the order changes and will contain an updated dataset containing the items in their current
+        order. This can also be used to ensure <code>ngFor</code> remains in sync.
```
donna-marijne: Regarding "This can also be used to ensure <code>ngFor</code> remains in sync." - when and why would this be needed?

docs/app/pages/components/components-sections/tables/reorderable-table/reorderable-table.component.html
```diff
@@ -0,0 +1,97 @@
...
+</p>
+
+<p>
+    The following input can be to provide the dataset:
+</p>
+
+<uxd-api-properties tableTitle="Inputs">
+    <tr uxd-api-property name="sortableModel" type="any[]">
+        This can be used to bind a dataset to the list of items that can be reordered. This can be used along with <code>ngFor</code> to keep the 
+        model in sync. If specifed each reorderable item must have a <code>uxSortableModel</code> attribute with the value that item represents.
```
donna-marijne: This might need some rewriting. We should mention that it's a 2-way binding (since it's listed under Inputs) and that it's updated to reflect the order of items. If mentioning ngFor, it could be clearer and specify that this should be the same collection that ngFor iterates over.

What happens if sortableModel is not specified? Is it useful to emphasise that this is optional?

docs/app/pages/components/components-sections/tables/reorderable-table/reorderable-table.component.html
```diff
@@ -0,0 +1,97 @@
...
+
+<uxd-api-properties tableTitle="Outputs">
+
+    <tr uxd-api-property name="sortableModelChange" type="any[]">
+        This event will be triggered when the order changes and will contain an updated dataset containing the items in their current
+        order. This can also be used to ensure <code>ngFor</code> remains in sync.
+    </tr>
+
+    <tr uxd-api-property name="sortStart" type="Element">
+        This event is triggered when a user begins dragging an item. The event will contain the element being moved.
```
donna-marijne: Is it possible to include the source data for events as well as/instead of the DOM element? Also, can this be cancelled to prevent dragging?

docs/app/pages/components/components-sections/tables/reorderable-table/reorderable-table.component.html
```diff
@@ -0,0 +1,97 @@
...
+    </tr>
+
+    <tr uxd-api-property name="sortCancel" type="Element">
+        This event is triggered when the item being dragged is returned to the same location as it began. The event will contain the element that was being moved.        
+    </tr>
+
+    <tr uxd-api-property name="sortEnd" type="Element">
+        This event is triggered when a user has relocated an item. The event will contain the element that was moved.
```
donna-marijne: Is this the same as sortableModelChange?

ashley-hunter:
Changes made:

- Changed sortable -> reorderable
- Fixed indentation - i'm not sure what is going on here, VS code has done this quite a few times recently for no apparent reason.
- Improved some descriptions as mentioned - if you have any further suggestions let me know
- Changed events to emit both the element and model associated with it if available
- sortableModelChange is similar to sortEnd (now reorderableModelChange and reorderEnd), but the reorderableModelChange provides the updated dataset, not the element/model that was moved which may be useful to determine which item was just moved.

**Slow IE issue**

Yeah, I did a bit of investigation on this, I have made some performance improvements here. I have realised we can specify the container for the cloned element so we don't need to inline all the styles - the only case is fixing the widths of table cells which I have changed it to do.

In a more general issue, the table section is really slow. In particular the AngularJS components. Most notably the following are completely hogging the CPU constantly:

- Detail row header - I have changed to use new fixed header table instead of deprecated version as the old one was crazy slow - also removed unneeded overflow tooltip from example
- Preview Pane Window - JScrollPane need I say any more

Hopefully performance should be a bit better now - at least in plunker where all the other slow stuff isn't running.

src/directives/reorderable/reorderable.directive.ts
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * This function will get all computed styles on the source element
+     * and apply them inline to the target element and this will repeat for all 
+     * all of its children. This ensures the element will be identical regardless
+     * of the location in the DOM tree
+     * @param source the element to get all the computed styles from
+     * @param target the element to apply all the inline styles to - this should be a clone of the source element
+     */
+    // private setStyles(source: Element, target: Element): void {
```
donna-marijne: Remove commented code.

docs/app/pages/components/components-sections/tables/reorderable-table/reorderable-table.component.ts
```diff
@@ -0,0 +1,70 @@
...
+                document: `Document ${idx}`,
+                author: chance.name(),
+                date: chance.date({ year: 2018 }) as Date,
+                completed: chance.integer({ min: 10, max: 100 }),
+                active: chance.bool()
+            });
+        }
+    }
+
+    movedown(data: ReorderableTableData, index: number, event: MouseEvent): void {
```
donna-marijne: I think it should be `event: KeyboardEvent` here.

ashley-hunter:
Well spotted! Issues should now be fixed! Thanks

docs/app/pages/components/components-sections/tables/reorderable-table/snippets/app.ts
```diff
@@ -0,0 +1,50 @@
...
+                document: `Document ${idx}`,
+                author: chance.name(),
+                date: chance.date({ year: 2018 }) as Date,
+                completed: chance.integer({ min: 10, max: 100 }),
+                active: chance.bool()
+            });
+        }
+    }
+
+    movedown(data: ReorderableTableData, index: number, event: MouseEvent): void {
```
donna-marijne: Also need to change here.

ashley-hunter:
Completely forgot about the snippet. Should be updated now, thanks

PR #624, opened 2018-03-23
Fixing dragula plunker issue

ashley-hunter:
Turns out ng-packagr keeps `require` statements dynamic so I had to use a different workaround to allow us to use the dragula library. Should work as expected now.


donna-marijne:
Are we shipping the dragula code with UX Aspects? Or how does the code get loaded into plunker?

ashley-hunter:
Made changes to make dragula an external dependency

PR #648, opened 2018-04-12
Fixing build issue

ashley-hunter:
Did a bit of digging and the build issue seems to be being caused by ng-packagr which has a dependency on uglify-js^3.0.7.

One of the recent uglify releases seems to have broken this. I have fixed the version of uglify so we don't use a broken version. Fixes the issue at least when I try it on my local machine

### DIFF
--- a/configs/loaders/uglifyjs-loader.js
+++ b/configs/loaders/uglifyjs-loader.js
@@ -1,0 +1,17 @@
+const { minify } = require('uglify-js');
+
+module.exports = function (source) {
+
+    if (this.cacheable) {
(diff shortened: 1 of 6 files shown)


ashley-hunter:
No problem, good spot, will look into it now

ashley-hunter:
- Updating package-lock.json
- Removed uglify-loader, it is really old and outdated
- Replaced with our own uglifyjs-loader (there are no other uglify loaders available on npm that I can see - only webpack plugins which are no use for what we need) that uses the latest version of uglify and fixes build issues

PR #649, opened 2018-04-13
EL-2970 - Expanding textarea migration

ashley-hunter:
- Created new Angular directive
- Added documentation section and deprecated old section
- Added protractor tests

Old component had two attributes which I have removed:
- max lines - This is better just set using css as that gives more flexibility eg, flexbox and % etc..
- disable return - This was fairly useless before as you could paste stuff in that had multiple lines. In the rare case that they would want this, this should be handled using ngModel rather than part of this directive.

### DIFF
--- a/docs/app/data/components-page.json
+++ b/docs/app/data/components-page.json
@@ -535,8 +535,24 @@
                 },
                 {
                     "title": "Expanding Text Area",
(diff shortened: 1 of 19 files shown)


ashley-hunter:
Line breaks should now be added

PR #656, opened 2018-04-18
EL-2972 - side modal

donna-marijne:
* Added `width` property to control expanded width.
* Changed `top` property to a string to accept CSS units (for consistency with `width`).
* Renamed `open` and `close` functions to avoid a name collision with the `open` property.

If this is fine then I'll make a final PR with tests, and also update the item display panel to use this component internally.

https://jira.autonomy.com/browse/EL-2972


src/components/side-panel/side-panel.component.ts
```diff
@@ -0,0 +1,131 @@
...
+                target = target.parentElement;
+            }
+        }
+
+        if (hidePanel) {
+            this.closePanel();
+        }
+    }
+
+    @HostListener('document:keyup.escape')
```
ashley-hunter: Could this be added to the `closePanel` function instead of having a separate function?

src/components/side-panel/side-panel.component.ts
```diff
@@ -0,0 +1,131 @@
...
+        console.log('document:click');
+
+        let target = event.target as HTMLElement;
+
+        // if the target node is the HTML tag, then this was triggered by scrolling and we should not close the panel
+        if (target.nodeName === 'HTML') {
+            return;
+        }
+
+        let hidePanel = true;
```
ashley-hunter: I'm not sure, but we might be able to simplify this by quite a bit. I think we could use the elementRef to check if it contains the target instead, eg:

```typescript
if (!this._elementRef.nativeElement.contains(event.target)) {
   this.closePanel();
}
```

src/components/side-panel/side-panel.component.ts
```diff
@@ -0,0 +1,131 @@
...
+        this._service.close();
+    }
+
+    @HostListener('document:click', ['$event'])
+    clickHandler() {
+        if (!this.open || !this.closeOnExternalClick) {
+            return;
+        }
+
+        console.log('document:click');
```
ashley-hunter: Can this be removed?

src/components/side-panel/side-panel.component.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    get hostWidth() {
+        return this.inline ? '100%' : this.width;
+    }
+
+    private _subscription: Subscription;
+
+    constructor(
+        private _service: SidePanelService,
+        private _renderer: Renderer2
```
ashley-hunter: Renderer doesn't seem to be used.

src/components/side-panel/side-panel.module.ts
```diff
@@ -0,0 +1,21 @@
...
+    SidePanelCloseDirective
+];
+
+@NgModule({
+    imports: [
+        CommonModule
+    ],
+    exports: EXPORTS,
+    declarations: EXPORTS,
+    providers: [SidePanelService]
```
ashley-hunter: Should this service be scoped to the SidePanelComponent instead? If there are multiple side panels I think this would cause them all to open and close whenever any of them did.

donna-marijne:
Thank you for the feedback! I have made these changes.

ashley-hunter:
I think stylelint may also be failing in side-panel.component.less

docs/app/pages/components/components-sections/panels/side-panel/side-panel.component.ts
```diff
@@ -1,14 +1,54 @@
...
+    animate = true;
+    closeOnExternalClick = false;
+
+    get attachTo(): string {
+        return this._attachTo;
+    }
+
+    set attachTo(value: string) {
+        this._attachTo = value;
+        if (value === 'window') {
```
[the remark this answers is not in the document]
donna-marijne: Yes, but there will most likely be a new documentation section if/when this gets added to the MF site, since the example would probably be based on something from the style guide.

donna-marijne:
Thanks for noticing, I have fixed the stylelint errors too.

PR #657, opened 2018-04-18
EL-2969 - Card Tabs

ashley-hunter:
- Migrated Card Tabs
- Added Documentation & plunker
- Deprecated AngularJS section
- Added e2e tests

Additionally:
- Updated our dev webpack config, my incremental builds were really slow and freezing up a lot. I realised our dev config was not using the official Angular webpack loader, so I have changed it to now use them seeing decent improvements. It now also matches the production build and our MF webpack configs. Also removed json-loader config as this is now built into webpack since v2

- The resize service and resize directive got a few tweaks:
    - Added destroy functions to unsubscribe correctly
    - Changed service to use BehaviorSubject rather than Subject and set initial sizes earlier - this ensures we get the size faster and the event will always fire initially - as before there was a small chance it would not fire initially
    - Added scoped service to the uxResize directive. This will allow the scoped service to be disposed when the uxResize directive is destroyed freeing up some memory.

### DIFF
--- a/docs/app/pages/components/components-sections/tabs/card-tabs/card-tabs.component.html
+++ b/docs/app/pages/components/components-sections/tabs/card-tabs/card-tabs.component.html
@@ -1,0 +1,71 @@
+<ux-card-tabset [position]="position">
+
+    <ux-card-tab class="card" *ngFor="let tab of tabs">
+        <h3 class="card-title">{{ tab.title }}</h3>
+        
+        <div class="card-image" [ngClass]="tab.image"></div>
+        
+        <h3 class="card-value">{{ tab.value }}<small>{{ tab.unit }}</small></h3>
+        <h4 class="card-subtitle">{{ tab.subtitle }}</h4>
+
+        <ng-template uxCardTabContent>
+            <h4>{{ tab.title }}</h4>
+            <p>{{ tab.content }}</p>
+        </ng-template>
+    </ux-card-tab>
+
+</ux-card-tabset>
+
+<hr>
+
+<div class="row uxd-customize-example">
+    <div class="col-md-12">
+        <accordion>
+            <accordion-group class="accordion-chevron" heading="Customize Example...">
+                <div class="row uxd-customize-row">
+                    <div class="col-md-4 col-sm-12">
+                        <label>Position</label>
+                        <ux-radio-button [(value)]="position" option="top">top</ux-radio-button>
+                        <ux-radio-button [(value)]="position" option="bottom">bottom</ux-radio-button>
+                    </div>
+                </div>
+            </accordion-group>
+        </accordion>
+    </div>
+</div>
+
+<p>
+    Card tabs are defined inside a <code>ux-card-tabset</code> component using a <code>ux-card-tab</code> component.
+    Inside the <code>ux-card-tab</code> component insert any content you want to display in the card.
+</p>
+<p>
+    Also inside the <code>ux-card-tab</code> component an element with a <code>uxCardTabContent</code> directive should be defined which will contain
+    the content that will be shown when that card is selected.
+</p>
+
+<p>
+    If the <code>ux-card-tabset</code> has more cards than can be displayed in the space available arrow buttons will appear
+    to allow the user to scroll through all the cards.
+</p>
+
+<p>
+    The following can be used to configure the appearance of the card tabs:
+</p>
+
+<uxd-api-properties tableTitle="Inputs">
+    <tr uxd-api-property name="position" type="string" defaultValue="top">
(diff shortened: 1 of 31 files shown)


docs/app/pages/components/components-sections/tabs/card-tabs/card-tabs.component.html
```diff
@@ -0,0 +1,71 @@
...
+
+<hr>
+
+<div class="row uxd-customize-example">
+    <div class="col-md-12">
+        <accordion>
+            <accordion-group class="accordion-chevron" heading="Customize Example...">
+                <div class="row uxd-customize-row">
+                    <div class="col-md-4 col-sm-12">
+                        <label>Position</label>
```
donna-marijne: Could you change this to "position" to match the input name?

docs/app/pages/components/components-sections/tabs/card-tabs/card-tabs.component.html
```diff
@@ -0,0 +1,71 @@
...
+        </accordion>
+    </div>
+</div>
+
+<p>
+    Card tabs are defined inside a <code>ux-card-tabset</code> component using a <code>ux-card-tab</code> component.
+    Inside the <code>ux-card-tab</code> component insert any content you want to display in the card.
+</p>
+<p>
+    Also inside the <code>ux-card-tab</code> component an element with a <code>uxCardTabContent</code> directive should be defined which will contain
```
donna-marijne: Could we swap this around so that the directive defines the tab header and the remainder is the content? Then it would be consistent with the Bootstrap tabs:

````html
<tabset>
    <tab>
        <ng-template tabHeading>
            <span class="hpe-icon hpe-platform-chrome"></span> <span>Chrome</span>
        </ng-template>
        <p class="m-t-md">
            Google Chrome is a fast, secure and free web browser, built for the modern web.
        </p>
    </tab>
</tabset>
````

File: docs/app/pages/components/components-sections/tabs/card-tabs/card-tabs.component.html [hunk 3]
```diff
@@ -0,0 +1,71 @@
...
+    If the <code>ux-card-tabset</code> has more cards than can be displayed in the space available arrow buttons will appear
+    to allow the user to scroll through all the cards.
+</p>
+
+<p>
+    The following can be used to configure the appearance of the card tabs:
+</p>
+
+<uxd-api-properties tableTitle="Inputs">
+    <tr uxd-api-property name="position" type="string" defaultValue="top">
```
donna-marijne: This should be `type="'top' | 'bottom'"`.

docs/app/pages/components/components-sections/tabs/card-tabs/card-tabs.component.html
```diff
@@ -0,0 +1,71 @@
...
+
+        <ng-template uxCardTabContent>
+            <h4>{{ tab.title }}</h4>
+            <p>{{ tab.content }}</p>
+        </ng-template>
+    </ux-card-tab>
+
+</ux-card-tabset>
+
+<hr>
```
donna-marijne: hr separates the example from the documentation, so this should be moved down.

docs/app/pages/components/components-sections/tabs/card-tabs/snippets/app.html
```diff
@@ -0,0 +1,35 @@
...
+            <h4>{{ tab.title }}</h4>
+            <p>{{ tab.content }}</p>
+        </ng-template>
+    </ux-card-tab>
+
+</ux-card-tabset>
+
+<hr>
+
+<div class="row uxd-customize-example">
```
donna-marijne: I was thinking about this and I don't think we should include the whole "Customize" section in the code snippets. It adds a lot of code that isn't relevant to the component itself, and could potentially be confusing. The idea is to allow experimentation with the inputs, and when this is launched in plunker that can be done by modifying the variable values directly.

src/components/card-tabs/card-tabset/card-tabset.component.html
```diff
@@ -0,0 +1,18 @@
+<div class="card-tab-content" *ngIf="(tab$ | async); let tab">
+    <ng-container [ngTemplateOutlet]="tab.content"></ng-container>
+</div>
+
+<div class="card-tabs">
+
+    <button class="card-tabs-paging-btn card-tabs-paging-btn-previous" (click)="previous()" *ngIf="offset < bounds.lower">
```
donna-marijne: Can you add `aria-label` to the buttons here?

ashley-hunter:
Review changes should now be addressed.

PR #658, opened 2018-04-19
EL-2973 - Adding side inset panel example

ashley-hunter:
JIRA: https://jira.autonomy.com/browse/EL-2973

Added documentation section that gives an example of how to use a button to toggle the visible state of the splitter panel

### DIFF
--- a/src/styles/splitter.less
+++ b/src/styles/splitter.less
@@ -69,3 +69,44 @@
         }
     }
 }
+
+.splitter-toggle-btn {
+    position: absolute;
+    display: flex;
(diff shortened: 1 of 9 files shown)


src/styles/splitter.less
```diff
@@ -68,4 +68,45 @@
             float: none;
         }
     }
+}
+
+.splitter-toggle-btn {
```
donna-marijne: Is it possible to prefix this class with "ux-"?

ashley-hunter:
Class now prefixed with ux

PR #662, opened 2018-04-20
EL-2969 - Card Tabs - Fixing QA Issues

ashley-hunter:
- Missing Images - images were getting corrupted when processed by both the file-loader and extract-text-webpack-plugin. Moved them to a component stylesheet which resolves this issue.

- IE tab issue - IE was scrolling the element into view even though there is no scrollbars. The only workaround I could come up with was once any tab gains focus it ensures the scroll position is set to 0.

- Focus issue - for some reason firefox just wouldn't show the outline as usual. Had to style the outline manually to make it consistent across all browser.


ashley-hunter:
Using mixin, forgot it existed, thanks!

PR #683, opened 2018-05-04
EL-2942 - Hierarchy bar firefox fix

ashley-hunter:
- Firefox doesn't seem to like having elements with a click event as a child of a button. Replaced it will a div with a tabindex


ashley-hunter:
Changed node content to two separate buttons & updating tests

PR #685, opened 2018-05-04
EL-3093 - Select NgModel support

ashley-hunter:
JIRA: https://jira.autonomy.com/browse/EL-3093

- Made some improvements to select component and moved some logic to setter rather than the rather than onChanges hook to ensure it works with ngModel. Also added some onDestroy cleanup etc.. and using pipe for the rxjs operators.

- Added e2e tests to cover use of ngModel - used ngModel rather than a reactive form as it allowed me to reuse a lot of the test code and reactive forms just uses the same ControlValueAccesor code in our component anyway.

- Removed the pre-push hook as it was breaking the CI build, do think this is something we should look into re-adding in the future


ashley-hunter:
Plunker issue should now be resolved

PR #686, opened 2018-05-08
EL-2965  - drag & drop cards

donna-marijne:
* Moved dragula instance to ReorderableService to enable groups of drop containers.
* Modified drop logic to check source and target before adding/removing data items from the array.
* Modified other event handlers to check that the element is a child before modifying.
* Added `ux-reorderable-container-moving` class to uxReorderable.
* Subscription cleanup.
* Updated Tables > Reorderable Table documentation with `reorderableGroup` details and mentioned the dynamic classes.

Seems to be problems with reliably running plunker with localhost, let me know if you are able to test this.

Protractor tests to be added when this is approved,

https://jira.autonomy.com/browse/EL-2965


src/directives/reorderable/reorderable.service.ts
```diff
@@ -0,0 +1,246 @@
...
+     * Creates the dragula instance with the current config and attaches the events, if not already created.
+     */
+    initialize(): void {
+
+        if (this._instance) {
+            return;
+        }
+
+        // for performance gains lets run this outside ng zone
+        // this._ngZone.runOutsideAngular(this.initDragula);
```
ashley-hunter: Can this be removed?

donna-marijne:
Thanks for the review - redundant code was removed.

PR #693, opened 2018-05-14
EL-3113 - float label

donna-marijne:
All new styles were added for this directive in order to avoid the extra structural parent element.

Protractor tests to be added following code review.

https://jira.autonomy.com/browse/EL-3113

### DIFF
--- a/src/directives/float-label/float-label.directive.ts
+++ b/src/directives/float-label/float-label.directive.ts
@@ -33,7 +33,13 @@ export class FloatLabelDirective implements OnInit, OnChanges, OnDestroy {
             this._renderer.listen(this.input, 'input', this.inputChange.bind(this))
         );
 
+        // Check initial input value
         this.raised = !!this.input.value;
+
+        // Ensure that the `for` attribute is set
+        if (!this._elementRef.nativeElement.getAttribute('for') && this.input.getAttribute('id')) {
+            this._renderer.setAttribute(this._elementRef.nativeElement, 'for', this.input.getAttribute('id'));
+        }
     }
 
     ngOnChanges(): void {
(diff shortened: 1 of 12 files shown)


src/directives/float-label/float-label.directive.ts
```diff
@@ -0,0 +1,71 @@
...
+
+    @Input()
+    mode: 'focus' | 'input' = 'focus';
+
+    private _eventHandles: any[] = [];
+
+    constructor(private _elementRef: ElementRef, private _renderer: Renderer2) { }
+
+    ngOnInit(): void {
+        this._eventHandles.push(
```
ashley-hunter: Could shorten this to:

```typescript
this._eventHandles.push(
    this._renderer.listen(this.input, 'focus', this.inputFocus.bind(this)),
    this._renderer.listen(this.input, 'blur', this.inputBlur.bind(this)),
    this._renderer.listen(this.input, 'input', this.inputChange.bind(this))
);
```

src/directives/float-label/float-label.directive.ts
```diff
@@ -0,0 +1,71 @@
...
+        );
+        this._eventHandles.push(
+            this._renderer.listen(this.input, 'blur', this.inputBlur.bind(this))
+        );
+        this._eventHandles.push(
+            this._renderer.listen(this.input, 'input', this.inputChange.bind(this))
+        );
+
+        // Check initial input value
+        this.raised = !!this.input.value;
```
ashley-hunter: In plunker I tried to initially set `[raised]="true"`, this seemed to have no effect, I assume this might be because raised is getting overridden by this line here?

ashley-hunter: Ah right, I see what you mean now. That's fine then, I can't really think of any better way. Just wanted to run it by you and make sure this was expected

PR #694, opened 2018-05-15
EL-3055 - WCAG Input Controls

ashley-hunter:
Related PR: https://github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/pull/201 & https://github.houston.softwaregrp.net/caf/quantum/pull/464

- Fixing examples - some were using attributes rather than bindings so the value was being passed incorrectly as a string

- Adding keyboard controls to slider
- Adding roles, aria-labels, aria-disabled, aria-checked to radio buttons, checkboxes, toggle switches, number picker, sliders and date picker
- Also fixed issue with date time picker. Previously changing the time had no effect on the value displayed in the input. Now fixed

- Note, I had to use a box shadow for the slider focus, rather than outline. The way the slider elements are caused outline to show around the tooltip as well, eg:

image

Changing this would have been a pretty big job, as I would have had to rework the component and all the MF styles as well. I can revert this if you'd like or display focus a different way, or we can raise a ticket to cover the rework of the tooltips.


src/components/date-time-picker/day-view/day-view.component.html
```diff
@@ -3,16 +3,24 @@
...
-    <tr *ngFor="let row of days">
-      <td *ngFor="let day of row" class="date-cell" [class.current]="day.today" 
-        [class.active]="day.active" [class.preview]="!day.currentMonth" 
-        (mousedown)="select(day.date)" (keyup.enter)="select(day.date)" 
-        tabindex="0">{{ day.date.getDate() }}</td>
+  <tbody role="grid">
+    <tr role="row" *ngFor="let row of days">
+      <td *ngFor="let day of row"
+          class="date-cell"
+          role="gridcell"
```
donna-marijne: Somehow we should indicate that this acts like a button, i.e. you can push enter to select. Since there is already a role, I'm not sure what is best other than adding a child element.

ashley-hunter:
Closing in favor of https://github.com/UXAspects/UXAspects/pull/699/files

PR #696, opened 2018-05-15
EL-3073 - WCAG Tree View & Tree Grid

ashley-hunter:
- Added focus on tab to Tree View examples (3rd party library and not build in so added code to achieve the same effect to the example code)

- Added aria labels and roles to both tree view and tree grid

- Updated tree grid directive to focus the first row when tabbed to


src/ng1/directives/treegrid/treegrid.html
```diff
@@ -1,45 +1,89 @@
...
+                treegrid-row-key-handler
+                treegrid-expand="vm.expand(row)"
+                treegrid-contract="vm.contract(row)"
+                ng-focus="vm.rowFocus(row, $event)"
+                aria-expanded="{{ row.expanding }}"
+                aria-checked="{{ row.selected }}"
+                role="{{ row.canExpand ? 'group' : 'treeitem' }}"
+                tabindex="-1">
+
+                <td ng-repeat="col in vm.columns" ng-class="col.cellClass">
```
donna-marijne: It might help to add aria-labels with the column name on each td, since otherwise Chromevox just reads the whole row out in one go. (Unless `role="treegrid"` improves this.)

src/ng1/directives/treegrid/treegrid.html
```diff
@@ -1,45 +1,89 @@
...
+            <tr ng-repeat="row in vm.getGridRows()"
+                class="clickable hover-actions"
+                ng-class="[row.rowClass, row.selected ? vm.allOptions.select.rowClass : '', row.level > 0 ? 'sub-row' : '']"
+                treegrid-multiple-select-item="row"
+                treegrid-multiple-select-options="vm.allOptions.select"
+                treegrid-row-key-handler
+                treegrid-expand="vm.expand(row)"
+                treegrid-contract="vm.contract(row)"
+                ng-focus="vm.rowFocus(row, $event)"
+                aria-expanded="{{ row.expanding }}"
```
donna-marijne: Should be `aria-expanded="{{ row.expanded }}"`

src/ng1/directives/treegrid/treegrid.html
```diff
@@ -1,45 +1,89 @@
...
-          <treegrid-cell row="row" column="col"></treegrid-cell>
-        </td>
-      </tr>
-    </tbody>
-  </table>
-  <div ng-if="vm.loading" class="treegrid-loading">
-    Loading...
-  </div>
+
+    <table class="table table-hover treegrid" keyboard-navigable-table tabindex="0" ng-focus="vm.focusFirstRow()">
```
donna-marijne: There is [`role="treegrid"`](https://www.w3.org/TR/wai-aria-1.1/#treegrid) in the ARIA spec - can we use that instead? There doesn't seem to be much in the way of supporting examples for treegrid, but it's more similar to the grid role than the tree role. I think that `aria-level` is used to indicate the hierarchy level.

ashley-hunter:
Checkbox change made

ashley-hunter:
Good spot! Thanks. It seem IE doesnt use 'space' as the key name so keydown.space doesn't work. Changed it to check against the keycode instead

PR #708, opened 2018-05-25
EL-3150 - Tab Set

ashley-hunter:
- Created a new WCAG compliant tabset component
- Added documentation example
- Added protractor tests


src/components/tabset/tabset.component.html
```diff
@@ -0,0 +1,45 @@
...
+
+        <a class="nav-link"
+            [id]="tab.id"
+            role="tab"
+            [tabindex]="(tab.active$ | async) ? 0 : -1"
+            (mousedown)="tabset.focused$.next(true)"
+            (focus)="tabset.focused$.next(true)"
+            (blur)="tabset.focused$.next(false)"
+            (mousedown)="tabset.select(tab)"
+            (keydown.arrowleft)="tabset.selectAtIndex(index - 1); $event.preventDefault()"
```
donna-marijne: Possibly need some logic here to skip over disabled tabs?

src/components/tabset/tabset.component.ts
```diff
@@ -0,0 +1,32 @@
...
+    changeDetection: ChangeDetectionStrategy.OnPush,
+    providers: [TabsetService],
+    host: {
+        '[class.tabs-left]': 'stacked === "left"',
+        '[class.tabs-right]': 'stacked === "right"',
+    }
+})
+export class TabsetComponent {
+
+    @Input() type: 'tabs' | 'pills' = 'tabs';
```
donna-marijne: Do we support this? Should probably document or remove.

src/components/tabset/tabset.component.ts
```diff
@@ -0,0 +1,32 @@
...
+    @Input() justified: boolean = false;
+    @Input() vertical: boolean = false;
```
donna-marijne: Also document or remove.

src/components/tabset/tabset.component.html
```diff
@@ -0,0 +1,45 @@
+<!-- Nav tabs -->
+<ul role="tablist"
+    class="nav nav-{{ type }}" 
+    [class.nav-justified]="justified"
+    [class.nav-stacked]="vertical"
+    [class.minimal-tab]="minimal"
+    [attr.aria-label]="ariaLabel">
+
+	<li role="presentation"
```
donna-marijne: I think if you set this to `role="none"` then it will remove the "list item" role implied by the li element. Current this is dictated as "1 of 4 list item."

ashley-hunter:
1. Fixed
2. Non active tab content now uses aria-hidden - ngIf loses tab state when they are changed so I didn't use that
3. This should now be fixed, I believe the issue was the focusIf directive which kept refocusing the active item. Now replaced with an internal tabFocus directive that handles focus more appropriately
4. Removed justified as I realised our styles did not fully support it
5. Added deprecatedFor
6. Removed customize region from snippet
7. Tab selection now takes into account disabled tabs
8. Removed tab type - our styles don't support it
9. Removed vertical - our styles don't support it
10. Tried role="none", but that didn't seem to fix the issue. From what I can tell the only way would be to change it to not be ul & li - however this breaks all bootstrap styling and our styling as it all explicitly uses selectors like `.nav > li > a`. I can redo the tab styles if you prefer as but they will no longer be 'bootstrap' tabs.

ashley-hunter:
Fixed customize section

PR #719, opened 2018-06-11
Allow facets to be sorted in the list

bthurgoo:
Our application required the facet list to be ordered. Added the uxReorderable directive to the list of facets in facet-container allowing them to be re-ordered using drag and drop.

### DIFF
--- a/src/components/facets/facet-container.component.html
+++ b/src/components/facets/facet-container.component.html
@@ -23,13 +23,14 @@
     </div>
 
     <!-- Display Tags For Selected Items -->
-    <div class="facets-selected-list">
+    <div class="facets-selected-list" uxReorderable [(reorderableModel)]="facets">
 
         <!-- Show Selected Tags -->
-        <div class="facet-selected-tag" tabindex="0" *ngFor="let facet of facets" (mousedown)="$event.preventDefault()" (click)="deselectFacet(facet)" (keyup.enter)="deselectFacet(facet)">
+        <div class="facet-selected-tag" tabindex="0" *ngFor="let facet of facets" (mousedown)="$event.preventDefault()" (click)="deselectFacet(facet)" (keyup.enter)="deselectFacet(facet)"
+             [uxReorderableModel]="facet">
 
             <!-- Display Label -->
-            <span class="facet-selected-tag-label">{{ facet.title }}</span>
+            <span class="facet-selected-tag-label" uxReorderableHandle>{{ facet.title }}</span>
 
             <!-- Display Remove Icon -->
             <span class="hpe-icon hpe-close"></span>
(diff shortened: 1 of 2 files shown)


src/components/facets/facet-container.component.html
```diff
@@ -23,13 +23,14 @@
     </div>
 
     <!-- Display Tags For Selected Items -->
-    <div class="facets-selected-list">
+    <div class="facets-selected-list" uxReorderable [(reorderableModel)]="facets">
```
ashley-hunter: Thanks very much for the PR, works great! I have a few very minor suggestions:

1. We try to retain existing default functionality when adding new features such as this, so by default facet items should remain fixed and reordering should be enabled using an input. This way anyone already using this component won't automatically get reorderable facets if they don't want them.

Now currently our Reorderable directive doesn't have the option to disable reordering, but this should be very easy so add:

**reorderable.directive.ts**
```typescript
@Input() reorderingDisabled: boolean = false;
```
and add some logic to the canMove function

```typescript
canMove(element: Element, container: Element, handle: Element): boolean {
    if (this.reorderingDisabled) {
        return false;
    }

    return this.handles.length === 0 ? true : !!this.handles.find(_handle => _handle.nativeElement === handle);
}
```

Now we can add an input to the facet-container component to enabled reordering:

**facet-container.component.ts**

```typescript
@Input() facetsReorderable: boolean = false;
```

And now only allow reordering when facetReordering is enabled

**facet-container.component.html**

```html
<div class="facets-selected-list" uxReorderable [reorderingDisabled]="!facetsReorderable [(reorderableModel)]="facets>
```


2. We should emit the facetsChange event when reordering occurs so the the application can perform events on reorder. This can be done by adding the following `(reorderableModelChange)="facetsChange.emit(facets)"` to the `.facets-selected-list` element, eg:

```html
<div class="facets-selected-list" uxReorderable [reorderingDisabled]="!facetsReorderable" [(reorderableModel)]="facets" (reorderableModelChange)="facetsChange.emit(facets)">
```

3. This is just a slight appearance change. It could be helpful to slightly alter the display of a tag that is being dragged eg:

image

The reorderable directive adds a `.ux-reorderable-moving` class to any element being reordered so we should just be able to add some styling, for example, reducing the opacity:

**facet-container.component.less**

```less
.facet-selected-tag {
    margin-right: 5px;
    margin-bottom: 5px;
    cursor: pointer;
    outline: none;
    max-width: 100%;
    color: @tag-input-tag-color;
    background-color: @tag-input-tag-bg;
    border: 1px solid @tag-input-tag-border;
    border-radius: 1px;
    padding: 0 5px;

    &.ux-reorderable-moving {
         opacity: 0.5;
    }
```

4. With these changes we have added a new input to both the reorderable directive and our facet container component. It would be great if the documentation was updated just to include these two inputs.

Thanks very much for your contribution, if you have any questions or issues let me know!

bthurgoo:
Implemented requested changes to allow enabling/disabling reordering facets. 
Also updated documentation and added a button to test enabling/disabling reordering facets.

PR #720, opened 2018-06-13
EL-3134 - Angular Upgrade Guide - ng-packagr upgrade

ashley-hunter:
As part of some testing I did with Angular 6 for the upgrade guide, I discovered that our bundle was including RxJS, which is problematic as Angular 6 requires RxJS but this is forcing our components to use RxJS 5. This turns out to be a bug in the version of ng-packagr we are using.

I have updated ng-packagr and its config. The new version also performs stricter type checking on templates which revealed a lot of issues such as template referencing private variables and functions and calling functions with the incorrect number of arguments etc. Those have all been fixed. 

I also found many components with unused imports and variables so I have removed them, and also some where we were not unsubscribing from RxJS which have now also been fixed.

These changes also reduce the bundle size by 200kb.


ashley-hunter:
Ah, good find! Thanks, missed updating that file path.

ashley-hunter:
Copy path should now be updated

PR #721, opened 2018-06-15
EL-3066 - WCAG Select

donna-marijne:
Added IDs, roles, and ARIA attributes to describe ux-select as a combobox. This also affected ux-tag-input (Input Controls > Tags).

https://jira.autonomy.com/browse/EL-3066


donna-marijne:
Thanks for the review - changes have been made.

PR #733, opened 2018-06-29
Fixes for ux-typeahead issues

donna-marijne:
* Added a mapping to TypeaheadVisibleOption when using a callback function.
* Added a highlightedKey property to avoid errors in the view.
* Removed faulty code in getDisplayHtml.
* Fixed the "Expression has changed after it was checked" error.

https://plnkr.co/edit/FzoDcXDpjJlzjblP8mWh?p=preview

https://jira.autonomy.com/browse/EL-3177


src/components/typeahead/typeahead.component.ts
```diff
@@ -88,7 +89,24 @@ export class TypeaheadComponent implements AfterViewInit, OnChanges, OnDestroy {
 
         this.loadOptionsCallback = (pageNum: number, pageSize: number, filter: any) => {
```
ashley-hunter: In the docs site, if i go to the Select component and enable paging I get console errors:

```
TypeError: Cannot read property 'key' of null
```

ashley-hunter: Also, in tag inputs, if I `Enable Typeahead` and then click on 'selectFirst' I also get this error.

src/components/typeahead/typeahead.component.ts
```diff
@@ -202,12 +221,7 @@ export class TypeaheadComponent implements AfterViewInit, OnChanges, OnDestroy {
      * @param option
      */
     getDisplayHtml(option: any) {
-        let displayText;
-        if (typeof option === 'string') {
-            displayText = this.getDisplay(option).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
-        } else {
-            displayText = this.getDisplay(option.name).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
-        }
+        const displayText = this.getDisplay(option).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
```
ashley-hunter: When running the select protractor tests I'm seeing an error here:

```
"TypeError: Cannot read property 'name' of undefined
    at TypeaheadComponent.getDisplayHtml (http://localhost:8080/3.js:28615:380)
```

donna-marijne:
Hi sorry I couldn't reproduce any of the above issues, could you check that it built correctly and if so let me know which combination of options causes the error. Thanks!

PR #735, opened 2018-07-03
EL-3153 - Conduit

ashley-hunter:
- Created Conduit Implementation
- Added 2 documentation sections - one giving a basic example and a second showing a more complex example with the ability to change conduit properties dynamically

Will add tests as a separate PR


ashley-hunter:
Made a few additional changes. I noticed our docs failed when building in production mode. Turns out there is an issue with the Angular compiler when using custom decorators (https://github.com/angular/angular-cli/issues/8434) in some versions of Angular v5 when building in production mode.

There is a simple workaround using `forwardRef` so I have update our docs to use forwardRef to allow it to build and mentioned this workaround in our docs

ashley-hunter:
Thanks for the feedback. I have fixed the snippet overflow and moved the functions to the zone service so they can be used by injecting it if a base class cannot be used.

ashley-hunter:
Thanks! No problem, I have updated the documentation to include an explanation about what to do if you can't extend the `ConduitZoneComponent`. No code changes required.

PR #737, opened 2018-07-03
Add openChange event to floating action button

dachristensen:
I have a component that consumes the FloatingActionButtonsComponent and needs to track the open state. There was no way that I could find to do this with the current API. So, I added a get property to return the open$ observable from the FloatingActionButtonsService.

### DIFF
--- a/docs/app/pages/components/components-sections/buttons/floating-action-button/floating-action-button.component.html
+++ b/docs/app/pages/components/components-sections/buttons/floating-action-button/floating-action-button.component.html
@@ -49,6 +49,12 @@
     </tr>
 </uxd-api-properties>
 
+<uxd-api-properties tableTitle="Outputs">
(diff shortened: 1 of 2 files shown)


ashley-hunter:
Hi Dan,

Thanks for your contribution!

Looking at how we usually expose things like this, we often have an event emit the value, something like this:

```typescript
@Output() openChange = new EventEmitter<boolean>();
...

constructor() {
    this._subscription = this.fab.open$.subscribe(value => this.openChange.emit(value));
}
```

Do you require access to observable specifically or would an event work for your use case?

Thanks!

dachristensen:
Ashley, yes, that makes more sense to use an event rather than expose the observable. I updated my branch with the mechanism you propose. I also took a stab at updating the documentation.

PR #745, opened 2018-07-06
EL-3123 - Dashboard Drag Handle

ashley-hunter:
Adding disabled property to dashboard drag handle


ashley-hunter:
It is designed to prevent movement of the panel.

The resizing is not part of this directive, but is already currently possible by adding a `[resizable]="false"` property to a widget.

donna-marijne:
The reason I ask is because you can effectively move it quite freely if any of the other panels are not disabled. Is that intentional?

If the property only affects dragging then maybe [draggable]="false" might be a better name.

ashley-hunter:
I have renamed the attribute to `dragging` instead, which I agree makes more sense, especially as this is being added to the drag directive and not limited to just the dashboard. 

I believe they have an "edit" mode where you can resize and move panels. As there is currently a resizable property used to enable/disabled resizing I believe they will be binding all handles to the same 'editMode' variable.

PR #746, opened 2018-07-06
Ensure the selectChildren option applies to rows that are not visible

donna-marijne:
* For pre-determined data, traverse the source data tree and update the items in the multipleSelectProvider directly.
* For lazy loaded data, update the items in the multipleSelectProvider as soon as they are loaded (after expanding parent).

https://jira.autonomy.com/browse/EL-3181


donna-marijne:
Thank you for catching this issue! There was a problem with the two watches bouncing off each other. There was also a problem with the change intended to support child selection is lazy loaded rows, so this has been removed.

PR #749, opened 2018-07-10
EL-3175 - Column Sorting

ashley-hunter:
JIRA: https://jira.autonomy.com/browse/EL-3175

Updated the Column Sorting component and example to have keyboard accessibility and screen reader support.


ashley-hunter:
Thanks! Missed passing a param into the `getColumnAriaLabel` function.

PR #751, opened 2018-07-10
EL-3051 - Flippable Cards WCAG

ashley-hunter:
JIRA: https://jira.autonomy.com/browse/EL-3051

- Reworked Example Code - the layout spilled outside it container a lot causing focus ring to have weird bits sticking out. Now reworked to be simpler and container within the flippable card (with the exception of the icon at the top). Also improved accessibility.
- Made cards focusable and added support for enter/spacebar
- Updated code snippets
- Nested donut chart - IE by default focuses SVGs on tabbing whereas other browsers do not. Added `focusable="false"` attribute to have consistent tabbing behavior


ashley-hunter:
I wasn't able to reproduce the issue you were getting, but I have added in an assertive announcement when the card flips so hopefully it will stop reading the card content when flipped.

There currently isn't a plunker example because there is a nested donut chart - which is an upgraded component which wont work in plunker as it isn't a hybrid app, however I can remove the nested donut chart from the plunker example if you want.

PR #753, opened 2018-07-13
EL-3193 - Single Line Overflow Tooltip Performance

ashley-hunter:
Jane raised a ticket on single line overflow performance (https://jira.autonomy.com/browse/EL-3193). Roland happy to include in current sprint.

- Removed use of $timeout as this was running a $rootScope.$digest after each tooltip was applied.
- We were initially determining if all tooltips needed to be shown when the directive initialised and we also were re-checking any time the content changed or the element resized which was unnecessary. Instead we now only check when the tooltip would normally be shown, so when the mouse enters the element we do the check.


ashley-hunter:
Jasmine test should now be fixed

PR #758, opened 2018-07-17
EL-3170 - Tree Grid Hybrid Fix

ashley-hunter:
I have removed the documentation example wrappers for the tree grid and replaced them with the hybrid component so they we can see in the docs whether the hybrid component works as expected.

Made the required changes to have the tree grid functioning as expected when used in a hybrid application


ashley-hunter:
Well spotted! I have made that change now. Thanks

PR #759, opened 2018-07-17
EL-3176 - WCAG Table Filters

ashley-hunter:
Improving accessibility of Table Filters


ashley-hunter:
Is this in the Custom Filter section? If so I can change it to be a normal dropdown, as this is only an example of how to make a custom filter

PR #761, opened 2018-07-18
EL-3049 - Fixing WCAG Facet Issues

ashley-hunter:
Fixing QA issues


src/components/facets/facet-check-list/facet-check-list.component.ts
```diff
@@ -30,6 +30,12 @@ export class FacetCheckListComponent extends FacetBaseComponent implements After
         this._focusKeyManager.change.pipe(takeUntil(this._onDestroy)).subscribe(index => this.activeIndex = index);
     }
 
+    onFocus(index: number): void {
```
donna-marijne: Thanks, this works well, could you add it to the typeahead list as well?

ashley-hunter:
That should be changed in the typeahead now too. Thanks!

PR #779, opened 2018-08-01
EL-3057 - Media Player QA Fixes

ashley-hunter:
- Making controls remain visible until focus leaves the controls
- Removing faded tick on focus to meet wcag contrast requirements
- Positioning subtitles higher to avoid being behind the controls


ashley-hunter:
Buttons will now blur when clicked

PR #787, opened 2018-08-02
EL-3206 - Select Table

ashley-hunter:
- Created a `ux-select-list` component (renamed it from select-table as it isn't really a table and I thought it best to avoid confusion as we have a selection section under the tables category already.
- Added documentation section
- Added plunker
- Added protractor tests


docs/app/data/components-page.json
```diff
@@ -1297,13 +1297,33 @@
                     "title": "Single Select Table",
                     "component": "ComponentsSingleSelectTableNg1Component",
                     "version": "AngularJS",
-                    "hybrid": true
+                    "hybrid": true,
+                    "deprecated": true
```
donna-marijne: Could you use deprecatedFor  to provide a link to the new section?

PR #789, opened 2018-08-02
EL-3059 - WCAG Notifications

ashley-hunter:
Improving the notifications example to be more accessible


ashley-hunter:
Added in buffer to announce multiple notifications

PR #790, opened 2018-08-07
EL-3047 - WCAG Drag and Drop

ashley-hunter:
Making draggable cards and drag and drop card sections more accessible.


ashley-hunter:
Issues should now be resolved

PR #791, opened 2018-08-08
EL-3208 - New unlock icon

ashley-hunter:


ashley-hunter:
Icon should now appear correctly

PR #796, opened 2018-08-15
EL-3047 -Drag and Drop Card Outline

ashley-hunter:
Improving outline visibility


docs/app/pages/components/components-sections/drag-and-drop/drag-and-drop-cards/drag-and-drop-cards.component.less
```diff
@@ -29,6 +29,13 @@
     box-shadow: 1px 1px 3px #ccc;
     border: 1px solid #ccc;
     padding: 10px 30px;
+
+    &:focus {
+        outline: 2px dotted;
```
donna-marijne: Would it be possible to use `.aspects-outline` from mixins.less here instead?

ashley-hunter:
Review comment addressed

PR #800, opened 2018-08-17
EL-3213 - Select Table Shift Click

ashley-hunter:
- Reworked Select List to use uxSelection behind the scenes (with a new custom selection strategy)
- Refactored uxSelection to made it slightly more generic
- Refactored Select List AngularJS component:
-- Greatly simplified logic
-- Added shift select
-- Added JSDoc annotations for TypeScript autocompletion 
-- Improved example code
- Updated webpack dev config not to use minified versions of libraries for easier debugging

Note, you may notice shift selection in AngularJS example can be slow sometimes, this is due to a performance issue in the AngularJS select component section above - I will raise a ticket on it.


ashley-hunter:
Focus outlines should now be fixed, thanks

ashley-hunter:
Well spotted, forgot about that. Should be fixed now! thanks

PR #815, opened 2018-09-10
EL-3217 - Table Column Resize

ashley-hunter:
- Adding directive to allow table columns to be resized


ashley-hunter:
Made fixes - will add to MF docs site

PR #817, opened 2018-09-11
EL-3067 - WCAG navigation

donna-marijne:
* Updated ux-navigation to allow an `items` property to be specified as an alternative to the existing functionality. Navigation items can thereby be populated using the NavigationItem type.
* Modified uxTabbableList and uxTabbableListItem to support a hierarchy of items. Hopefully this should also be applicable to dropdowns and be able to replace the other directive that I made previously for the masthead.
* Fixed a "changed after checked" error in uxSelection.
* Corrected a documentation styling issue affecting the width of the "Angular Compatible" popover.

This component is not documented in the UX Aspects site. See accompanying pull request.

https://jira.autonomy.com/browse/EL-3067

### DIFF
--- a/src/components/navigation/navigation-link/navigation-link.directive.ts
+++ b/src/components/navigation/navigation-link/navigation-link.directive.ts
@@ -51,11 +51,13 @@ export class NavigationLinkDirective implements OnInit, OnChanges, OnDestroy {
 
     ngOnInit(): void {
 
-        this._expanded$.pipe(delay(0), takeUntil(this._onDestroy)).subscribe(expanded => {
-            if (this.navigationItem.children && this.navigationItem.children.length > 0) {
-                this.ariaExpanded = `${expanded}`;
-                this._navigationService.setExpanded(this.navigationItem, expanded);
-            }
+        this._expanded$.pipe(takeUntil(this._onDestroy)).subscribe(expanded => {
+            requestAnimationFrame(() => {
+                if (this.navigationItem.children && this.navigationItem.children.length > 0) {
+                    this.ariaExpanded = `${expanded}`;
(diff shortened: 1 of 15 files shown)


donna-marijne:
Sorry - removed those now.

src/components/navigation/navigation-link/navigation-link.directive.ts
```diff
@@ -0,0 +1,139 @@
...
+    constructor(
+        private _router: Router,
+        private _locationStrategy: LocationStrategy,
+        private _navigationService: NavigationService
+    ) { }
+
+    ngOnInit(): void {
+
+        this._expanded$.pipe(delay(0), takeUntil(this._onDestroy)).subscribe(expanded => {
+            if (this.navigationItem.children && this.navigationItem.children.length > 0) {
```
ashley-hunter: This may not be an issue, but I notice when autocollapsing a different section there is a bit of jumpiness happening.

https://i.gyazo.com/405042b0109659b55083a03b09e302ea.mp4

This seems to be because there is a slight delay - I assume this is to prevent an expression changed after check error. I notice that using `requestAnimationFrame` for the delay seems to resolve this, eg:
```typescript
this._expanded$.pipe(takeUntil(this._onDestroy)).subscribe(expanded => {
            requestAnimationFrame(() => {
                if (this.navigationItem.children && this.navigationItem.children.length > 0) {
                    this.ariaExpanded = `${expanded}`;
                    this._navigationService.setExpanded(this.navigationItem, expanded);
                }
            });
        });
`

This may be fine as is, but just thought i'd mention it.

donna-marijne:
Thank you for the suggestion, this does indeed work better!

PR #827, opened 2018-09-17
EL-3220 - Make dynamic select component much more performant

ashley-hunter:
Previously, when the select component loaded it was constantly checking if the infinite scroll needed to load another page, even when the dropdown was not visible causing the function to be called over and over again.

Reworked component to be simpler, more performant and added some additional cleanup logic to correctly destroy watchers


ashley-hunter:
Thanks for the feedback! Those issues should now be resolved

PR #831, opened 2018-09-20
Fixing Angular 5 Issue

ashley-hunter:
Angular 5 doesn't support the generic type on ElementRef causing build errors. Fixing the issue


donna-marijne:
Thanks, did someone raise this as an issue, or did you notice it yourself?

ashley-hunter:
Alex was seeing it and posted in Slack, but he actually intended to install an older version of UX Aspects to match their other project anyway, so it isn't a blocking issue for him.

donna-marijne:
Thanks for taking care of this. Could you make a ticket in the current sprint? I'll add it to the release notes in case someone else bumped into this.

PR #834, opened 2018-09-24
EL-3238 - Splitter Accessibility

ashley-hunter:
Adding splitter accessibility.

- Implements Splitter Accessibility directive
- Updated documentation
- Added tests to cover accessibility functionality


ashley-hunter:
Fixed the outline colour (also in Quantum https://github.houston.softwaregrp.net/caf/quantum/pull/545).

I wasn't able to see the issue you were seeing in Firefox and IE, it appears to be working ok for me? Check to make sure they are showing the Angular sections and not AngularJS as they look virtually identical and it caught me out at first.

PR #836, opened 2018-09-26
EL-3231 - Tree grid

donna-marijne:
https://autjira.microfocus.com/browse/EL-3231


donna-marijne:
Thanks, yes that is weird! I'll see if I can figure out what's going on.

donna-marijne:
I could not figure out a solution here. box-shadow does not work either. So I have used the same background color that the AngularJS tree grid uses, in addition to the outline.

PR #841, opened 2018-09-28
EL-3235 - Color Picker Focus

ashley-hunter:
- Focus is now only returned when escape is pressed


ashley-hunter:
Focus should now return on selection

PR #845, opened 2018-10-03
EL-3242 - Table Column Resize

ashley-hunter:
- Making the column resizing work with uxFixedHeaderTable.
- Refactored existing code to make it simpler
- An additional directive is required when using is with a fixed header table - added documentation to include this.
- Fixed an issue with the custom responsive table example which produced an error in the console.


ashley-hunter:
I have added an example using the fixed header table and the issue with the excess white space should now also be fixed!

PR #847, opened 2018-10-04
EL-3245 - Tidy Node Scripts

ashley-hunter:
- Went through our node scripts and renamed some of them. We had a wide variety of naming conventions, now they follow {target}-{action} naming convention, eg: `library-build`, `library-package`.

- Removed any tasks that I didn't think were being used (correct me if I have removed anything that is still required)

- Updated grunt tasks to point to the updated npm scripts

- Added npm scripts for `grunt build` and `grunt build:library`


package.json
```diff
@@ -15,14 +15,13 @@
...
-    "package:library": "grunt build:library",
-    "lint": "grunt lint",
-    "e2e:develop": "concurrently \"webpack --watch --progress --colors --config configs/webpack.e2e.config.js\" \"tsc -w --project .\\e2e\\tsconfig.json\" \"npm run serve:e2e\""
+    "start-https": "npm run start -- --https --pfx configs/webpack.docs.dev.pfx",
+    "e2e": "concurrently \"npm run e2e-serve\" \"grunt protractor:e2e\" --kill-others --success first",
+    "e2e-serve": "lite-server -c=e2e/bs-config.e2e.json",
+    "e2e-develop": "concurrently \"webpack --watch --progress --colors --config configs/webpack.e2e.config.js\" \"tsc -w --project .\\e2e\\tsconfig.json\" \"npm run e2e-serve\"",
+    "library-build": "ng-packagr -p ./src/ng-package.json",
+    "grunt-build": "grunt build",
+    "grunt-build-library": "grunt build:library"
```
donna-marijne: Without the former build:library task, these can just be "build" and "build:library".

package.json
```diff
@@ -15,14 +15,13 @@
...
-    "start:https": "webpack-dev-server --config configs/webpack.dev.config.js --https --pfx configs/webpack.docs.dev.pfx",
-    "serve:e2e": "lite-server -c=e2e/bs-config.e2e.json",
-    "e2e": "concurrently \"npm run serve:e2e\" \"grunt protractor:e2e\" --kill-others --success first",
-    "build:library": "ng-packagr -p ./src/ng-package.json",
-    "build:documentation:prod": "node --max-old-space-size=4096 ./node_modules/webpack/bin/webpack.js --colors --config ./configs/webpack.docs.prod.config.js",
-    "package:library": "grunt build:library",
-    "lint": "grunt lint",
-    "e2e:develop": "concurrently \"webpack --watch --progress --colors --config configs/webpack.e2e.config.js\" \"tsc -w --project .\\e2e\\tsconfig.json\" \"npm run serve:e2e\""
+    "start-https": "npm run start -- --https --pfx configs/webpack.docs.dev.pfx",
+    "e2e": "concurrently \"npm run e2e-serve\" \"grunt protractor:e2e\" --kill-others --success first",
```
donna-marijne: As I recall, `npm run e2e` does not work for running the e2e tests. So this script should just call `grunt e2e`. Something like https://github.com/sindresorhus/grunt-concurrent might help to move this out of package.json and also remove the need for this global tool.

Comparing with the package.json for Material, perhaps "test:e2e" would be a better name here. Then we can have something like `"test": "grunt test && grunt e2e"` for a full test run (which would be the nightly build).

ashley-hunter:
Made Changes:

- Remove lite-server. We needed it because there was no easy way to know when webpack-dev-server was finished building and serving the app for us to start running the tests, but I wrote a small script which handles this and allows us to remove this dependency and concurrently as well.

- Refactored our e2e setup. It is now more aligned with how the Angular CLI does it, by using ts-node allowing us to remove the typescript build step and allowing us to remove the grunt-typescript dependency as well. I also removed a lot of commented out code from our protractor config.

- Remove the `configuration` folder as this contained scripts for the selenium test grid we used to use ages ago. As we are now using protractor this is not needed.

- Removed the grunt `server` task, as this was used by selenium, not required now we use protractor and webpack-dev-server. Also removed the `grunt-contrib-connect` dependency as it is no longer required.

- Reverted script names/changed as per review comments

I don't we will have any issues with global node tasks as npm scripts use webpack, grunt etc, from the node_modules folder if they are installed as dependencies.

I will update the MF PR after you have reviewed this.

PR #849, opened 2018-10-09
EL-3243 - Docs tgz

ashley-hunter:
- Added an additional compress task to produce a tgz matching the contents of the ux-aspects-docs npm package
- Removed dockerfile that was used for old selenium test grid
- Removed documentation:serve grunt task as this was used before we had `npm start` and is no longer required.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -32,14 +32,14 @@ module.exports = function (grunt) {
     grunt.registerTask('webpack_import_cert', ['run:webpack_import_cert']);
     grunt.registerTask('package:library', ['run:npm_pack', 'copy:npm_tgz', 'clean:npm_tgz']);
     grunt.registerTask('package:library_bower', ['compress:bower']);
-    grunt.registerTask('package:docs', ['compress:documentation', 'run:npm_pack_docs', 'copy:npm_docs_tgz', 'clean:npm_docs_tgz']);
+    grunt.registerTask('package:docs', ['run:npm_pack_docs', 'copy:npm_docs_tgz', 'clean:npm_docs_tgz']);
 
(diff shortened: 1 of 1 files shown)


ashley-hunter:
I will add it to MF repo once reviewed

grunt/compress.js
```diff
@@ -19,5 +19,26 @@ module.exports = {
...
+    },
+    documentation: {
+        options: {
+            archive: path.join(process.cwd(), 'target', 'docs', 'ux-aspects-docs-<%= package.version %>.tgz'),
+            mode: 'tgz'
+        },
+        files: [
+            {
+                cwd: path.join(process.cwd()),
+                src: [
```
donna-marijne: Shouldn't this be `src: 'dist/docs'`?

ashley-hunter: So just to clarify, is this supposed to contain the site that will be published to gh-pages? or contain the contents of the ux-aspects-docs package? I wasn't sure which it was supposed to be

donna-marijne: Yes, it's the documentation site from dist/docs.

ashley-hunter: Great thanks! The changes should now be made!

donna-marijne:
Just one other thing I noticed, this should not be run as part of `build:library`. Now that we are packaging the actual documentation as well as the documentation resources, the task `package:docs` has a bit of a misleading name.

ashley-hunter:
No problem, do you want this part of any task in particular or will it just be called by running `grunt compress:documentation`.

donna-marijne:
Let's just chain it onto the end of the current 'build' task for now. We can see if the build process needs it to be a separate step once we start integrating.

ashley-hunter:
Great! I have added it to the end of the build task

PR #850, opened 2018-10-09
EL-3244 - Environment Variable

ashley-hunter:
- Created a script to update the versions in all package.json and bower.json files.
- Wasn't able to find a way to pass an environment variable to `npm version` (at least not on Windows)
- Added grunt task to call the script - `grunt execute:version`.
- Checked with Dermot and Tony about the name of the environment variable. We will have full control over the name of it in the pom file whenever those changes are merged in.


ashley-hunter:
I have added an npm script to set the version. Thats strange it didn't work for you, I tried in bash using:

```bash
VERSION=1.2.3 node ./scripts/version.js
```

That seemed to work for me. Importing `env` just prevents linting complaining about `process` not being defined, but is the same as doing `process.env`.

donna-marijne:
Ah, okay yes that works. I'm not sure how it would be invoked by the build environment but we'll be able to test that soon. Thanks.

PR #854, opened 2018-10-12
EL-3254 - Table Filters

ashley-hunter:
- Added filters input/output on the filter container components (while these were there before they didn't work and weren't documented). This can now be used to preselect filters
- Refactored the filters to use a service rather than extending a base component. This way anytime we need to inject an additional dependency it is not a breaking change if the consuming app has custom filters. Retained support for the old base component but marked as deprecated and updated documentation to reflect new way.
- Split out event classes and interfaces into separate files to prevent circular dependencies
- Added new TSlint rule to ban importing directly from "rxjs". This is the import suggestion vs code suggests, however importing from here would break Angular 5 support, so now linting will fail if this goes unnoticed,
- Moved live announcer to the example code, rather than within the filter components to allow more customisation and localisation.


docs/app/pages/components/components-sections/tables/filters/filters.component.ts
```diff
@@ -158,19 +164,28 @@ export class ComponentsFiltersComponent extends BaseDocumentationSection impleme
         // apply a newly added filter
         if (event instanceof FilterAddEvent) {
             this.applyFilter(event.filter);
+
+            // announce the selection
+            this._announcer.announce(`Filter ${event.filter.name} selected.`);
```
donna-marijne: Probably should rename the function containing this, since we now have an event emitter with almost the same name. My suggestion would be to use the `(events)` emitter for LiveAnnouncer and `(filterChange)` to update the table's filtered content. Then if one modifies the `filter` array directly, the table state would still be updated.

docs/app/pages/components/components-sections/tables/filters/filters.component.ts
```diff
@@ -152,65 +151,51 @@ export class ComponentsFiltersComponent extends BaseDocumentationSection impleme
...
-        }
+                case 'status':
+                    this.documents = this.documents.filter(item => item.active === (filter.name === 'Active'));
+                    break;
+            }
+        });
+    }
 
-        this.activeFilters.push(filter);
+    applyFilter(filter: Filter): void {
```
donna-marijne: I didn't notice where this function is used - could it be removed?

ashley-hunter:
That should be it removed now! Thanks

PR #862, opened 2018-10-17
EL-3069 - WCAG Tables

ashley-hunter:
Ticket Comments Below:

Deprecated remaining AngularJS table sections as they are examples only or things we are not looking to migrate.

Column Sorting - **Previously Done**

Custom Filters - Example Updated - Now uses a list and menu navigation directives

Filters - **Done**

Fixed Header Table:
- Possibly aria-setsize and aria-posinset. **Done**
- Maybe aria-live for start and end of async loading? Compare with infinite scroll/paging select component. **The other components didn't do this, however I can add it in if you think it should be there**

Hover Actions:
- Update example with aria-labels on the action buttons. **Done**

Selection:
- Update the directive to add aria-selected as appropriate. **Done**
- Checkbox should become aria-hidden. **Done**

Toolbar Sorting:
- Add keyboard control and focus management to the dropdown. Compare with masthead dropdowns or use uxTabbableList. **Done**
- Sort direction button requires an aria-label. **Done**

Column Resizing:
- Allow focus on the resize thumb. Keyboard control similar to the splitter. **Done**
- Make the size thumb visible so that the user knows that it is focusable. **Done**
- Aria label/description to indicate functionality. **Done**

### DIFF
--- a/docs/app/pages/components/components-sections/tables/custom-filters-component/sample/sample-filter.component.html
+++ b/docs/app/pages/components/components-sections/tables/custom-filters-component/sample/sample-filter.component.html
@@ -1,12 +1,44 @@
-<div class="btn-group ux-filter-custom" dropdown>
-    <button dropdownToggle type="button" class="filter-dropdown btn dropdown-toggle"
-    [class.active]="selected !== initial">{{ selected?.title }}
+<div class="btn-group ux-custom-filter"
+    dropdown
+    #dropdown="bs-dropdown"
+    [autoClose]="true">
+
+    <button type="button"
+        tabindex="0"
+        dropdownToggle
+        uxMenuNavigationToggle
+        #toggle="uxMenuNavigationToggle"
+        [(menuOpen)]="dropdown.isOpen"
+        aria-haspopup="true"
+        [attr.aria-expanded]="dropdown.isOpen"
+        class="filter-dropdown btn dropdown-toggle"
+        [class.active]="selected !== initial">
+        {{ selected?.title }}
         <span class="hpe-icon hpe-down"></span>
     </button>
-    <ul *dropdownMenu class="dropdown-menu" role="menu">
-        <li class="dropdown-list-item" *ngFor="let filter of filters" role="menuitem">
-            <ux-radio-button (click)="selectFilter(filter)" name="filterGroup" [(value)]="selected"
-            [option]="filter">{{ filter.name }}</ux-radio-button>
+
+    <ul *dropdownMenu
+        uxMenuNavigation
+        [toggleButton]="toggle"
+        class="dropdown-menu"
+        role="menu">
+
+        <li class="dropdown-list-item"
+            *ngFor="let filter of filters"
+            role="none">
+
+            <a class="dropdown-item"
+                role="listitem"
+                tabindex="-1"
+                uxMenuNavigationItem
+                [attr.aria-selected]="filter === selected"
+                (click)="selectFilter(filter, $event); toggle.focus()"
+                (keydown.enter)="selectFilter(filter, $event); toggle.focus()"
+                (keydown.escape)="toggle.focus()">
+
+                <i class="hpe-icon" [class.hpe-checkmark]="filter === selected"></i>
+                <span class="filter-dropdown-title">{{ filter.name }}</span>
+            </a>
         </li>
     </ul>
 </div>
--- a/docs/app/pages/components/components-sections/tables/custom-filters-component/sample/sample-filter.component.html
+++ b/docs/app/pages/components/components-sections/tables/custom-filters-component/sample/sample-filter.component.html
@@ -28,7 +28,7 @@
             role="none">
 
             <a class="dropdown-item"
-                role="listitem"
+                role="menuitem"
                 tabindex="-1"
                 uxMenuNavigationItem
                 [attr.aria-selected]="filter === selected"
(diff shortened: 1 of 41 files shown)


docs/app/pages/components/components-sections/tables/custom-filters-component/sample/sample-filter.component.html
```diff
@@ -1,12 +1,44 @@
...
+        [toggleButton]="toggle"
+        class="dropdown-menu"
+        role="menu">
+
+        <li class="dropdown-list-item"
+            *ngFor="let filter of filters"
+            role="none">
+
+            <a class="dropdown-item"
+                role="listitem"
```
donna-marijne: Should be `role="menuitem"`.

docs/app/pages/components/components-sections/tables/sorting/sorting.component.html
```diff
@@ -1,17 +1,30 @@
...
+            uxMenuNavigationToggle
+            [(menuOpen)]="dropdown.isOpen"
+            dropdownToggle
+            class="btn dropdown-toggle filter-dropdown">
             {{ selected }} <span class="hpe-icon hpe-down"></span>
         </button>
 
         <!-- Dropdown List -->
         <ul *dropdownMenu class="dropdown-menu" role="menu">
             <li role="menuitem" *ngFor="let option of options">
```
donna-marijne: Also here I think `role="menuitem"` needs to be on the focusable element which presumably is the `a`.

src/components/table/table-column-resize/resizable-table-column.component.less
```diff
@@ -15,13 +15,26 @@
...
 
+    &.cdk-keyboard-focused {
+        .aspects-outline();
+    }
+
+    .ux-resizable-table-column-handle-icon {
+        width: 5px;
+        height: 24px;
+        background-repeat: no-repeat;
+        background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAUAAAAeCAYAAADkftS9AAAAIklEQVQoU2M4c+bMfxAGAgYYmwGrIIiDjrELjpo5aiZeMwF+yNnOs5KSvgAAAABJRU5ErkJggg==);
```
donna-marijne: I think this is better as a separate file, or potentially an addition to our icon set. We are still delivering our CSS as one combined file, so keeping it small should help load times.

docs/app/pages/components/components-sections/tables/sorting/snippets/app.html
```diff
@@ -21,10 +34,12 @@
         </ul>
 
         <!-- Sort Direction Toggle -->
-        <span class="sort-icon hpe-icon text-muted"
+        <button class="sort-icon hpe-icon text-muted"
+            aria-label="Toggle sort direction"
+            [attr.aria-sort]="descending ? 'descending' : 'ascending'"
```
donna-marijne: aria-sort only applies to table/grid header elements. The state of the button could be communicated via the aria-label.

ashley-hunter:
That should be the points mentioned above corrected, with one exception. The drag handle image cannot reference a file. This is currently an issue with how stuff is currently bundled. For our documentation to build correctly the image would have to point to a relative path, because we are using webpack, however when it comes to building the library we build the less using the less compiler which doesn't rewrite the paths, so the image file would be missing. I've had this issue before and never been able to find an easy fix for this.

ashley-hunter:
That should be the points mentioned above corrected, with one exception. The drag handle image cannot reference a file. This is currently an issue with how stuff is currently bundled. For our documentation to build correctly the image would have to point to a relative path, because we are using webpack, however when it comes to building the library we build the less using the less compiler which doesn't rewrite the paths, so the image file would be missing. I've had this issue before and never been able to find an easy fix for this.

docs/app/pages/components/components-sections/tables/hover-actions/hover-actions.component.html
```diff
@@ -18,9 +18,15 @@
                 <ux-spark theme="chart2" [value]="document.complete" [inlineLabel]="document.complete" barHeight="3"></ux-spark>
             </td>
             <td>
-                <i class="hpe-icon hpe-view hover-action-icon" aria-label="View" uxTooltip="View" uxHoverAction></i>
-                <i class="hpe-icon hpe-share hover-action-icon" aria-label="Share" uxTooltip="Share" uxHoverAction></i>
-                <i class="hpe-icon hpe-trash hover-action-icon" aria-label="Delete" uxTooltip="Delete" uxHoverAction></i>
+                <button class="hover-action-btn" aria-label="View" uxTooltip="View" uxHoverAction>
```
donna-marijne: The default `type` for a button is submit, so on the offchance that this appears in a form it would be better to specify `type="button"`.

docs/app/pages/components/components-sections/tables/hover-actions/hover-actions.component.less
```diff
@@ -6,6 +6,19 @@
         background-color: #f5f5f5;
     }
 
+    .hover-action-btn {
```
donna-marijne: Can this be part of the library stylesheet?

docs/app/pages/components/components-sections/tables/reorderable-table/reorderable-table.component.less
```diff
@@ -2,6 +2,13 @@
     background-color: #fff;
 }
 
+.reorderable-row:focus {
```
donna-marijne: Can this be part of our stylesheet too? We already have styles for rows which can receive focus so we should be able to reuse those.

PR #869, opened 2018-10-22
EL-3042 - Buttons WCAG

ashley-hunter:
- New Pagination Control that is now accessible
- Update FAB component to be keyboard accessible

### DIFF
--- a/docs/app/pages/components/components-sections/buttons/pagination/pagination.component.html
+++ b/docs/app/pages/components/components-sections/buttons/pagination/pagination.component.html
@@ -21,7 +21,7 @@
 
 <ul>
     <li>To customize the content of the previous button add <code>#previousBtnTemplate</code> to the <code>ng-template</code> element.</li>
-    <li>To customize the content of the previous next add <code>#nextBtnTemplate</code> to the <code>ng-template</code> element.</li>
+    <li>To customize the content of the next button add <code>#nextBtnTemplate</code> to the <code>ng-template</code> element.</li>
 </ul>
 
 <p>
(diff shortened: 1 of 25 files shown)


ashley-hunter:
Thanks for the feedback. Made the changes requested. In regards to the focus style being moved out to a general rule, I did this however because many bootstrap styles have more specific selector, the focus styling had to be made !important, this also had to be done in Quantum also.

docs/app/pages/components/components-sections/buttons/pagination/pagination.component.html
```diff
@@ -15,6 +15,15 @@
     large datasets. The pagination component can also be used in forms using <code>ngModel</code> or reactive forms.
 </p>
 
+<p>
+    The direction button content can be customized by adding an <code>ng-template</code> as a child of the <code>ux-pagination</code> component.
+</p>
+
+<ul>
+    <li>To customize the content of the previous button add <code>#previousBtnTemplate</code> to the <code>ng-template</code> element.</li>
+    <li>To customize the content of the previous next add <code>#nextBtnTemplate</code> to the <code>ng-template</code> element.</li>
```
donna-marijne: Typo - "previous next"

src/components/pagination/pagination.component.html
```diff
@@ -1,14 +1,18 @@
-<ul class="pagination" [ngClass]="classes" uxTabbableList direction="horizontal" [allowBoundaryKeys]="true">
+<ul #container class="pagination"
+  [ngClass]="classes"
+  direction="horizontal"
+  (blur)="isKeyboardEvent = false"
+  (keydown.ArrowLeft)="select(page - 1); isKeyboardEvent = true"
+  (keydown.ArrowRight)="select(page + 1); isKeyboardEvent = true">
```
donna-marijne: Could you add in the home/end functionality as well? I think that would be a convenient addition.

ashley-hunter:
Review changes made

donna-marijne:
Referring to the link I posted earlier (http://www.a11ymatters.com/pattern/pagination/) could we wrap it in a `nav` element and pass through an aria-label? Also as noted in the article, `aria-current` is more appropriate than `aria-selected`.

ashley-hunter:
Wrapped in nav element. If they add an aria label to the component it will now get forwarded to the nav element. Also changed to use aria-current instead of aria-selected

PR #878, opened 2018-10-26
EL-3275 - Column Resizing Improvements

ashley-hunter:
Adding width input to allow a custom width to be set. Also added a widthChange emitter to emit whenever dragging occurs. Updated documentation to have an api table with the new properties


ashley-hunter:
Thanks for the feedback!

1) I have fixed the issue where sibling columns did not emit the latest width.
2) I have fixed the issue where column widths are emitted. There would be quite a lot of rework involved to make the last column always variable as it has to take into account the disabled and min width of columns etc, so the last column might not always be possible to scale. I have documented that columns are scaled proportionally to fill any remaining width if column widths dont total the table width, but that if they provide a variable width column then it will be resized. If this is a problem I can raise a ticket to change this. I have fixed the emission issue with this that you mentioned.

3) I have updated the example to show column widths and removed the checkbox column CSS (let me know if you think the widths should be presented any differently)

ashley-hunter:
Review changes made, should now work as expected in IE.

ashley-hunter:
Issue should now be fixed. Thanks!

PR #881, opened 2018-11-01
EL-3232 - Dashboard WCAG

ashley-hunter:
I have made the functionality changes to allow a 'grab' mode which can be used to move and resize an announcing changes etc.. the documentation example has also been updated to demonstrate this feature.

I have not yet added any tests or documentation. I wondered if perhaps you could review the functionality before I write tests or document stuff, in case there are any review changes required and I have to redo them afterwards.

### DIFF
--- a/docs/app/pages/components/components-sections/dashboard/dashboard/dashboard.component.html
+++ b/docs/app/pages/components/components-sections/dashboard/dashboard/dashboard.component.html
@@ -1,7 +1,7 @@
 <ux-dashboard class="customizable-dashboard" [options]="options">
 
     <ux-dashboard-widget id="analytics-1-widget" [colSpan]="4" [rowSpan]="2" [resizable]="true">
-        <div class="widget-content" uxDashboardWidgetDragHandle>
+        <div class="widget-content" uxDashboardWidgetDragHandle uxDashboardGrabHandle>
             <h3 class="widget-title">Usage Analytics</h3>
 
             <div class="widget-chart">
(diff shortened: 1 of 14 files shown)


ashley-hunter:
1. All Widgets Resizable - **Fixed**
2. Column Aware moving - **Pressing the down arrow would move to any widget directly below and up arrow would move to any directly above**
3. This was a tricky one, everything I tried I run into additional edge cases that didn't work, the only thing I could do to make it work was disable all other widgets from shifting until the movement was either confirmed or cancelled. This allows the user to move the widget to any position on the grid and then confirm the new position, after which the other widgets would re-position themselves. This may not be desired, if not I can revert the changes.
4. Grab Handle cursor changed and made it the drag handle
5. I didn't change this to a service to allow localising strings, I have looked into this a bit and if a team is using the built in Angular localization, they cannot get the strings in JS code, they can only apply them in the view through a property. Having them properties allows us to support for Angular's i18n and ngx-translate. However there is future work planned to allow runtime access to the strings so it might be good to review localization throughout our components once this work has been completed.
6. Sort order changed
7. Fixed widget name property

donna-marijne:
Regarding #3, could you set up a demo of what you are suggesting for Roland and myself after tomorrow's standup?

Re #5, no problem, didn't think of that.

ashley-hunter:
Review changes made.
1. Made smaller and increased shadow when in grab mode
2. Fixed
3. Fixed
4. I wasn't able to reproduce this? I was getting screen reader feedback after every move/resize

ashley-hunter:
Changes should now be made, thanks

ashley-hunter:
Added e2e tests to cover new functionality. Note there is an existing e2e test for dashboard that has been failing for a while. I tried to fix it but not sure what the issue is, will probably need to fix it as part of another ticket.

PR #882, opened 2018-11-02
EL-3299 - Tree Grid Selection

ashley-hunter:
- Adding 2 way binding to tree view selected property
- Exposing the multipleSelectProvider instance
- List hover actions - I noticed when hovering on a focused and selected item in the tree grid the list hover actions got the wrong background colour. While looking into this we had styles setting the background colour to match the various states. This was completely unnecessary, removing the background colour all together ensures they will not differ from the table row colour
- Fixing up unit tests (they all should now pass)
- Found two minor issues when fixing unit tests - page header didn't explicitly state its injectable which causes it not to work when minified and select table reselect option didn't work as it was clearing the selection before storing it.


ashley-hunter:
Fixing review comments

PR #885, opened 2018-11-07
EL-3232 - Dashboard Improvements

ashley-hunter:
Moving offset code to the documentation styles and making grab movement and resizing animated


ashley-hunter:
No problem, should I keep the animation as it moves and resizes in Grab mode?

donna-marijne:
I also noticed that the dropshadow and a bit of the outline are being clipped in the current build with the offset:
image

donna-marijne:
Yes, the animations looks good, keeping those would be fine I think.

ashley-hunter:
Issues mentioned above should now be fixed. Will raise a quantum ticket to prevent offset in MF

ashley-hunter:
Quantum PR raised: https://github.houston.softwaregrp.net/caf/quantum/pull/582

PR #890, opened 2018-11-08
EL-3281 - Angular 7 Documentation

ashley-hunter:
- Updated Angular to v7
- Updated TypeScript (also fixed type error in Media Player)
- Removed UglifyWebpackPlugin in favour of new TerserPlugin as used by the CLI
- Updated Babel for our AngularJS code
- Removed grunt-webpack as there were a lot of bugs in newer versions which were needed to update webpack. Moved this to a run task as we had already done this for our docs and e2e webpack builds


ashley-hunter:
Resolving conflicts and review issue. Thanks

PR #892, opened 2018-11-09
EL-3288 - Fixed Header Table Layout Update

ashley-hunter:
- Exposing the setLayout function
- Making the directive accessible via exportAs
- Adding dataset input that can be used to trigger this setLayout automatically when the dataset changes
- Updating documentation
- Updated example code

### DIFF
--- a/docs/app/data/components-page.json
+++ b/docs/app/data/components-page.json
@@ -1568,6 +1568,10 @@
                         {
                             "title": "Module",
                             "content": "FixedHeaderTableModule"
+                        },
+                        {
+                            "title": "Exported As",
+                            "content": "ux-fixed-header-table"
                         }
                     ]
                 },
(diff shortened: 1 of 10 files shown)


donna-marijne:
What is the repro case for the ticket? I created the following plunker according to my interpretation of the ticket but it seems to work fine on the latest release: https://plnkr.co/edit/q7gVyeJUm8YgSDGu4CNN?p=preview

ashley-hunter:
Im not sure if there was ever a plunker created. I believe the consuming application did not want a scrollbar to be visible unless it is required, so overflow need to be set to auto on the table body. Then have a table with fewer items than would cause a scrollbar to show, then add additional items to the table which causes the table to show a scrollbar. When the scrollbar appears the columns in the header did not correctly align with the columns in the table body

ashley-hunter:
Fixed the issue, just needed to delay for the ui to update before calling the setLayout method. I was also getting a weird error in plunker regarding the colorpicker, so I fixed it, not sure what changed to start causing it.

PR #895, opened 2018-11-19
EL-3260 - Optimise Protractor

ashley-hunter:
Made some optimisations - instead of 30-40 mins, now completes in ~11 minutes on my machine

- Made optimisations to webpack config
- Removed AngularJS, and any AngularJS tests and test pages
- E2E application is no longer a hybrid application
- Uses built version of library and aliased to `@ux-aspects/ux-aspects`
- Removed some useless tests
- Tests now run in parallel with up to 5 tests being able to be run simultaneously


ashley-hunter:
Yes I will make a new PR for the MF protractor test changes, just wanted to get these changes reviewed first before I made the changes in the other repo, thanks!

PR #897, opened 2018-11-20
EL-3294 - Facet Selection Improvements

ashley-hunter:
- Deprecating FacetBaseComponent in favour of an injectable service (now the same as filters)
- Updated facet all facets to use new service
- Fixing Facet e2e tests
- Found some uses of the old RXJS prototype patching, so removed it


ashley-hunter:
Found a small thing I need to change in this PR, will comment once it has been made

ashley-hunter:
Changes now made, should be ready for review

PR #900, opened 2018-11-21
EL-3280 - audit breaking

donna-marijne:
* Updated less.
* Updated angular-tree-component and moved it to peer dependencies.

https://autjira.microfocus.com/browse/EL-3280


donna-marijne:
Thanks for catching this, I have corrected the error.

PR #905, opened 2018-11-26
EL-3273 - WCAG Spark Charts

ashley-hunter:
- Most of the Spark chart WCAG work was already completed
- Added support for an array of aria-labels for multi value spark charts
- Added E2E tests
- Also added an .editorconfig file, I'm often having to change editor spacing to use 4 spaces, so this will enforce consistency automatically
- Removing old unused testng.xml files


ashley-hunter:
Added aria labels to multi value spark chart

PR #913, opened 2018-11-29
EL-3316 - AngularJS Fixing Minification Issue

ashley-hunter:
- Webpack `mode` was set to production which minifies the bundle by default
- Now set to development mode and lets the existing grunt task minify the bundle as it used to


donna-marijne:
Reminder to please kick off a build in Jenkins for non trivial changes - thanks.

PR #915, opened 2018-11-29
Fixing build issue

ashley-hunter:
Moving type definition before component class to prevent a breaking optimisation


ashley-hunter:
Hmm, strange, fixed it for me. I have tried another way

ashley-hunter:
Im using 3.1.6 i believe. I'm getting the same result as you now, it was the dynamic import statement in the .d.ts file that was causing the errors so the output you are now getting now seems to work.

PR #917, opened 2018-11-30
EL-3311 - Select List Event Emitter Fix

ashley-hunter:
The setter was always calling deselectAll before storing the new selection whenever any change was made, which resulted in a large amount of events getting emitted.

Now we only deselect and update the selection whenever the selection has actually changed. Also added an additional check to see if the array are different but both have 0 items then also do nothing.

Now instead of like 80 events emitted each click, we get 1 during multiple selection and 2 in single selection (single selection must first deselect the previously selected item then select the new one which results in 2 emissions).


ashley-hunter:
Made changes, there is now only one event emitted on single select

PR #924, opened 2018-12-06
Add localization support for meridians in date-time-picker

taufanardi:
#### Summary
There is an @input for time-picker component to assign meridians (array of string).
This @input though, is not used by date-time-picker component.
For our project in opsbridge-x (mondrian), we need to give localization for this meridians string.
With this pull request we add new @input in date-time-picker and use it when displaying time-picker.

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_


taufanardi:
Hi Donna,

Yes please, where should I add the documentation?
Or maybe you can point out the guide on how to make pull request for this project in general?
For example i left the "ticket" empty since I (and I assume any dev outside your team) not familiar with this process. Thank you.

donna-marijne:
No problem, you can find the documentation pages under `docs/app/pages`; in this case, the following file is what needs to be updated: https://github.com/UXAspects/UXAspects/blob/develop/docs/app/pages/components/components-sections/date-time-picker/date-time-picker/date-time-picker.component.html.

You may have seen the [CONTRIBUTING](https://github.com/UXAspects/UXAspects/blob/develop/CONTRIBUTING.md) file and the linked developer standard. The PR template is something that I have started trialling in the past few days, so we will probably refine that a bit more. You are correct that the additional fields are currently only useful for our dev team.

Thanks again for your contribution!

PR #928, opened 2018-12-12
Spark segment tooltips

donna-marijne:
#### Summary
* Added spark-tooltips since the existing spark-tooltip attribute will not support an array.
* The UpgradeComponent did not bind the new property however, so only the codepen example has a working multi-value tooltip.

#### Ticket
https://autjira.microfocus.com/browse/EL-3041

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3041-spark-ng1-tooltips


ashley-hunter:
Changes look good, I just realised why the binding wasn't working in our documentation, so i pushed the change to make it work. If you want to take a quick look to see if you are happy with the changes then it can be merged

donna-marijne:
Great, thanks for figuring that one out!

PR #937, opened 2019-01-02
EL-3345-Improve-Localization

ashley-hunter:
- Removing all uses of the i18n Angular functionality. Instead all strings that previous used this are now defined using inputs instead.
- Because some aria labels need a function to be called to produce a valuable label, I have had to rework a couple of components a bit to allow them to use OnPush change detection, otherwise the aria label functions were getting called many time every time change detection runs.

#### Ticket
https://autjira.microfocus.com/browse/EL-3345

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3345-Improve-Localization

### DIFF
--- a/docs/app/pages/components/components-sections/color/color-picker/color-picker.component.html
+++ b/docs/app/pages/components/components-sections/color/color-picker/color-picker.component.html
@@ -149,6 +149,15 @@
         The default input mode to display in the input panel.
         The user can switch modes using the toggle button.
     </tr>
+    <tr uxd-api-property name="colorAriaLabel" type="(color: ColorPickerColor) => string">
+        Define a function that returns an aria-label for a <code>ColorPickerColor</code>.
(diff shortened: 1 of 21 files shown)


ashley-hunter:
Just verified there, there shouldn't be any breaking changes, if they have already defined the strings then they will just be ignored by the Angular compiler at build time, its not essential that they remove them. It appears errors would only appear if we add new translations to our templates that they don't have listed but if we remove them from the html everything compiles fine even if they are still listed in the translation file.

donna-marijne:
If someone has already provided strings is there something we can do to help them migrate? I see that we can't get access to i18n strings in code until ivy is released, but if we could wait until then perhaps we could make a shim library which contains the i18n definitions and makes them available via a service.

PR #938, opened 2019-01-04
EL-3336 - Date Time Picker

ashley-hunter:
- Adding support for `startOfWeek` input which allows the user to define which day of the week should be in the first column.
- Added test case to cover this new feature.

#### Ticket
https://autjira.microfocus.com/browse/EL-3336

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3336-Date-Start-Of-Week

### DIFF
--- a/docs/app/pages/components/components-sections/date-time-picker/date-time-picker/date-time-picker.component.html
+++ b/docs/app/pages/components/components-sections/date-time-picker/date-time-picker/date-time-picker.component.html
@@ -114,7 +114,7 @@
         Defines the text to be displayed in the button used to set the selected time to the current time.
     </tr>
     <tr uxd-api-property name="startOfWeek" type="WeekDay" defaultValue="WeekDay.Sunday">
-        Defines the day of the week that should appear in the first column.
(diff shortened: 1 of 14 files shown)


docs/app/pages/components/components-sections/date-time-picker/date-time-picker/date-time-picker.component.html
```diff
@@ -113,6 +113,9 @@
     <tr uxd-api-property name="nowBtnText" type="string">
         Defines the text to be displayed in the button used to set the selected time to the current time.
     </tr>
+    <tr uxd-api-property name="startOfWeek" type="WeekDay" defaultValue="WeekDay.Sunday">
+        Defines the day of the week that should appear in the first column.
```
donna-marijne: Could you note that the `WeekDay` enumeration comes from @angular/common, since the code example doesn't show it?

ashley-hunter: Thanks! Updated the documentation to mention this.

PR #944, opened 2019-01-07
EL-3358 - Number Picker Improvements

ashley-hunter:
Adding `id` and `aria-labelledby` bindings to the number input picker. Also updated setters to use the Angular CDK coercion functions rather than rolling our own for each setter.

#### Ticket
https://autjira.microfocus.com/browse/EL-3358

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3358-Number-Picker-Id


ashley-hunter:
Thanks for the feedback, good point, I have updated to add a different id to the child input and updated the documentation. Thanks

PR #953, opened 2019-01-11
EL-3322 - Focus Directive

ashley-hunter:
- Adding `uxFocusIndicator` directive to allow us to control when the element should show the focus ring
- Added FocusIndicator service which allows us to programmatically control when the focus ring should show
- Added `forRoot` function to the `AccessibilityModule` so if someone wishes to configure the default options for the entire application then they can add this to their `AppModule` also.
- Added Documentation for directive, service and how to configure defaults
- Note, I had to slightly update our focusIf example as it was hijacking document:click events stopping any checkboxes on the page from working
- Added protractor tests for the new directive (the directive just wraps the service, so testing the directive covers both)

#### Ticket
https://autjira.microfocus.com/browse/EL-3322

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3322-Focus-Directive


ashley-hunter:
Thanks for the feedback, I have made some changes:

- Created a separate directive (that just extends the FocusIndicatorDirective) that targets the .btn elements, just so we can add additional classes over time if we need to without polluting the focus indicator directive
- Added accessibility module to all plunkers by default
- Our DocumentationComponentsModule is included in all documentation modules so I exported AccessibilityModule from it so I didn't have to manually add it into all documentation section modules
- Floating action button now imports Accessibility Module so the button focus targets any .btns in its templates
- I can revert these changes if you'd prefer but I thought it might be a good to also include this in the Tabset and Accordion components to start with as we use them so much in our docs site. I have made these use the uxFocusIndicator directive so they only get focus when navigated to using the keyboard rather than on click

PR #958, opened 2019-01-17
EL-3372 - Fixing Protractor Tests

ashley-hunter:
Fixing protractor tests. Note I have commented out one dashboard test as this appears to be a bug in the component (however I cannot reproduce it in our docs, showcase page or plunker, only in this test page). I have raised a ticket to correct this: https://autjira.microfocus.com/browse/EL-3381 after which that test can be included again.

#### Ticket
https://autjira.microfocus.com/browse/EL-3372

#### Documentation CI URL
N/A

### DIFF
--- a/e2e/pages/app/reorderable-table/reorderable-table.testpage.component.html
+++ b/e2e/pages/app/reorderable-table/reorderable-table.testpage.component.html
@@ -18,7 +18,7 @@
             <td class="drag-handle-column">
                 <span uxReorderableHandle id="handle-{{ idx }}" class="hpe-icon hpe-more hpe-rotate-90"></span>
             </td>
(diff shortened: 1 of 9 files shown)


ashley-hunter:
It appears there are issues when running the tests on the SEPG machine. From what I understand it is failing because chrome is not installed on the machine so we can't run the tests. Any suggestions?

donna-marijne:
We have our own build image at https://github.com/UXAspects/buildenv-image if anything needs to be added.

donna-marijne:
I'd like to merge this, so could you revert the pom.xml change and we will sort that out in another PR? Thanks.

ashley-hunter:
Done, thanks

PR #961, opened 2019-01-23
EL-3322 - Firefox Focus Issue

ashley-hunter:
I'm really not sure how to go about fixing this, essentially there is a check in the CDK to determine if the window has focus or not, (if it does not have focus then it detects any focus event as "programmatic" origin instead of keyboard). For some strange reason Firefox can sometime detect the window as not focused when it is - I can only assume its based on the order event are triggered.

As a workaround I have set it so buttons will show a focus ring when triggered programmatically in Firefox. It will still prevent the focus ring on click, but I suppose there may now be a few cases in Firefox where the focus ring shows up. My thinking was it is better to show a focus ring in a few extra cases than cause accessibility issues by potentially not showing it.

I'm open to any feedback or suggestions around this, a few other alternative solutions could be:
- Make programmatic focus on in all buttons in all browsers
- Remove the automatic button focus control until we do further investigation

#### Ticket
https://autjira.microfocus.com/browse/EL-3322

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3322-Firefox-Focus-Issue

### DIFF
--- a/src/directives/accessibility/accessibility.module.ts
+++ b/src/directives/accessibility/accessibility.module.ts
@@ -1,4 +1,5 @@
 import { A11yModule } from '@angular/cdk/a11y';
+import { PlatformModule } from '@angular/cdk/platform';
 import { ModuleWithProviders, NgModule } from '@angular/core';
 import { DefaultFocusIndicatorDirective } from './focus-indicator/default-focus-indicator.directive';
 import { FocusIndicatorDirective } from './focus-indicator/focus-indicator.directive';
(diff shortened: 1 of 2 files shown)


donna-marijne:
Hi Ashley, thanks very much for looking into this. For the time being, can we make programmatic focus default on? This will preserve the behavior of the previous release. We also have yet to consider the case where a programmatic focus is triggered by a keyboard operation, e.g. when launching a modal and assigning focus. You are correct to state that showing the focus ring to all users is better than potentially hiding it to a keyboard user.

Could you create a ticket in the backlog for us to discuss in a future storytime?

PR #971, opened 2019-01-29
EL-3378 - Toolbar Search - Always Expanded

ashley-hunter:
Related PR: https://github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/pull/338

- Added `alwaysExpanded` option to make sure the input never closes
- Updated documentation to include the new property
- The placeholder element was being created using `document.createElement`, now switched to use renderer and correctly dispose of the element. Also added a safety check so we don't try to use `getComputedStyle` in a Server Side Rendered environment.
- Changed toolbar search component to use `OnPush` change detection
- There were quite a few subscriptions that were never disposed of so I have added in the teardown code
- We supported ngModel however we did not support ReactiveForms so I have changed this to use the `ControlValueAccessor` allowing us to target both
- Added new tests to cover the `alwaysExpanded` option
- Updated tests to use reactive forms as well as ngModel

#### Ticket
https://autjira.microfocus.com/browse/EL-3378

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3378-Toolbar-Search

### DIFF
--- a/docs/app/pages/components/components-sections/search/toolbar-search/toolbar-search.component.html
+++ b/docs/app/pages/components/components-sections/search/toolbar-search/toolbar-search.component.html
@@ -38,7 +38,7 @@
             </button>
         </ux-toolbar-search>
     </div>
-    
+
 </div>
 
 <p class="m-t-md">Searched for: <strong>{{searchedFor}}</strong></p>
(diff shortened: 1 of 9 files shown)


donna-marijne:
Could you please go through this with Alana so that she can review your changes and also get a feel for the code review process?

PR #983, opened 2019-02-11
EL-3406 - Focus Rework

ashley-hunter:
- Using `uxFocusIndicator` directive (or service when applying to a host element) to prevent focus on click
- Updated the `uxFocusIndicator` directive to allow overriding configs by injecting new settings using the `ACCESSIBILITY_OPTIONS_TOKEN` in a components providers.
I have also documented this.
- Reorderable Table and Draggable Cards examples have been updated to use immutable object as `*ngFor` sometimes did not update correctly resulting in duplicate rows/cards
- Had to change the `uxFocusWithin`/`uxBlurWithin` directive. They were previously using the `FocusMonitor` (with check children enabled) this meant that any child elements were
getting a "programmatic" focus origin instead of mouse/keyboard preventing us from showing the correct focus ring. Update the directives to use native focus/blur events instead of focus monitor.

Found a few other issues when working my way through all the components which I also fixed:
- Contacts Overflow Documentation pointing to an invalid template path
- Facet Typeahead template - added a null check as value could sometime be null causing a console error
- Media Player - There were some issues with fullscreen/exit full screen, now should be working correctly
- Tag Input - There was a console error when toggling the tag input by clicking in the input field. We were checking if `this.typeahead.open` was true, whenever
`this.typeahead` is null when not in multiple select mode. Updated the condition to account for this.

#### Ticket
https://autjira.microfocus.com/browse/EL-3406

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3406-Focus-Rework


ashley-hunter:
Thanks for the feedback. The fixes should now be made. In regards to the error when navigating from the search section, this has been fixed as part of https://github.com/UXAspects/UXAspects/pull/981.

PR #984, opened 2019-02-11
EL-3399 - Single Line Overflow Tooltip

ashley-hunter:
When the text changes we now destroy any previous tooltip and create a new one.

#### Ticket
https://autjira.microfocus.com/browse/EL-3399

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3399-Single-Line-Overflow-Tooltip


ashley-hunter:
That's for the feedback, turns out this is an issue in Bootstraps tooltip destroy code, essentially if we destroy the tooltip before the hide animation completes it causes an error (https://github.com/twbs/bootstrap/issues/21830). They are not going to fix this in v3 so I have added a workaround which removes the tooltip transition, destroys it, then restores the transition.

PR #988, opened 2019-02-19
EL-3404 - Focus Visual Rework

ashley-hunter:
Related PR: https://github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/pull/355 & https://github.houston.softwaregrp.net/caf/quantum/pull/625

- Created a `ContrastService` that can be used to get the best contrasting colours. I found the WCAG AA Contrast Ratio algorithm in the spec and have implemented it so the contrast will be accurate according to the spec.
- Also created a `uxColorContrast` directive to set the color based on the background color automatically.
- Improved facet header focus behavior
- Improved pagination focus behavior (previous using CDK focus directives, now using `uxFocusIndicator` for consistency)
- Wizard now using focus-indicator. Also found a bug where the first wizard step was not initially tabbable, as we were basing the tab index of whether or not the step had been visited before. Changed to take into account if the step is the first step and if the step is visited.
- Added `uxFocusIndicator` to marquee wizard close button (sample code)

#### Ticket
https://autjira.microfocus.com/browse/EL-3404

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3404-Focus-Visual-Rework

### DIFF
--- a/src/components/color-picker/color-picker.component.less
+++ b/src/components/color-picker/color-picker.component.less
@@ -30,6 +30,10 @@ ux-color-picker {
             height: 32px;
             box-shadow: 0 0 6px 2px #ddd;
 
+            > .hpe-icon {
+                display: none;
+            }
+
             &.ux-focus-indicator-active {
                 outline: none !important; // required until we move all focus styling to box-shadows
                 box-shadow: 0 0 0 3px fade(@primary, 50%);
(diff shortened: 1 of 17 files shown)


src/components/color-picker/color-picker.component.less
```diff
@@ -55,10 +59,8 @@ ux-color-picker {
         }
 
         &.ux-selected {
-            border: 5px solid fade(@primary, 50%);
-
-            > button {
-                box-shadow: none;
+            > button > .hpe-icon {
+                display: block;
```
donna-marijne: Could we add `font-weight: 700` here, to increase visibility of the checkmark a bit?

ashley-hunter: Thanks, I will created a ticket now. Added in the `font-weight` style

PR #996, opened 2019-02-22
EL-3416 - Auto Grow Improvements

ADuff1912:
- Updated AutoGrow directive
- Email discussing the multi-line textarea suggested:
```
Will we support the expanding text area capability as well? (in which case we could perhaps have an accordion box with a control for enabling that as well for the example).
```
- Currently there was no option to disable/enable autogrow functionality.

#### Ticket
https://autjira.microfocus.com/browse/EL-3416

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3416-Auto-Grow-Improvements

### DIFF
--- a/docs/app/pages/components/components-sections/input-controls/expanding-text-area/expanding-text-area.component.html
+++ b/docs/app/pages/components/components-sections/input-controls/expanding-text-area/expanding-text-area.component.html
@@ -16,7 +16,8 @@
     If the content becomes larger than the max height a vertical scrollbar will appear.
 </p>
 
-<p>To create an expanding textarea use the <code>uxAutoGrow</code> directive.</p>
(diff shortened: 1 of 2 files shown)


donna-marijne:
Could you add the URL of your ticket and the CI URL to this PR please?

donna-marijne:
I also need you to update the description of the change to tell me what changes you have made and most importantly why, because the ticket relates to a style update, and did not call for any changes to this directive. Thanks.

ADuff1912:
Description updated.

donna-marijne:
Thanks for this. However this is not part of EL-3416 which is about updating the quantum style and adding a documentation section. This was noted in the same email conversation that you are quoting, which is why the ticket description was not updated. Hope this makes sense.

PR #1006, opened 2019-03-06
EL-3428 - Tree grid toggle from indeterminate state

donna-marijne:
* Click event will now set the selected state directly instead of calculating indeterminate state.

#### Ticket
https://autjira.microfocus.com/browse/EL-3428

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3428-treegrid-indeterminate


donna-marijne:
Thanks for catching this! I think this alternate fix should account for both scenarios.

PR #1007, opened 2019-03-06
EL-3335 - Date Range Picker QA

ashley-hunter:
1. Keyboard focus starts in the "end date" panel - should start in the left panel.
- Fixed

2. Focus is not initially visible when using the keyboard.
- Fixed (by fixing 1, this fixes this)

3. Screen reader - no indication of which panel the user is editing.
- Added aria labels to the start and end date pickers

4. "Select Start Date" and "Select End Date" labels do not appear to be localizable
- Added Inputs to localize these

5. There does not appear to be a way to localize the aria-label on the "Today" button
- Added Input to provide an aria label

#### Ticket
https://autjira.microfocus.com/browse/EL-3335

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3335-Date-Range-Picker-Fixes


docs/app/pages/components/components-sections/date-time-picker/date-range-picker/date-range-picker.component.html
```diff
@@ -146,6 +146,21 @@
...
+        Defines the title to display above the start picker.
+    </tr>
+    <tr uxd-api-property name="selectEndTitle" type="string" defaultValue="Select End Date">
+        Defines the title to display above the end picker.
+    </tr>
+    <tr uxd-api-property name="nowBtnAriaLabel" type="string" defaultValue="Set date to now">
+        Defines the aria label of the now button.
+    </tr>
+    <tr uxd-api-property name="startPickerAriaLabel" type="string" defaultValue="Range Start Date Picker">
+        Defines the aria label of the range start date picker button.
```
donna-marijne: I don't think "button" is intended here.

donna-marijne:
Are there still more changes to make? 2 is still an issue.

ashley-hunter:
Thanks for the feedback. Those issues should now be fixed!

ashley-hunter:
Hmm, when I tab to the right panel I can use arrow keys, is there something I need to do differently?

donna-marijne:
Hopefully this shows what I'm doing:
https://i.gyazo.com/3e134f06ddd91951b9ffda602d604a5b.gif

donna-marijne:
Seems like the problem is related to making a selection in the left pane before tabbing over.

ashley-hunter:
Ah right, I see, its when the focused date is disabled. Thanks

donna-marijne:
Ideally the initial focus on the right will be the selected start date. This may need some special handling if the right panel is entirely disabled, e.g.
image

Possibly the right panel should move on to the first month containing selectable days in this case (and vice versa if right panel is selected first).

ashley-hunter:
Made some changes, if there is no end selection made then when we select the range start it will show the same month in the end range picker (and vice versa) so that there is a valid date to tab to.

ashley-hunter:
Thanks for the feedback, hopefully those issues should now be fixed

ashley-hunter:
Thanks for finding that issue, should be fixed now, thanks!

PR #1025, opened 2019-03-21
Fab-Test-Improvements

ashley-hunter:
The FAB test error you encountered is caused when an element is not present in the DOM. Previously we were triggering a click but not waiting for the promise to resolve which could potentially lead to the expectation running before the click had completed.

Updated tests to use the async/await pattern for all events which should hopefully prevent test flakiness. 

#### Ticket
N/A

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_FAB-Test-Improvements

### DIFF
--- a/e2e/tests/components/common/common.spec.ts
+++ b/e2e/tests/components/common/common.spec.ts
@@ -1,94 +1,94 @@
 import { browser, ElementFinder } from 'protractor';
 
 export class Constants {
-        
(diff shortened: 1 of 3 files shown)


donna-marijne:
Thanks! Since this doesn't affect the release, I'll merge it post-release.

PR #1029, opened 2019-03-25
EL-3465 - Virtual Scroll Tabbable

ashley-hunter:
- Updated example to use tabbable list
- Updated uxVirtualFor and Tabbable List directives to work as expected
- Added tests

#### Ticket
https://autjira.microfocus.com/browse/EL-3465

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3465-Virtual-Scroll-Tabbable

### DIFF
--- a/docs/app/pages/components/components-sections/scrollbar/scrollbar.module.ts
+++ b/docs/app/pages/components/components-sections/scrollbar/scrollbar.module.ts
@@ -3,7 +3,7 @@ import { CommonModule } from '@angular/common';
 import { ComponentFactoryResolver, NgModule } from '@angular/core';
 import { FormsModule } from '@angular/forms';
 import { RouterModule } from '@angular/router';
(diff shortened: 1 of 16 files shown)


ashley-hunter:
Currently, the Home/End functionality is present in the tabbable list directive, however you need to enable it through an input. However, additional specific code was required to get this working in a virtualFor list so I have added in the code to make those buttons work.

ashley-hunter:
Hi Donna, do I need to do anything specific to get this to happen? I don't seem to be able to reproduce the issue?

donna-marijne:
Ah, I was testing in Firefox, didn't notice that it was browser specific.

ashley-hunter:
Hi Donna, I have made a change which should improve the issue. Essentially, Firefox and IE do not support the `preventScroll` option when focusing an element, which meant on scroll the virtual elements changed and we had to focus the new element, chrome was only showing the outline, whereas Firefox and IE were bringing it back into view.

I have added a fix where we store the scroll position before we focus and then restore it after focus only when the browser is not chrome. It can be a little bit janky as we are setting the scroll position we are losing some of the scroll momentum, but from what I can find online its the only solution that seems to work for people.

PR #1043, opened 2019-04-01
EL-3426 - Button Dropdown

ashley-hunter:
- Updating button dropdown example to make it keyboard accessible.
- Updated our `uxMenuNavigation` directives to allow for nested menus
- Updated our focus indicator to ensure we only monitor every element once - eg, `uxMenuNavigationItem` directive and `uxMenuNavigationToggle` directives both try to add focus monitoring, and if we use both previously the onfocus event fires twice, detecting the correct origin the first time but not the second causing the focus indicator not to show in cases where it should.

#### Ticket
https://portal.digitalsafe.net/browse/EL-3426

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3426-Button-Dropdown


ashley-hunter:
Thanks for the feedback, I have updated the directive to return focus to the toggle button on escape. I have also documented the inputs/outputs of the menu navigation directives.

PR #1044, opened 2019-04-02
EL-3479 - Tabbable List Fixes

ashley-hunter:
Made some changes to tabbable list directives. Previously when the items in the list change, we either try to refocus the item that was previously focused or if it is no longer present then we focus the first item to ensure there is always at least one tabbable item. 

We need the refocusing to occur as in a virtualFor list cells are reused, ie. the element that previously displayed the data for item one might now display the data for cell two, if cell two was previously focused so we should make the cell that now displays the data for cell two receive focus.

However this has an unintended consequence when filtering by search, as while searching the items in the list change, so if we do a search that returns no items, then clear the search, it tries to focus the first item in the list to ensure there is always a tabbable item - removing focus from the search input.

To fix this I have changed the `uxTabbableList` to to capture the current focused state, so if an item in the list is focused and the list items changed then we will refocus, if no items in the list are focused (for example because the search input is focused) and the list items change we simply update the tab indexes to ensure there is a tabbable item rather than actually performing the focus.

I also added quite a few protractor tests covering all the basic behavior to prevent any further regressions.

#### Ticket
https://portal.digitalsafe.net/browse/EL-3479

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3479-Tabbable-List-Fixes


ashley-hunter:
Hi Donna, in their plunker they have the url for the library hard coded to the released library in the systemjs config file, it just needs updated to point to the SEPG library too and it should work without any changes

PR #1046, opened 2019-04-03
Tabbable List Fix

ashley-hunter:
I have just discovered a bug in the tabbable list that I would like to get fixed before release. 

It is hard to reproduce, but i can get it to happen consistently when hiding a wizard step. Essentially, previously any time the tab indexes in a tabbable list changed we updated the tabindex attributes by running change detection, which seems to cause an issue in this case as change detection was already running. Switching to setting the tabindex using the renderer instead of relying on CD seems to resolve this issue.

https://plnkr.co/edit/uit7Wgd4wTdUDXdjLwGs?p=preview

#### Ticket
https://portal.digitalsafe.net/browse/EL-3479

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_Tabbable-List-Issue-Fix

### DIFF
--- a/e2e/tests/components/tabbable-list/tabbable-list.e2e-spec.ts
+++ b/e2e/tests/components/tabbable-list/tabbable-list.e2e-spec.ts
@@ -301,6 +301,21 @@ describe('Tabbable List Tests', () => {
         expect(await page.getFocused(4)).toBe(false);
     });
 
+    it('should ensure there is always one tabbable item when the list changes', async () => {
+
+        // expect the third document to not be tabbable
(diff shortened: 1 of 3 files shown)


ashley-hunter:
Sorry, my mistake for not clarifying. The expression has change error will be something we would likely want to fix as part of https://portal.digitalsafe.net/browse/EL-3467 as it is an existing issue with the wizard when we hide steps. The error regarding the tabbable list is the following, which should no longer appear:

image

PR #1052, opened 2019-04-09
EL-3117 - Partition Map

ashley-hunter:
**Bugs Fixed**
- Found an issue in our Timeline chart plugin, we were importing chart using the `import * as` syntax which is never tree shakable. This is problematic, because a blank app that imports anything from UX Aspects library automatically includes chart.js library, which by default also includes moment.js. This was increasing the application bundle of a nearly empty app to nearly 1mb. Chart.js is registered as a global when anyone imports it or imports ng2-charts, so I have changed the timeline plugin to reflect this. We also need the `@types/chartjs` package to be moved to our dependencies instead of devDependencies as an application that install the Aspects library will not install devDependencies, as as we are no longer directly importing chart.js we would get a build error if the type definitions are not present as we do still reference the chart js types in our code for type safety.

- Found an issue in the color contrast service, we passed it in a light colour and dark colour however it always was returning `#fff` or `#000`. We were also forcibly returning hex codes which means we cannot have a rgb, value or an rgba which could be required if an alpha value is needed etc. So now returns a the appropriate light or dark `ThemeColor` that was passed in.

- When building a seed project in production mode we get 2 "WARNING in Invalid font values at 11971:2. Ignoring." warnings from our style sheet. I have fixed the font properties used so these issues no longer appear.

- Fixed issue that tooltip can flicker when mouse moves over it, this is because by default the CDK detects mouse events on overlays to support things like popovers. I have added updated our tooltip to override the default `pointer-events` behavior which should stop any future flickering issues.

**Partition Map**

- Added partition map component
- Added documentation and plunker
- Added Karma tests (also added karma tests to the test step of the build)

#### Ticket
https://portal.digitalsafe.net/browse/EL-3117

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3117-Partition-Map


ashley-hunter:
In regards to the tooltip I emailed Roland about their appearance, he suggested to use regular tooltips for now and we would revisit it at a later point:

> I like the tooltip. I don’t think we need to do anything fancy at this stage.

As far as the chart.js change goes, there shouldn't be any breaking change, as you need to import ng2-charts to use any of our charts, this imports chart.js automatically, so the import we had before was unnecessary. But I am happy to revert this if you prefer.

donna-marijne:
That's okay, I misunderstood the description in that case. If there is no impact to library consumers then it's a fine change to make.

ashley-hunter:
Review comments should now be addressed. I also double checked the chart.js change in our seed project to verify there is no breaking change.

donna-marijne:
Thanks for making the changes and double checking the chart import.

I just wanted to check if there is any feasible way of enforcing the minSegmentWidth parameter? I noted that segments shrink below this limit after a point. I think that the preferable behavior would be prevent the chart from being shrunk below that threshold. However if it's not feasible then perhaps we can instead mention that minSegmentWidth is not a hard limit and that min-width should be used on the chart to prevent segments from shrinking too far.

ashley-hunter:
Hi Donna, this is something I tried out quite a bit and there is no way to guarantee that this minimum width can be achieved as there are many layouts which would make it impossible to achieve the minimum node width based on the number of nodes and the available width, especially if they increase the minimum width in a chart will many segments.  Currently any widths less than the minimum are increased as much as possible until there is no more space. I will update the description to make this more clear

donna-marijne:
Thanks, that will be fine, I just wanted to check!

ashley-hunter:
Thanks, description updated

PR #1053, opened 2019-04-11
Added option to use Less and CSS sourcemap

donna-marijne:
I've been using this config to identify where style rules are defined. It would be useful to make this part of the webpack config.

````
SOURCEMAP=true npm start
````

#### Ticket
N/A

#### Documentation CI URL
N/A


ashley-hunter:
Looks good! Just checking, is there a reason why we wouldn't want this on all the time in the dev build? In which case we could just set it to true and skip the environment variable?

donna-marijne:
Certainly! I think the only downside is somewhat slower compilation and the page looks a bit wonky on first load. I'll set it to true and we can back that out if it's too annoying.

ashley-hunter:
Sounds good!

PR #1054, opened 2019-04-12
EL-3431 - Nested Donut Chart

ashley-hunter:
- Added Nested Donut Chart
- Added documentation section and plunker
- Added tests

Also made a few tweaks to the tooltip component. I wanted to use a tooltip in the chart but currently the tooltip can only be launched using the tooltip directive (which cannot be applied to dynamically generated SVG elements). I have changed the tooltip component to allow properties to be set using inputs as well, so I can reuse the existing tooltip component without the directive.

#### Ticket
https://portal.digitalsafe.net/browse/EL-3431

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3431-Nested-Donut-Chart

### DIFF
--- a/src/components/nested-donut-chart/nested-donut-chart.component.less
+++ b/src/components/nested-donut-chart/nested-donut-chart.component.less
@@ -1,0 +1,33 @@
+ux-nested-donut-chart {
+    display: inline-block;
+    position: relative;
+    width: 165px;
+    height: 165px;
+
+    .ux-nested-donut-chart-content {
+        position: absolute;
+        display: flex;
+        justify-content: center;
+        align-items: center;
+        border-radius: 50%;
+    }
+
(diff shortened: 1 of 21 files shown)


src/components/nested-donut-chart/nested-donut-chart.component.less
```diff
@@ -0,0 +1,33 @@
+ux-nested-donut-chart {
+    display: inline-block;
+    position: relative;
+    width: 165px;
+    height: 165px;
```
donna-marijne: This width and height should be part of the example css. By default, the component ought to fit the size of its parent container.

ashley-hunter:
Thanks for the feedback, I have now moved the CSS to set the size to the example code.

PR #1056, opened 2019-04-12
EL-3467 - Wizard - step modifications not updating UI

donna-marijne:
* Added `tick` operator to the steps change handler, to allow the view to update.
* Also added an `update` call to keep the step index accurate.
* Added a test case for adding and removing a final step.
* Consolidated a number of wizard tests into one test case for efficiency.

#### Ticket
https://portal.digitalsafe.net/browse/EL-3467

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3467-wizard-step-modifications

### DIFF
--- a/e2e/pages/app/wizard/wizard.module.ts
+++ b/e2e/pages/app/wizard/wizard.module.ts
@@ -1,11 +1,13 @@
+import { CommonModule } from '@angular/common';
 import { NgModule } from '@angular/core';
 import { RouterModule } from '@angular/router';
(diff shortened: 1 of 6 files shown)


donna-marijne:
I'm not sure about this. If someone chooses to replace or remove an already visited or active step, they should probably decide to either reset the wizard or reassign the step parameter. The behavior would depend on individual requirements I think.

PR #1058, opened 2019-04-12
EL-3466 - Sortable Header Icon Alignment

ashley-hunter:
- Adding option to inline the icon so it appears immediately after the text rather than right aligned

#### Ticket
https://portal.digitalsafe.net/browse/EL-3466

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3466-Inline-Sort-Icon


ashley-hunter:
Ah my mistake, I didn't realise single column sorting and multiple column sorting were two different directives, now updated multiple column sorting directive to also allow inline icons

ashley-hunter:
Thanks for the feedback, the overflow issue should now be resolved

PR #1060, opened 2019-04-16
EL-3402 - Select Icon

ashley-hunter:
- Adding option to customize the chevron icon
- Added example
- Updated sample code
- Added missing directories from the grunt `clean` task

#### Ticket
https://portal.digitalsafe.net/browse/EL-3402

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3402-Select-Icon


docs/app/pages/components/components-sections/select/select/select.component.less
```diff
@@ -0,0 +1,18 @@
+.custom-icon {
```
donna-marijne: We could add these rules to the library, and document it as a kind of helper class (ux-select-custom-icon) perhaps. It would save some copy/pasting.

ashley-hunter:
Thanks for the suggestion. I have moved them to the stylesheet and updated the example.

PR #1061, opened 2019-04-16
EL-3117 - Partition Map Collapse Height

ashley-hunter:
As discussed yesterday during storytime, I have updated the partition map to use pixel values for the collapsed height rather than a percentage.

#### Ticket
https://portal.digitalsafe.net/browse/EL-3117

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3117-Partition-Map-Collapse-Height


ashley-hunter:
Thanks for the feedback, I have fixed the bug and moved the sizes to the example css

PR #1066, opened 2019-04-23
EL-3496 - Tree Grid Performance Improvements

ashley-hunter:
After some investigation I found that the poor tree grid performance is largely not down to the tree grid but down to the list hover actions and tooltips being used on table grid rows. Another bottleneck is the large amount of event listeners being added to the DOM on each row.

**List Hover Actions**
- Refactored the code to use less event listeners
- Used lazy event listener directives
- Tooltips are now not added using the tooltip directive as it is very expensive to use. Instead we attach a one time mouse enter event listener, which only when triggered will the tooltip be set up. In our example we had 3 hover actions per row, so with 100 rows this would be 300 tooltips being set up on initial render, now we only set up if the user tries to show it allowing us to skip a lot of logic. The the previous directive also caused a lot of major garbage collection which slowed the UI even further.

**Tree Grid**
- Tree grid templates now also use the same tooltip technique as mentioned above.

**Event Listeners**
- Create a new set of event directives thats attach themselves to the elements after they have rendered rather than while being rendered.

#### Ticket
https://portal.digitalsafe.net/browse/EL-3496

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3496-Tree-View-Performance


ashley-hunter:
Thanks for the feedback, the issue you mentioned should now be resolved

PR #1073, opened 2019-04-30
EL-3468 - Side Panel focus

donna-marijne:
* Using Angular animations in order to set `visibility` with the proper timing. `visibility: hidden` prevents focus landing on panel content when it is closed.
* Converted the duration and timing function to match quantum.

#### Ticket
https://portal.digitalsafe.net/browse/EL-3468

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3468-side-panel

### DIFF
--- a/docs/app/pages/components/components-sections/panels/item-display-panel/item-display-panel.component.html
+++ b/docs/app/pages/components/components-sections/panels/item-display-panel/item-display-panel.component.html
@@ -132,6 +132,10 @@
 
 <p>To add a footer to the item display panel the <code>uxItemDisplayPanelFooter</code> directive can be added as an attribute allowing you to add any required content to the bottom of the item display panel</p>
 
+<blockquote>
+    <p><strong>Note</strong>: The root module must import <code>BrowserAnimationsModule</code> from <code>'@angular/platform-browser/animations'</code>.</p>
+</blockquote>
+
 <p>The following code will produce the example shown above:</p>
 
 <ux-tabset [minimal]="false">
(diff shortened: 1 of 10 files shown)


ashley-hunter:
There appears to be Side Panel tests failing at the moment?

donna-marijne:
There's an error on the e2e site. Is there some extra dependency required for loading a component with animations?

ashley-hunter:
Ah, BrowserAnimationsModule will need imported in AppModule

ashley-hunter:
Hmm, actually it looks like it is already there, I will take a look into it now

donna-marijne:
It might be the enum, I just found this: https://github.com/angular/angular/issues/24047

ashley-hunter:
I see someone mentions there to add `export` to the enum and it should work?

ashley-hunter:
Yeah, adding `export` seems to fix it for me

donna-marijne:
For some reasons the animations don't apply to the subclass component ux-item-display-panel, could you take a look and let me know if you can see what's missing?

ashley-hunter:
Sure no problem

ashley-hunter:
That should be it fixed, the `ItemDisplayPanelComponent` class had an `ngOnInit` function so the `ngOnInit` in the `SidePanelComponent` was never getting called which is where the animation state changes were setup.

donna-marijne:
Great catch, thank you very much!

donna-marijne:
Should be good for review again now.

PR #1079, opened 2019-05-03
EL-3269 - Side Menu Route Options

ashley-hunter:
**NavigationItem Interface**
- Added new `routerOptions` property to the interface to contain all router related configuration as we may want to add more options over time. I did not want to use the existing `routerExtras` property as this maps directly to the Angular `routerExtras` properties used by `routerLink`. It also allows use to share the interface with the global config as mentioned below.
- Added option to specify the `exact` property (defaults to true to retain existing functionality)
- Added option to `ignoreQueryParams` (defaults to false to retain existing functionality)

**Navigation Module**
- Added a `forRoot` option to the module. This will allow the `exact` and `ignoreQueryParams` (and any others that we may add in future) to be specified on a global level rather than having to add these properties to ever item in the menu. If a menu item has these properties they will take precedence over the global options.

**Navigation Components**
- Updated to use `OnPush` change detection for better performance
- Did some refactoring and general tidying up, eg. removing `itemClick` function as it is not used anymore.

**Tests**
- Added protractor tests

**Documentation**
- This component is only documented in the Micro Focus documentation site so no changes required

#### Ticket
https://portal.digitalsafe.net/browse/EL-3269

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3269-Side-Menu-Route-Options

### DIFF
--- a/e2e/tests/components/navigation/navigation.e2e-spec.ts
+++ b/e2e/tests/components/navigation/navigation.e2e-spec.ts
@@ -1,0 +1,248 @@
+import { browser } from 'protractor';
+import { NavigationPage } from './navigation.po.spec';
+
+describe('Number Picker Tests', () => {
+
+    let page: NavigationPage;
+
+    beforeEach(() => {
+        page = new NavigationPage();
+        page.getPage();
+    });
+
+    it('should intially be collapsed with no items selected', async () => {
+
+        // there should be 3 top level items
+        const items = await page.getTopLevelItems();
+        expect(items.length).toBe(3);
+
+        // none of these items should be selected
+        expect(await page.isItemActive(items[0])).toBeFalsy();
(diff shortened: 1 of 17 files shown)


e2e/tests/components/navigation/navigation.e2e-spec.ts
```diff
@@ -0,0 +1,248 @@
+import { browser } from 'protractor';
+import { NavigationPage } from './navigation.po.spec';
+
+describe('Number Picker Tests', () => {
```
donna-marijne: Incorrect component name

ashley-hunter:
Thanks for the feedback, incorrect test name has now been fixed.

PR #1085, opened 2019-05-08
EL-3515 - Organization Chart Tests

ashley-hunter:
- Adding tests for Organization chart

#### Ticket
https://portal.digitalsafe.net/browse/EL-3515

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3515-Organization-Chart-Tests


ashley-hunter:
Karma allows me to more easily access the inputs and outputs and have direct DOM access, so I can do things like spy on event emitters to ensure they are called and pass data directly to inputs without having to add lots of buttons to the page to change inputs and print out text to check if outputs are emitted or two way bindings have updated. They also run a lot faster and are easier to debug. 

When looking into how Angular Material do testing, they pretty much use Karma for all component tests, with e2e tests only really being used when the component requires some complex interactions such as dragging which aren't easily achieved through karma.

I can easily change them to Protractor if you'd prefer.

donna-marijne:
No, that makes sense. I don't know how much benefit we really get from protractor anyway. What we really need for e2e is something that compares screen grabs.

ashley-hunter:
Yes, I totally agree, that would be very helpful! Wonder if this would be any use: https://www.npmjs.com/package/protractor-image-comparison

donna-marijne:
Thanks, I'll make a ticket to investigate. Another option was StormRunner but I don't know whether we'd be able to sort out licensing for it.

PR #1086, opened 2019-05-09
EL-3515 - Organization Chart

ashley-hunter:
**Hierarchy Bar**
- Added new option to provide a custom template for the icons, so we can now use icons from out icon set (or any icon set) rather than having to be an image. Also updated documentation to include this.
- Fixed several issues regarding the focus indicator being restored

**Focus Indicator Origin**
- Currently we have no way of apply this functionality to a host element.
- I have separated this out into a class rather than a directive
- The original directive now just uses the class to avoid any duplication

**Accessibility Folder**
- Our `index.ts` file had grown pretty large. It clearly was originally meant to export one or two directives but now that we have many there were a lot of exports. I have made each subfolder have its own `index.ts` file which is then re-exported by the original `index.ts` file. This matches the rest of our components/directives which each have their own `index.ts` file.

#### Ticket
https://portal.digitalsafe.net/browse/EL-3515

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3515-Organization-Chart


docs/app/pages/components/components-sections/hierarchy-bar/hierarchy-bar/snippets/icon.html
```diff
@@ -0,0 +1,3 @@
+<ux-hierarchy-bar [root]="node" [mode]="mode" [(selected)]="selected">
+    <i class="hpe-icon hpe-user-manager" *uxHierarchyBarNodeIcon="let node"></i>
```
donna-marijne: Could we have an example that makes use of the `node` parameter here?

ashley-hunter: Sure, updated it to add a class based on the `node.icon` property

PR #1110, opened 2019-05-31
EL-3540 - Date Range Picker Time

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3540

- Previously that date range duration always calculate the range based on 00:00 on the start date and 23:59 on the end date. I have updated this to now take into account the time provided by the time picker, so if the time was set to 7pm on both start and end time pickers the duration would be 2 days rather than 3 days

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3540-Date-Range-Picker-Duration

### DIFF
--- a/src/components/date-range-picker/date-range.service.ts
+++ b/src/components/date-range-picker/date-range.service.ts
@@ -45,6 +45,12 @@ export class DateRangeService {
     /** Indicate if we are currently changing the time */
     isChangingTime: boolean = false;
 
+    /** Store the current start time */
+    startTime: DateRangeTime = { hours: 0, minutes: 0, seconds: 0 };
+
+    /** Store the current end time */
+    endTime: DateRangeTime = { hours: 23, minutes: 59, seconds: 59 };
+
     setStartDate(date: Date | null): void {
 
         // if the start date is after the end date the clear the end date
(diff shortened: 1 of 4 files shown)


src/components/date-range-picker/date-range.service.ts
```diff
@@ -45,6 +45,12 @@ export class DateRangeService {
     /** Indicate if we are currently changing the time */
     isChangingTime: boolean = false;
 
+    /** Store the current start time */
+    startTime: DateRangeTime = { hours: 0, minutes: 0, seconds: 0 };
```
donna-marijne: Could this not be part of `start`/`end` above?

ashley-hunter: I had to keep them separate because if I select a start date, but do not have an end date (or vice versa), then when I hover over a date in the end picker it shows the duration from the start date to the hover date, so in this instance setting it directly on the `end` property would cause an end date to actually be chosen, rather than just showing a preview of what the duration would be, if that makes sense?

donna-marijne: Understood! Thanks for the clarification.

PR #1122, opened 2019-06-25
EL-3492 - Remove AngularJS source and resources from the main library (additional changes)

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3492

I have some additional changes that could be considered as part of EL-3492 or we may want to create separate tickets for them if they don't fit into the existing ticket:

### Webpack Configuration
- Removed Angular.js from the webpack config - which allows use to remove the entire `scripts.js` bundle and `ScriptsWebpackPlugin`

### Documentation
- Flippable Cards example was using `nested-donut` AngularJS component - switched to use Angular component.
- Converted documentation site to a pure Angular application - removed all hybrid code. This will now allow us to detect expression changed errors in our docs site (which I have now spotted a few that i need to raise tickets for).
- Removed all uses of the `HybridModule` in our documentation site as this is no longer required

### package.json
- We had a load of dependencies that were in `package.json` that were only used by Angular.js so I have removed them
- Removed `dist` and `ng1` files from the list of included files in `src/package.json`

### Miscellaneous
- Fixed some incorrect typings in the `InfiniteScrollDirective`
- Added markdown file to keep track of any breaking changes (you may already already be doing this in which case this can be removed)

### Hybrid Module
~~**[Breaking Change]** I have moved the `HybridModule` to a separate entrypoint (https://github.com/ng-packagr/ng-packagr/blob/master/docs/secondary-entrypoints.md)~~
- ~~We can keep the hybrid code in this repository as it does depend on `@angular/*` packages.~~
- ~~It will no longer be included in the `@ux-aspects/ux-aspects` import, so we no longer require applications to install `@angular/upgrade` if they are not using any hybrid components and we ensure no upgrade or hybrid code gets included in their bundle unnecessarily.~~
- ~~`@angular/upgrade` is now an optional dependency (previously mandatory even if not used)~~
- ~~`HybridModule` would now be imported from `@ux-aspects/ux-aspects/hybrid`~~
- ~~`ng-packagr` handles multiple entry points automatically, no changes to build steps etc.. required.~~

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3492-remove-angularjs-2


donna-marijne:
I overlooked HybridModule, but this has actually been moved to @ux-aspects/ux-aspects-ng1, so the remaining task is to remove it from the main library.

ashley-hunter:
Ah, even better! I have removed it now, thanks

PR #1136, opened 2019-07-17
EL-3564 Masthead Clickable Product Logo

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3564

- Whole of logo container is now clickable
- Event is emitted when clicked
- Unit Tests added for testing event emitted when logo is clicked, back button is clicked and to check the contents of the product header.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3564-Masthead-clickable-product-logo

### DIFF
--- a/docs/app/pages/components/components-sections/page-header/page-header/page-header.component.html
+++ b/docs/app/pages/components/components-sections/page-header/page-header/page-header.component.html
@@ -50,6 +50,9 @@
     <tr uxd-api-property name="backClick" type="EventEmitter">
         The event fired when the back button is clicked.
     </tr>
+    <tr uxd-api-property name="logoClick" type="EventEmitter">
+        The event fired when the image at the top left part of the header is clicked.
+    </tr>
 </uxd-api-properties>
 
 <p>The following interfaces are used within the configuration of the page header:</p>
--- a/src/components/page-header/page-header.component.html
+++ b/src/components/page-header/page-header.component.html
@@ -3,7 +3,7 @@
     <div *ngIf="!condensed" class="page-header-content">
 
         <!-- Logo/product acronym -->
-        <div class="page-header-logo-container" role="presentation" [style.backgroundColor]="logoBackground" [style.color]="logoForeground">
+        <div uxFocusIndicator class="page-header-logo-container" role="presentation" [style.backgroundColor]="logoBackground" [style.color]="logoForeground" (click)="logoClicked()">
             <img *ngIf="logo" [attr.src]="logo" [alt]="header" class="page-header-logo">
             <h1 *ngIf="header && !logo" class="page-header-acronym">{{header}}</h1>
         </div>
--- a/src/components/page-header/page-header.component.ts
+++ b/src/components/page-header/page-header.component.ts
@@ -123,7 +123,10 @@ export class PageHeaderComponent {
     }
 
     /** Emit whenever the back button is clicked */
-    @Output() backClick = new EventEmitter();
+    @Output() backClick = new EventEmitter<MouseEvent>();
+
+    /** Emit whenever the product logo in the left corner is clicked. */
+    @Output() logoClick = new EventEmitter<MouseEvent>();
 
     /** @deprecated - Access a custom template title. Use subheaderTemplate instead */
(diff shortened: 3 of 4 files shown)


src/components/page-header/page-header.component.ts
```diff
@@ -153,4 +156,8 @@ export class PageHeaderComponent {
     select(item: PageHeaderNavigation): void {
         this._pageHeaderService.select(item);
     }
+
+    logoClicked(): void {
+        this.logoClick.emit();
```
ashley-hunter: The `logoClick` type above mentions this should emit a `MouseEvent`, for this to be true, this line here would need to pass the mouse event into the `emit` function. Same with the `backClick`. Currently the type would actually be `EventEmitter<void>` because the emit function is not being passed anything.

ashley-hunter: In fact these two functions can be removed and we can just call them in the html, eg:
```html
<... (click)="logoClick.emit($event)">
```

docs/app/pages/components/components-sections/page-header/page-header/page-header.component.html
```diff
@@ -50,6 +50,9 @@
     <tr uxd-api-property name="backClick" type="EventEmitter">
         The event fired when the back button is clicked.
     </tr>
+    <tr uxd-api-property name="logoClick" type="EventEmitter">
```
ashley-hunter: We have recently changed how we document Outputs, this was never updated, but as all outputs are EventEmitters we now just specify the type of the event that is emitted, eg: `MouseEvent`

src/components/page-header/page-header.component.ts
TaraTodd123: Functions removed and moved with html

docs/app/pages/components/components-sections/page-header/page-header/page-header.component.html
TaraTodd123: Added <code>MouseEvent</code> to documentation

docs/app/pages/components/components-sections/page-header/page-header/page-header.component.html
```diff
@@ -47,9 +47,12 @@
 </uxd-api-properties>
 
 <uxd-api-properties tableTitle="Outputs">
-    <tr uxd-api-property name="backClick" type="EventEmitter">
+    <tr uxd-api-property name="backClick" type="MouseEvent">
         The event fired when the back button is clicked.
     </tr>
+    <tr uxd-api-property name="logoClick" type="MouseEvent">
+        The event fired when the image at the top left part of the header is clicked.
```
ashley-hunter: Actually, could be remove this `logoClick` documentation here, I'm just thinking that the logo never appears in this theme. Instead we should create a separate PR to document this in the MF documentation only

src/components/page-header/page-header.component.html
```diff
@@ -3,7 +3,7 @@
     <div *ngIf="!condensed" class="page-header-content">
 
         <!-- Logo/product acronym -->
-        <div class="page-header-logo-container" role="presentation" [style.backgroundColor]="logoBackground" [style.color]="logoForeground">
+        <div uxFocusIndicator class="page-header-logo-container" role="presentation" [style.backgroundColor]="logoBackground" [style.color]="logoForeground" (click)="logoClick.emit($event)">
```
ashley-hunter: I'm wondering if it would be good to make this tabbable as a user way want to perform the "click" using the keyboard, as you can see there is no way of doing this currently:

https://plnkr.co/edit/SCvXpWj52MuKueHXGjeR?p=preview

docs/app/pages/components/components-sections/page-header/page-header/page-header.component.html
TaraTodd123: Removed and will add to micro focus site documentation

src/components/page-header/page-header.component.html
TaraTodd123: tab index added

PR #1142, opened 2019-07-23
masthead clickable updates

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3564

- Making logo focusable/clickable only when logoClick event is emitted

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3564-masthead-clickable-logo

### DIFF
--- a/src/components/page-header/page-header.component.html
+++ b/src/components/page-header/page-header.component.html
@@ -3,7 +3,7 @@
     <div *ngIf="!condensed" class="page-header-content">
 
         <!-- Logo/product acronym -->
-        <div uxFocusIndicator (keydown.enter)="logoClick.emit($event)" [keyboardFocusIndicator]="hasLogoClick" [attr.tabindex]="hasLogoClick ? 0 : -1" [class.clickable]="hasLogoClick" class="page-header-logo-container" role="presentation" [style.backgroundColor]="logoBackground" [style.color]="logoForeground" (click)="logoClick.emit($event)">
+        <div uxFocusIndicator (keydown.enter)="logoClick.emit($event)" [attr.tabindex]="_hasLogoClick ? 0 : -1" [class.clickable]="_hasLogoClick" class="page-header-logo-container" role="presentation" [style.backgroundColor]="logoBackground" [style.color]="logoForeground" (click)="logoClick.emit($event)">
             <img *ngIf="logo" [attr.src]="logo" [alt]="header" class="page-header-logo">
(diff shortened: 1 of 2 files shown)


src/components/page-header/page-header.component.ts
```diff
@@ -152,4 +152,8 @@ export class PageHeaderComponent {
     select(item: PageHeaderNavigation): void {
         this._pageHeaderService.select(item);
     }
+
+    get hasLogoClick(): boolean {
```
ashley-hunter: Just a small thing, generally we put all variables including getters/setters above the constructor, this way every component is consistent and it makes everything very easy to find. 

We also generally prefix private variables with an `_`, I realise this is not technically private and can't be because it needs to be public to access it in the template, however when we have a public property that should only be used in the template it is often good to also prefix it with an underscore, this way anyone consuming this component should know this is an internal property and it should not be used.

src/components/page-header/page-header.component.html
```diff
@@ -3,7 +3,7 @@
     <div *ngIf="!condensed" class="page-header-content">
 
         <!-- Logo/product acronym -->
-        <div uxFocusIndicator tabindex="0" class="page-header-logo-container" role="presentation" [style.backgroundColor]="logoBackground" [style.color]="logoForeground" (click)="logoClick.emit($event)">
+        <div uxFocusIndicator (keydown.enter)="logoClick.emit($event)" [keyboardFocusIndicator]="hasLogoClick" [attr.tabindex]="hasLogoClick ? 0 : -1" [class.clickable]="hasLogoClick" class="page-header-logo-container" role="presentation" [style.backgroundColor]="logoBackground" [style.color]="logoForeground" (click)="logoClick.emit($event)">
```
ashley-hunter: Do we need `[keyboardFocusIndicator]="hasLogoClick"`? I would have assumed if the item has a tab index of `-1` when it is not tabbable then it would not get focused? I could be wrong, just checking.

src/components/page-header/page-header.component.ts
TaraTodd123: Moved to above the constructor an added in <code>_</code> prefix

src/components/page-header/page-header.component.html
TaraTodd123: Removed <code>[keyboardFocusIndicator]="hasLogoClick"</code>, checked and it isn't needed.

PR #1149, opened 2019-07-29
EL-3605 - Tag Input Reactive Forms

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3605

- Implemented `Validator` interface to ensure we remain compatible
- Added support for marking component as touched
- Added error styling based on the presence of the `ng-invalid` class which will get applied for both template and reactive form errors.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3605-Tag-Reactive-Form

### DIFF
--- a/src/components/tag-input/tag-input.component.less
+++ b/src/components/tag-input/tag-input.component.less
@@ -115,7 +115,9 @@ ux-tag-input {
         border-color: @form-control-focus-border;
     }
 
-    &.invalid {
+    &.invalid,
+    &.ng-invalid.ng-touched:not(.focus),
+    &.ng-invalid.ng-dirty {
         border-color: @critical;
     }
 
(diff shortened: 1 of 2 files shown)


ashley-hunter:
Here is a plunker showing reactive form validation in action:

https://plnkr.co/edit/ouajvfQf2FhZGXM1bZWV?p=preview

donna-marijne:
I'm not sure why but in the plunker the control is marked as valid if it has 0 tags. Are you able to reproduce?

ashley-hunter:
The `minLength` validator does not validate empty values. For example, a phone number might be an optional field (empty is valid), but if the user enters a value then the length should be 11 digits in length. If you want to invalidate when the list is empty you would also add the `Validators.required` validator.

PR #1155, opened 2019-08-02
EL-3559 - Side Panel Improvements

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3559

- Side Panel content is now not inserted into the DOM until it is opened
- Side panel now uses `OnPush` change detection for better performance
- Updated tests
- The `openChange` event was emitted initially even though it did not change. Now only emits once a change has occurred
- Pressing escape previously triggered the close event even if the side panel was not open
- Added utility functions to determine if a `click` event is triggered by the mouse or by the enter key
- Updated focus indicator directive to coerce boolean values, this way we can set the values using a string instead of a binding if the values never change, allowing Angular to skip checking these during change detection eg. `[programmaticFocusIndicator]="false"` could now just be `programmaticFocusIndicator="false"` and will now only be initially set.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3559-Side-Panel-Improvements

### DIFF
--- a/src/components/side-panel/side-panel.component.ts
+++ b/src/components/side-panel/side-panel.component.ts
@@ -1,6 +1,6 @@
 import { ChangeDetectionStrategy, Component, ElementRef, EventEmitter, HostBinding, HostListener, Input, OnDestroy, OnInit, Output } from '@angular/core';
 import { Subject } from 'rxjs';
-import { takeUntil } from 'rxjs/operators';
+import { distinctUntilChanged, takeUntil } from 'rxjs/operators';
 import { FocusIndicatorOriginService } from '../../directives/accessibility/index';
 import { SidePanelAnimationState, sidePanelStateAnimation } from './side-panel-animations';
 import { SidePanelService } from './side-panel.service';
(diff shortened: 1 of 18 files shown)


src/components/side-panel/side-panel.component.ts
```diff
@@ -88,53 +90,64 @@ export class SidePanelComponent implements OnInit, OnDestroy {
...
-    ngOnInit() {
+    ngOnInit(): void {
         this.service.open$.pipe(takeUntil(this._onDestroy)).subscribe(isOpen => {
             this.animationPanelState = isOpen
                 ? this.animate
                     ? SidePanelAnimationState.Open
                     : SidePanelAnimationState.OpenImmediate
                 : SidePanelAnimationState.Closed;
+
+            // only if the open state changed should we emit the latest value
```
donna-marijne: Should we have `distinctUntilChanged` above to ensure this?

ashley-hunter: Good idea, added in `distinctUntilChanged` operator.

PR #1156, opened 2019-08-02
El 3608 select clear button follow up

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3608

- removed tab index on clear is disabled
- greyed out clear button when disabled
- returned focus to input when clear button is activated via keyboard

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3608-select-clear-button-follow-up

### DIFF
--- a/src/components/select/select.component.html
+++ b/src/components/select/select.component.html
@@ -66,7 +66,7 @@
     <div class="ux-select-icons">
         <i *ngIf="clearButton && allowNull && value"
            uxFocusIndicator
-           [attr.tabindex]="disabled ? null : 0"
+           [attr.tabindex]="disabled ? -1 : 0"
            [attr.aria-label]="clearButtonAriaLabel"
            class="ux-select-icon ux-icon ux-icon-close ux-select-clear-icon"
            (click)="clear(); $event.stopPropagation()"
--- a/src/components/select/select.component.html
+++ b/src/components/select/select.component.html
@@ -66,13 +66,12 @@
     <div class="ux-select-icons">
         <i *ngIf="clearButton && allowNull && value"
            uxFocusIndicator
-           tabindex="0"
+           [attr.tabindex]="disabled ? null : 0"
            [attr.aria-label]="clearButtonAriaLabel"
            class="ux-select-icon ux-icon ux-icon-close ux-select-clear-icon"
            (click)="clear(); $event.stopPropagation()"
            (keydown.enter)="clear(); $event.stopPropagation()">
         </i>
-
         <i *ngIf="!icon"
            class="ux-select-icon ux-icon ux-select-chevron-icon"
            [class.ux-icon-up]="dropDirection === 'up'"
--- a/src/components/tag-input/tag-input.component.html
+++ b/src/components/tag-input/tag-input.component.html
@@ -38,7 +38,7 @@
 </ol>
 
 <!-- Insert the custom icon if provided -->
-<div class="ux-tag-icons" [class.disabled]="disabled" *ngIf="icon || _showClearButton">
+<div class="ux-tag-icons" *ngIf="icon || _showClearButton">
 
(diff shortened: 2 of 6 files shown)


src/components/select/select.component.html
```diff
@@ -66,13 +66,12 @@
     <div class="ux-select-icons">
         <i *ngIf="clearButton && allowNull && value"
            uxFocusIndicator
-           tabindex="0"
+           [attr.tabindex]="disabled ? null : 0"
```
ashley-hunter: Probably best to use `-1` as the value when making an item not-tabbable as a value of `null` will use the default tabindex for that element, so for example an `a` tag or `button` element would be tabbable with a value of `null`. So for the time being this would make it not tabbable, should we decide to change this to use a button element or something this issue would then reappear, so its just safer to use `-1` instead.

TaraTodd123: Updated to -1

ashley-hunter:
Looks good, the only query I have is in the original ticket it mentions the following:

> Clear button focus outline in MF theme is too large. (matches date range picker clear button focus though - RP)

Which currently looks like this:
image

Did we make a decision about whether or not this should be changed?

donna-marijne:
I think it's fine in Keppel but needs to be fixed in the Micro Focus theme.

TaraTodd123:
I changed the style for the Micro Focus theme clear button focus, the pull requests for this are also linked to the original Jira ticket.

src/components/tag-input/tag-input.component.html
```diff
@@ -38,12 +38,12 @@
 </ol>
 
 <!-- Insert the custom icon if provided -->
-<div class="ux-tag-icons" *ngIf="icon || _showClearButton">
+<div class="ux-tag-icons" [class.disabled]="disabled" *ngIf="icon || _showClearButton">
```
donna-marijne: The ux-tag-input element gets the class `disabled` when disabled, so I don't think you need to add another one in here.

TaraTodd123: Updated

ashley-hunter:
@TaraTodd123 it might be useful to setup [stylelint](https://www.jetbrains.com/help/webstorm/using-stylelint-code-quality-tool.html) in your editor, this will show you any style issues that will cause the build to fail.

donna-marijne:
Agreed, you can also do `grunt build` before pushing which will potentially save some time.

src/components/tag-input/tag-input.component.less
```diff
@@ -109,6 +109,12 @@ ux-tag-input {
 
     &.disabled {
         background-color: @tag-input-disabled-bg;
+
+        .ux-tag-icon,
+        .ux-select-icon {
+            cursor: default;
```
ashley-hunter: On multiple select I still see the pointer cursor when disabled?

TaraTodd123: That's it updated now

src/components/tag-input/tag-input.component.less
```diff
@@ -109,6 +109,11 @@ ux-tag-input {
 
     &.disabled {
         background-color: @tag-input-disabled-bg;
+
+        .ux-tag-icons .ux-tag-icon, .ux-tag-icons .ux-select-icon {
```
donna-marijne: We use Less and Sass to eliminate repetition, so we can write the above as:

````less
.ux-tag-icons {
    .ux-tag-icon,
    .ux-select-icon {
````

(We also tend to split comma-separated selectors with a new line as well, just to emphasize the different cases when reading the code.)

TaraTodd123: updated

PR #1161, opened 2019-08-06
Input dropdown component

tobias-schreiner:
#### Checklist

#### Ticket / Issue

The user wants to create a UI with a drop-down with customized content, which can be achieved with this new component.

Implemented the new custom drop-down component (ux-input-drowdown)

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/microfocusbbn_CI_UXAspects_master/


donna-marijne:
Hello Tobias,

Tony has granted you access to the SEPG Jenkins server, so please visit here:

https://jenkins.swinfra.net/job/SEPG/view/Templates/job/New%20SEPG%20Build/build?delay=0sec

Fill in the fields as required. In this case it would be:

github.com
microfocusbbn
UXAspects
master

Click Build. This will kick off a full build of your branch and run the automated tests. It takes nearly 30 minutes unfortunately. You'll find the job at:

https://jenkins.swinfra.net/job/SEPG/job/microfocusbbn/job/microfocusbbn~UXAspects~master~CI/

The Chrome/Firefox addon "Yet Another Jenkins Notifier" can track this job and notify you of successful or failed builds.

On a successful build, it will publish the documentation site including your changes at:

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/microfocusbbn_CI_UXAspects_master/

Note that you only have to create the build once per branch. After this, it will rebuild automatically when you push changes.

When you have got a clean build using SEPG, please update the PR template with the URL.

donna-marijne:
Secondly, the **Ticket / Issue** field in the PR template is blank. Please fill this in with a description of the component(s) and the use cases that they fulfil. E.g. "A user is trying to ___ which they can achieve with this component by ___." The use cases will form the basis of the end-to-end tests which will be written for the component.

ashley-hunter:
Hi Tobias, thanks very much for your contribution! I haven't had a chance to review this yet due to preparation for our upcoming release, however I did notice one thing that will need changed before we can merge this. 

I see there is a dependency on `ngx-bootstrap` which we are not able to have in our components as various teams use different versions of this library and there are API breakages across these versions.

Instead consider using our new `ux-menu` component instead, which is our replacement for the ngx-bootstrap dropdown component.

tobias-schreiner:
Hi Ashley,

> I see there is a dependency on `ngx-bootstrap` which we are not able to have in our components ...

I am using the same functionality as in your [split button](https://pages.github.houston.softwaregrp.net/sepg-docs-qa/microfocusbbn_CI_UXAspects_master/#/components/buttons#split-button-dropdowns), that's why I thought it was safe. Why can it be used there but not in my implementation?

Best regards,

Tobias

donna-marijne:
1. It is example code, which means that the consumer can amend it according to whatever version of bootstrap they are using in case of breaking API changes.
2. It's going away next sprint.

ashley-hunter:
Hi Tobias, the split button is a common pattern that we have documented, it is not an Angular component that we provide, simply an example of how it can be achieved using a third party component (which up until recently we did not provide an alternative). We have a ticket to update this example to use the `ux-menu` component in an upcoming sprint now that this component is available.

We just can't use it in our Angular components as this would force teams to include `ngx-bootstrap` in their applications, whereas currently it is completely optional, (if you search in our `src` folder you'll notice we don't use `ngx-bootstrap` in any of our components or directives).

It also means we would need to ensure applications used a specific version of `ngx-bootstrap` to ensure the API is as we expect, which is a problem if teams are currently using a newer or older version in their application. It also poses issues when we decide we want to update `ngx-bootstrap` as this could potentially cause a breaking change in applications that are using an older version.

tobias-schreiner:
Hi Donna, hi Ashley, thank you very much for your explanations! I will try to figure out how to implement the components without bootstrap.

ashley-hunter:
Thanks for changing to use the `ux-menu` component instead! I have pushed a few changes to include the latest changes in this sprint, as we upgraded to Angular 8 and a few additional changes were required. I also made a few changes to provide stricter type checking etc.. I will do a proper review of the changes later today.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.ts
```diff
@@ -0,0 +1,88 @@
...
+
+    constructor() {
+        super(require.context('./snippets/', false, /\.(html|css|js|ts)$/));
+    }
+
+    private index(text: string): number {
+        return text.toLowerCase().indexOf(this.filter.toLowerCase());
+    }
+
+    isHidden(name: string): boolean {
```
ashley-hunter: Instead of calling a function to check if an item is hidden, could we update the array whenever the filter text is changed? We try to avoid calling functions from a template as this cause significant slow down on larger pages as the function needs to be called every time change detection is run to check for a change, whereas updating a variable can be checked with a simple equality check.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
```diff
@@ -0,0 +1,156 @@
...
+
+                    <ux-radio-button
+                        *ngFor="let option of optionList; let index = index"
+                        name="group"
+                        [value]="selected"
+                        (valueChange)="selected = $event"
+                        [option]="option"
+                        [hidden]="isHidden(option.name)"
+                        [class.focused]="focusIndex === index">
+                        {{ beforeHighlight(option.name) }}
```
ashley-hunter: Instead of calling these three functions could we consider using a pipe as they are `pure` and avoid unnecessary change detection, as calling three functions for each item every time change detection runs could incur quite a performance hit on larger lists or heavy pages. We do something similar to this here: https://github.com/UXAspects/UXAspects/blob/develop-1.8.2/src/components/filters/filter-dynamic/filter-typeahead-highlight.pipe.ts

docs/app/pages/components/components-sections/select/select-custom/snippets/app.html
```diff
@@ -0,0 +1,45 @@
...
+                    <ux-radio-button
+                        *ngFor="let option of optionList; let index = index"
+                        name="group"
+                        [value]="selected"
+                        (valueChange)="selected = $event"
+                        [option]="option"
+                        [hidden]="isHidden(option.name)"
+                        [class.focused]="focusIndex === index">
+                        {{ beforeHighlight(option.name) }}
+                        <span class="highlight">{{ highlightText(option.name)}}</span>
```
ashley-hunter: I think the highlight color should be updated:

image

The blue color is not in our colour palette, instead consider something like this (see [Facet Typehead List](http://localhost:8080/#/components/facets#facet-typeahead-list)):

image

src/components/select-custom/select-custom.component.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    ngOnDestroy(): void {
+        this._onDestroy$.next();
+        this._onDestroy$.complete();
+    }
+
+    onToggle(): void {
+        setTimeout(
+            () => {
+                if (this._elementRef.nativeElement.classList.contains('open')) {
```
ashley-hunter: Is this going to have any effect? From what I recall the open class was added by the ngx-bootstrap dropdown but not by ux-menu, so perhaps this isn't needed? And if it is needed perhaps if it is needed we could just use the `opened` event triggered by the `ux-menu` component to avoid having to check whether it is opened or closed.

src/components/select-custom/select-custom.component.html
```diff
@@ -0,0 +1,51 @@
...
+
+    <ng-container [ngTemplateOutlet]="buttonContentRef"></ng-container>
+
+    <div class="ux-select-icons">
+        <i class="ux-select-icon ux-icon ux-icon-close"
+            *ngIf="allowNull && selected"
+            (click)="resetValue($event)">
+        </i>
+
+        <i class="ux-select-icon ux-icon ux-icon-chevron-down ux-icon-down"></i>
```
ashley-hunter: In the recent sprint we have went through our icon component to use instead of icon classes if they could be used here instead:

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_develop-1.8.2/#/components/icons#icon

src/components/select-custom/select-custom.component.html
```diff
@@ -0,0 +1,51 @@
...
+    <div [style.max-height.px]="maxHeight"
+         [style.width.px]="width">
+
+        <div [hidden]="busy">
+            <div [hidden]="hideFilter">
+
+                <input #filterInput
+                        type="text"
+                        [placeholder]="placeholder"
+                        class="filter-input"
```
ashley-hunter: We might be able to use our `form-control` class on this element as well, which might avoid a lot of the custom styles you had to add and give you the correct focus styling etc.. and hopefully will allow it to pick up a lot of the micro focus styles for inputs when we add this to the MF site.

donna-marijne:
Before we get too deep into documentation etc, can we rethink the "Select Custom" name? Basically every component in UX Aspects has a degree of customization, so to someone browsing the collection this name says very little.

Perhaps we can work on the use cases a little and this will help to narrow down an informative name? Currently we have "The user wants to create a UI with a drop-down with customized content, which can be achieved with this new component." Unfortunately this isn't much of a use case, because the end user does not want to create a UI, they want to accomplish tasks, and the UI serves that goal. So let's break it down a bit - what are your users currently doing with this component?

It's possible that "Select Custom" really is the best we can do, but I'd like to make an attempt to find something more unique and informative.

tobias-schreiner:
Unfortunately, I misunderstood what you meant by user (user of the component = developer vs. user of the end-product). The idea was that we have a lot of drop-down components and we wanted to have an implementation for the common parts. So to say, this is only half a component and the other half comes with the content (the two ng-templates it requires).

We currently use it in two cases (the radio-button list is just a simple, but silly show case):

- Select from a tree grid

grafik

- Select from stacked panels

grafik

And after clicking on the first entry:

grafik

Does this clarify it? I still do not know how to phrase it, though.

I am open to renaming the component. My idea was that there is already a ux-select-list, so it should be a ux-select-something. If you can help me find a better word (not only as shared components experts, but also as native speakers), I would appreciate it.

src/components/select-custom/select-custom.component.ts
```diff
@@ -0,0 +1,128 @@
...
+    /** Define the placeholder for the filter input */
+    @Input() placeholder: string = 'Type to filter...';
+
+    /** Emit when the selected item is changed */
+    @Output() selectedChange = new EventEmitter<T>();
+
+    /** Emit when the filter text is changed */
+    @Output() filterChange = new EventEmitter<string>();
+
+    @ContentChild('dropdownContent', { static: false }) dropdownContentRef: TemplateRef<void>;
```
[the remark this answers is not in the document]
tobias-schreiner: We have two templates: one might usually be bigger, but the other one is equally important. They work side-by-side and their implementations need to be connected. So they have an equal weight in the implementation. Putting both inside ng-templates stresses this "equality".

In addition, I like to put them both inside the component, so that it is clear where they belong to and are not kind of "floating" (even if that might not hinder implementing your proposal).

That's why I would like to keep it as it is.

src/components/select-custom/select-custom.component.ts
```diff
@@ -0,0 +1,128 @@
...
+    @ContentChild('buttonContent', { static: false }) buttonContentRef: TemplateRef<void>;
```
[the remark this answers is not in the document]
tobias-schreiner: I see. I am fine with changing it. How about "displayContent"? Or "toggleContent" / "dropdownToggleContent"?

docs/app/pages/components/components-sections/select/select-custom/snippets/app.html
```diff
@@ -0,0 +1,45 @@
...
+            [maxHeight]="componentMaxHeight"
+            [placeholder]="placeholder">
+
+            <ng-template #buttonContent>
+                <b>Selection:</b> {{ selected ? selected.name : '(none)' }}
+            </ng-template>
+
+            <ng-template #dropdownContent>
+
+                <div role="radiogroup"
```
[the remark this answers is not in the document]
tobias-schreiner: I'm a little torn here. The tabbable list will make the example easier, but the radio button list has the advantage that it shows all the implementation work that the user of the component (i.e. developer who incorporates the component) will have to consider. (If they insert a component that handles everything already, fine, but this example shows the "worst case".)

donna-marijne:
Thank you, that does clarify things. I'll get Roland to take a look at this and see if we can come up with some suggestions.

docs/app/pages/components/components-sections/select/select-custom/snippets/app.html
```diff
@@ -0,0 +1,45 @@
...
+                        class="radio-button-container"
+                        tabindex="0"
+                        (keydown.space)="selectButton($event)"
+                        (keydown.up)="navigateUp($event)"
+                        (keydown.arrowUp)="navigateUp($event)"
+                        (keydown.down)="navigateDown($event)"
+                        (keydown.arrowDown)="navigateDown($event)">
+
+                    <ux-radio-button
```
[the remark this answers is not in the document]
tobias-schreiner: The search icon should be in the search input. That is a style issue with Chrome (I work with Firefox) and I fixed it (wait for my next commit).

The purpose of the icon is to show that the input field is for search. It is required in all our mocks, so I would like to keep it. Do you have search fields without this icon?

I fixed the background color. I wanted to choose a different color to show that it is custom style, but I think your proposal is better.

The focus indicator on the container was only shown in Chrome again. I removed it.

src/components/select-custom/select-custom.component.html
```diff
@@ -0,0 +1,51 @@
...
+    </div>
+
+</button>
+
+<ux-menu #menu menuClass="select-menu">
+
+    <div [style.max-height.px]="maxHeight"
+         [style.width.px]="width">
+
+        <div [hidden]="busy">
```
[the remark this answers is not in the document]
tobias-schreiner: I like to use [hidden] since it has the advantage that it renders the hidden components already, which improves performance. Anyway, I changed it.

docs/app/pages/components/components-sections/select/select-custom/snippets/app.css
```diff
@@ -0,0 +1,25 @@
+.radio-button-container {
+    padding-top: 16px;
+    padding-bottom: 16px;
+}
+
+ux-radio-button {
+    padding-top: 4px;
```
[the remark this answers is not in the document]
tobias-schreiner: I changed it. I somehow thought it said different in your style guide, but I was mistaken.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.ts
tobias-schreiner: That would mean to maintain two arrays, that's why I preferred to use the function. And especially since this is demo code, I wanted to keep it simple. (By the way, I would need to check if it is contained in the array, not equality.) Therefore, I would prefer to leave it as it is, but I do not have a strong opinion. Does this cause performance issues on the demo page?

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
tobias-schreiner: I have the same (not-so-strong) opinion as before: This is demo code. I would prefer making it easy to understand over showing the consumer of the component how to implement optimal performance. If you still think it needs to be changed, I will do it.

docs/app/pages/components/components-sections/select/select-custom/snippets/app.html
tobias-schreiner: The space around the highlight had already been fixed in my previous commit. I fixed the color.

src/components/select-custom/select-custom.component.ts
tobias-schreiner: I overlooked it when I removed the bootstrap code. I had already fixed this in my previous commit (exactly in the way you described).

src/components/select-custom/select-custom.component.html
tobias-schreiner: I fixed that. (To be honest, I had copied it from your select implementation.)

donna-marijne:
I see that we are getting quite deep into development with this, but there are still some fundamental issues regarding the scope of the component and its use cases. If possible, I would suggest pausing on implementation until we have laid out some solid use cases.

For example, the hierarchy/tree demos that you have shown are very promising, and this would be a compelling component to provide in my opinion. However, the example with radio buttons is rather less exciting.

If the hierarchy-style select is the main/only use case here, then it may be that we want to provide that as a drop-in component with panel/tree modes, instead of something that is completely generic.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
ashley-hunter: I do agree that it is a bit simpler, but I have found that very often people just copy the code from our examples verbatim, so if we do something like this they end up with slow pages once they have a page with 100s of items etc... So we tend to document with the best practices, that way people who blindly copy the code at least will have fast code

donna-marijne: Agree with this, sample code should always adhere to best practices.

tobias-schreiner:
In fact, for these two use cases, we have derived shared components already, so we could contribute these palatable components as well (not now, since we have enhanced your tree grid and the stacked panels would need to be contributed first).

I forgot to show a third usage, which we implemented directly with the custom drop-down and not created a separate component for:

grafik

This is something where the component came in handy, and there is probably a variety of possible usages.

There were also some more ideas for these drop-downs, but they only exist as mocks so far:

https://t0h7na.axshare.com/#g=1&p=policy_variable_picker_playground - Password: 0psbr!dge

I will commit the changes I have made so far (see my comments above) and then wait for the discussion.

donna-marijne:
> In fact, for these two use cases, we have derived shared components already, so we could contribute these palatable components as well (not now, since we have enhanced your tree grid and the stacked panels would need to be contributed first).

That would be great!

One possibility is that we use the generic component internally as a basis for this and other specialized select-like components that site authors can drop into their pages quite easily.

Thanks for your continued input.

src/components/select-custom/select-custom.component.ts
```diff
@@ -0,0 +1,128 @@
...
+    @Input() selected: T;
+
+    /** Hide the filter input */
+    @Input() hideFilter: boolean;
+
+    /** Debounce the filter change event emitter */
+    @Input() filterChangeDebounce: number = 500;
+
+    /** Define the width of the component */
+    @Input() width: number;
```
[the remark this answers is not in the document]
tobias-schreiner: The button and the drop-down have to have the same width and I am not sure how to achieve that without explicitly setting the width on both components. If you have an idea, please let me know.

src/components/select-custom/select-custom.component.ts
```diff
@@ -0,0 +1,128 @@
...
+        }
+    ],
+    host: {
+        '[style.width.px]': 'width'
+    }
+})
+export class SelectCustomComponent<T> implements ControlValueAccessor, OnChanges, OnInit, OnDestroy {
+
+    /** Define the selected item */
+    @Input() selected: T;
```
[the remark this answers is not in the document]
tobias-schreiner: This is due to the "base" / "custom" nature of this component: The selection is always performed in the contained component. We include these components in reactive forms and therefore, they need to implement ControlValueAccessor and provide the selected element.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.ts
ashley-hunter: Yes, it would mean having two arrays, an original list and a visible list. You wouldn't have to check in the view if an item is hidden, You would use a `ReadonlyArray` for the visible items which you update when the query changes. NgFor would bind directly to the list of visible items and handle the visibility for you. Its a case of it might not cause issues on our demo page, but in a page in an enterprise app a simple change like this can make quite a significant improvement in performance which is why we try to avoid function calls in template attributes/inputs wherever possible.

docs/app/pages/components/components-sections/select/select-custom/snippets/app.html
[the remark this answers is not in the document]
ashley-hunter: I think if we are sticking to the example of radio buttons it might be better to use the tabbable list, as we would recommend the use of it for simple lists like this. It would also implement some of the additional accessibility features that the example is currently missing for you, such as only one item in the list should be tabbable at a time and if I click on an item it should be become the active tabbable item etc..

I agree that if someone was going to implement a more complicated component they may have to do this themselves, but we don't want to suggest that for a list like this someone should be implementing this manually when we have directives there for them to do this already.

Our demos are always limited in what we can show and will usually be much simpler than a real life example, but we aim to show the best practices for the example we are using which would be to use our well established and well tested directives.

e2e/tests/components/page-header/page-header.e2e-spec.ts
```diff
@@ -1,7 +1,7 @@
 // import { ElementArrayFinder, ElementFinder } from 'protractor';
 import { PageHeaderPage } from './page-header.po.spec';
 
-describe('Page Header Tests', () => {
+xdescribe('Page Header Tests', () => {
```
ashley-hunter: Think this might be a typo?

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
ashley-hunter: You could add a reference to the button and then set the width of the dropdown based on it, eg:

```html
<button #button>

<ux-menu [style.width.px]="button.offsetWidth">
```

Or something along similar to that

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
tobias-schreiner: Thank you for your feedback! I have implemented the pipe now.

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
donna-marijne: May want to compare with the way the [Date/Time Picker](https://pages.github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/#/ui-components/input-controls#date-time-picker) works.

The contents of the popover in this case does not need to be a form control. We simply bind the output `date` to the `ngModel` of the input field, which is this case is already usable in a form, but the same principle can apply to a custom ControlValueAccessor component.

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
donna-marijne: Check out how the `ux-select` component does it. Both input and dropdown are children of a common parent container and have `width: 100%`. The key to this is setting `position: relative` on the parent, so that it counts as the nearest positioned ancestor of both children.

image

src/components/select-custom/select-custom.component.ts
```diff
@@ -0,0 +1,128 @@
...
+
+    /** Hide the filter input */
+    @Input() hideFilter: boolean;
+
+    /** Debounce the filter change event emitter */
+    @Input() filterChangeDebounce: number = 500;
```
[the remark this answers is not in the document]
tobias-schreiner: I will change that in my next commit.

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
tobias-schreiner: But that means that whatever is in the drop-down content, needs to be wrapped a separate component, so no more chance for simple implementations.

Our source CI picker looks like this at the moment:

```
<opr-ngx-select-custom formControlName="sourceCi">

  <ng-template #buttonContent>
    <span *ngIf="sourceCIControl.value && (sourceCIControl.value.name.length > 0)">
      {{ sourceCIControl.value.label }}
      <span class="ci-name">({{ sourceCIControl.value.ciTypeLabel }})</span>
    </span>
  </ng-template>

  <ng-template #dropdownContent>
    <ul class="ci-list">
      <li *ngFor="let ci of sourceCIList"
          (click)="editorForm.patchValue({sourceCi: ci})"
          [class.selected]="selected(ci,sourceCIControl.value)">
        {{ ci.ciTypeLabel }} <span class="ci-name">({{ ci.label }})</span>
      </li>
    </ul>
  </ng-template>

</opr-ngx-select-custom>
```

Such a compact implementation will then not be possible anymore.

e2e/tests/components/page-header/page-header.e2e-spec.ts
tobias-schreiner: No, this is not a typo. xdescribe / xit have the effect that the tests are suppressed. There is also fdescribe / fit (f = focus) to run only this one test and skip everything else.

I commented these test out after Donna told me there were issues and I wanted my build to pass.

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
tobias-schreiner: Thank you for your suggestions! I implemented the first one, the second I tried yesterday - however, ux-menu does not put button and drop-down in the same container.

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
ashley-hunter: The example in our site does not necessarily need to have a separate component, but a separate component is a valid use case that we need to consider, in which case we may not be able to simply do a (click) event to patch the value.

The confusing thing is with the code above, the custom select is bound to a form control but doesn't actually do the updating. The list below is still the thing updating the form control value. It's not a pattern have I really seen before where the `formControlName` element is not the thing updating the form?

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
ashley-hunter: @donna-marijne Any suggestions or preference?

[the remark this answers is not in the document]
donna-marijne: I'd be in favor of making this the default content template, since it's the main view of the component.

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
donna-marijne: The above example is a little strange. I don't think binding between parent and child should be done via FormControl.

Referring back to the date/time picker example, it would look something like this. 

````html
<opr-ngx-select-custom [(value)]="selectCustomValue" formControlName="sourceCi">
  <ng-template #dropdownContent>
    <ux-number-picker [(value)]="selectCustomValue"></ux-number-picker>
  </ng-template>
</opr-ngx-select-custom>
````

It would make no difference whether a custom component is used or not.

**Comment by donna-marijne** [comment]
This PR was overlooked when retiring the 1.8.2 branch. It should be released in 1.8.3, so the merge target is `develop-1.8.3`.

File: src/components/select-custom/select-custom.component.ts (hunk 14 above)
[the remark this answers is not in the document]
Comment (tobias-schreiner) [reply src/components/select-custom/select-custom.component.ts]: I am not sure I get your point. Which change to the component are you proposing? Or does each of you propose a different solution?

File: src/components/select-custom/select-custom.component.ts (hunk 8 above)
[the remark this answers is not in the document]
Comment (tobias-schreiner) [reply src/components/select-custom/select-custom.component.ts]: I would like to keep the two ng-templates (see my reasoning discussion above).

File: src/components/select-custom/select-custom.component.ts (hunk 14 above)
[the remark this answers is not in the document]
Comment (donna-marijne) [reply src/components/select-custom/select-custom.component.ts]: I'm proposing that you don't misuse the reactive forms feature to sync the value between child and parent. I have already pointed you to the date & time picker example, as well as including a pseudocode example, showing how it should be structured. Unless there are good reasons not to follow this model, let's go with it.

File: src/components/select-custom/select-custom.component.ts (hunk 8 above)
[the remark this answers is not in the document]
Comment (donna-marijne) [reply src/components/select-custom/select-custom.component.ts]: The reason it's hard to name is because it's the main presentation of the component, hence my suggestion.

File: docs/app/pages/components/components-sections/select/select-custom/select-custom.component.ts (hunk 1 above)
Comment (tobias-schreiner) [reply docs/app/pages/components/components-sections/select/select-custom/select-custom.component.ts]: I changed an committed it.

File: src/components/select-custom/select-custom.component.ts (hunk 14 above)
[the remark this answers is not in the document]
Comment (tobias-schreiner) [reply src/components/select-custom/select-custom.component.ts]: I was asking mainly because your and Ashley's proposals seemed to contradict each other and I wanted to come to a common solution.

About your last proposal: it looks very much like my current implementation, except that you call it value and I call it selected. Is it only about renaming the attributes? Am I missing anything?

File: docs/app/pages/components/components-sections/select/select-custom/snippets/app.html (hunk 9 above)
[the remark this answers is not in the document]
Comment (tobias-schreiner) [reply docs/app/pages/components/components-sections/select/select-custom/snippets/app.html]: I have changed and committed this.

File: src/components/select-custom/select-custom.component.html (hunk 6 above)
Comment (tobias-schreiner) [reply src/components/select-custom/select-custom.component.html]: Changed and committed.

File: src/components/select-custom/select-custom.component.ts (hunk 8 above)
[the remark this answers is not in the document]
Comment (tobias-schreiner) [reply src/components/select-custom/select-custom.component.ts]: I do not think it is hard to name - "displayContent" sounds fine for me. Since the first name was obviously confusing, I wanted to double-check and asked for feedback.

File: src/components/select-custom/select-custom.component.ts (hunk 14 above)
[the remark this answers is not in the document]
Comment (donna-marijne) [reply src/components/select-custom/select-custom.component.ts]: It was a response to this: https://github.com/UXAspects/UXAspects/pull/1161#discussion_r312448433, and the ensuing discussion about whether the inner component should be the form control.

File: src/components/select-custom/select-custom.component.ts (hunk 8 above)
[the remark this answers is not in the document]
Comment (donna-marijne) [reply src/components/select-custom/select-custom.component.ts]: I think that this should be the default template, and not require a name.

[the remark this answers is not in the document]
Comment (donna-marijne) [reply src/components/select-custom/select-custom.component.ts]: I didn't really see the technical reason why the component shouldn't have a default template, but perhaps I missed something?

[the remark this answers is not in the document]
Comment (tobias-schreiner) [reply src/components/select-custom/select-custom.component.ts]: I think technically, both solutions are possible. I also don't see a technical reason for not staying with two named templates.

It is more a logical reason. From the discussion above:

> We have two templates: one might usually be bigger, but the other one is equally important. They work side-by-side and their implementations need to be connected. So they have an equal weight in the implementation. Putting both inside ng-templates stresses this "equality".

I do not have such a strong opinion, but since Ashley prefers to make the other template the default template, I am not sure how we proceed here.

File: src/components/select-custom/select-custom.component.ts (hunk 14 above)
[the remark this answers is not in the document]
Comment (tobias-schreiner) [reply src/components/select-custom/select-custom.component.ts]: So is there still a change requested regarding this topic, and which one, or can everything stay as it is?

File: src/components/select-custom/select-custom.component.ts (hunk 8 above)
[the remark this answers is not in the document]
Comment (ashley-hunter) [reply src/components/select-custom/select-custom.component.ts]: I do agree with Donna, that we do usually try to provide a sensible default template, and if this is something that might need to be customized eg, for localization purposes, then we can allow an `ng-template` to customize it but not require it, so having a sensible default makes it easiest for consumers to start using the component. This is the way our other components tend to work.

In regards to using `ng-content` there is a technical reason why we should choose `ng-content` over `ng-template` for the dropdown content. We normally use `ng-template`s to customize a part of a component (such as the button content), but for content that is essentially a part of the page component it is on, `ng-content` should be used. This is the case for this component, where the contents in the dropdown may bind to a `FormGroup` or be using `ngModel`, binding to values on the page it is on, not values related to the select component.

This is important because when change detection is run will differ. `ng-content` runs change detection when the page component changes, whereas `ng-template` runs change detection when the `select-custom` component updates. We can see the difference if we look at the Angular compiler output (Showing Ivy output as it is easier to read):

```html
<ux-select-custom>
  {{ text }}

  <ng-template #customTemplate>
    {{ text }}
  </ng-template>
</ux-select-custom>
```

Would compile to:

#### App Component
```typescript
template: function AppComponent_Template(rf, ctx) {
    if (rf & 1) { // create step
        i0.ɵɵelementStart(0, "ux-select-custom");
        i0.ɵɵtext(1); // ng-content inlined and part of app component
        i0.ɵɵtemplate(2, AppComponent_ng_template_2_Template, 1, 1, "ng-template", null, _c0, i0.ɵɵtemplateRefExtractor);
        i0.ɵɵelementEnd();
    }
    if (rf & 2) { // update step - run when changes occur in app component
        i0.ɵɵselect(1);
        i0.ɵɵtextInterpolate1(" ", ctx.text, " "); // update ng-content if a change occurs
    }
},
```

#### Select Component

```typescript
template: function ChildComponent_Template(rf, ctx) {
    if (rf & 1) { // create step
        i0.ɵɵprojectionDef();
        i0.ɵɵprojection(0);
        i0.ɵɵelementContainerStart(1, _c1);
        i0.ɵɵelementContainerEnd();
    }
    if (rf & 2) { // update step - run when changes occur in select component
        i0.ɵɵselect(1);
        i0.ɵɵproperty("ngTemplateOutlet", ctx.template); // template updated when changes occur in this component
    }
},
```

I hope this makes sense.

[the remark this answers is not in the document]
donna-marijne: Thank you very much for the clarification and examples, this helps a lot. I do think that binding dropdown content to the page's FormGroup is unlikely in practice, but there was an example with number pickers where one would probably bind the `ngModel`s to the page.

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
donna-marijne: I believe that the `ng-content` recommendation by Ashley will take care of all of the above concerns.

docs/app/pages/components/components-sections/select/select-custom/snippets/app.html
```diff
@@ -0,0 +1,76 @@
...
+                        </span>
+                    </ux-radio-button>
+                </div>
+            </ng-template>
+
+        </ux-select-custom>
+    </div>
+</div>
+
+<div class="row uxd-customize-example">
```
donna-marijne: Could you remove this section from the code snippet? It adds a lot of code to scroll through that is not relevant to the example. Also in plunker you can change the property values directly.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
```diff
@@ -0,0 +1,134 @@
...
+                    <div class="col-md-2 col-sm-12">
+                        <label>placeholder</label>
+                    </div>
+                    <div class="col-md-2 col-sm-12">
+                        <input class="form-control" [(ngModel)]="placeholder">
+                    </div>
+                </div>
+                <div class="row uxd-customize-row">
+                    <div class="col-md-4 col-sm-12">
+                        <button class="btn button-primary" (click)="selected = filteredOptionList[0]">Select first element</button>
```
donna-marijne: This does not belong in the customize section. Could you pull it out and place it alongside the component itself?

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
```diff
@@ -0,0 +1,134 @@
...
+                </div>
+            </ux-accordion-panel>
+        </ux-accordion>
+    </div>
+</div>
+
+<hr>
+
+<p>
+    The <code>select-custom</code> component is a drop-down with flexible content.
```
donna-marijne: `ux-input-dropdown` was the agreed name.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
```diff
@@ -0,0 +1,134 @@
...
+
+    <tr uxd-api-property name="selected" type="T">
+        If specified, the items provided will be selected.
+        To programmatically modify the selection, assign a new array object to this property.
+    </tr>
+    <tr uxd-api-property name="hideFilter" type="boolean" defaultValue="false">
+        If <code>true</code>, the filter field is not displayed.
+    </tr>
+    <tr uxd-api-property name="maxHeight" type="number">
+        Maximum height of the drop-down.
```
donna-marijne: Either specify "... in pixels" here, or change the input to accept CSS units as a string, e.g. `100px`.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
```diff
@@ -0,0 +1,134 @@
...
+        <b>Note:</b> The default value is not localized!
+    </tr>
+</uxd-api-properties>
+
+<uxd-api-properties tableTitle="Outputs">
+    <tr uxd-api-property name="filterChange" type="string">
+        Fires whenever the filter term changes, sends new filter string as event.
+    </tr>
+    <tr uxd-api-property name="selectedChange" type="T">
+        This event will emit any time the selection changes. The event will contain the new selected items.
```
donna-marijne: Above description begins "Fires..."; this begins "This event will emit...". I recommend changing both to "Emits when..." for consistency.

docs/app/pages/components/components-sections/select/select-custom/snippets/app.html
```diff
@@ -0,0 +1,76 @@
...
+                <div class="row uxd-customize-row">
+                    <div class="col-md-4 col-sm-12">
+                        <ux-checkbox [(value)]="showBusyIndicator">busy</ux-checkbox>
+                    </div>
+                </div>
+                <div class="row uxd-customize-row">
+                    <div class="col-md-2 col-sm-12">
+                        <label>placeholder</label>
+                    </div>
+                    <div class="col-md-2 col-sm-12">
```
donna-marijne: I think this one needs to be a little wider.

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
tobias-schreiner: After our discussion from yesterday, I changed and committed it.

src/components/select-custom/select-custom.component.ts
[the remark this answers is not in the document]
tobias-schreiner: I changed the name to displayContent and created a default template for it.

docs/app/pages/components/components-sections/select/select-custom/snippets/app.html
tobias-schreiner: Done.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
tobias-schreiner: Done.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
tobias-schreiner: Changed with the renaming.

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
tobias-schreiner: I did the first

docs/app/pages/components/components-sections/select/select-custom/select-custom.component.html
tobias-schreiner: Done.

docs/app/pages/components/components-sections/select/select-custom/snippets/app.html
tobias-schreiner: Done.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    ngOnInit(): void {
+        this.filterList();
+    }
+
+    private index(text: string): number {
+        return text.toLowerCase().indexOf(this.filter.toLowerCase());
+    }
+
+    isHidden(name: string): boolean {
```
ashley-hunter: Can this be removed now?

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
```diff
@@ -0,0 +1,94 @@
...
+        return this.filter && (this.filter.length > 0) && (this.index(name) === -1);
+    }
+    
+    selectButton(event: KeyboardEvent, option: RadioOption): void {
```
ashley-hunter: Should this still be called `button` or should this now be `input`?

src/components/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,49 @@
...
+                    [placeholder]="placeholder"
+                    class="form-control"
+                    [(ngModel)]="filterText"
+                    (input)="filterChange.emit(filterText)"
+                    (click)="$event.stopPropagation()" />
+
+            <button type="button"
+                    class="btn btn-flat btn-icon-only filter-button"
+                    (click)="resetFilter($event)"
+                    [tabindex]="filterText.length > 0 ? 0 : 1">
```
ashley-hunter: We try to avoid positive tabindexes. If you want to prevent tabbing change it to `-1` instead of `1`

src/components/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,49 @@
...
+                 class="ux-select-icon ux-select-chevron-icon"
+                 (click)="menuTrigger.toggleMenu(); $event.stopPropagation()">
+        </ux-icon>
+    </div>
+</div>
+
+<ux-menu #menu menuClass="select-menu"
+         (opened)="filterInputElement.nativeElement.focus()">
+
+    <div [style.max-height.px]="maxHeight"
```
ashley-hunter: Instead of forcing this to be a pixel value, we should allow any CSS value here. We can use the setter and the Angular CDK coercion utilities to default to pixels if no unit is provided (https://github.com/angular/components/blob/6.0.x/src/cdk/coercion/css-pixel-value.ts)

Something like this:
```typescript
set maxHeight(value: string | any) {
    this._maxHeight = coerceCssPixelValue(value);
}
```

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
```diff
@@ -0,0 +1,94 @@
...
+export class ComponentsInputDropdownComponent extends BaseDocumentationSection implements IPlaygroundProvider, OnInit {
+
+    selected: RadioOption;
+    optionList: ReadonlyArray<RadioOption> = [
+        { name: 'One' }, { name: 'Two' }, { name: 'Three' }, { name: 'Four' }
+    ];
+    filteredOptionList: ReadonlyArray<RadioOption> = this.optionList;
+    filter: string = '';
+    allowNull: boolean = false;
+    componentMaxHeight: number = 400;
```
ashley-hunter: Perhaps rename this to just `maxHeight` for consistency as all other variables are the same as the Input property

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
```diff
@@ -0,0 +1,94 @@
...
+            }
+        ]
+    };
+
+    constructor() {
+        super(require.context('./snippets/', false, /\.(html|css|js|ts)$/));
+    }
+
+    ngOnInit(): void {
+        this.filterList();
```
ashley-hunter: Is this needed? The list already gets set with the initial items above?

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,119 @@
+<div class="row">
+    <div class="col-md-5">
+
+        <ux-input-dropdown
```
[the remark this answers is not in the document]
tobias-schreiner: I moved the button down, so that it is not covered by the drop-down.

The gap between display and drop-down comes from the select component. I can remove it, if you wish, but that would make these two components inconsistent.

In the neutral style, I don't care, but in the Micro Focus style, these need to be consistent to comply with the requirements from our UX team.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,119 @@
...
+                <b>Selection:</b> {{ selected ? selected.name : '(none)' }}
+            </ng-template>
+
+            <div class="radio-button-container"
+                 uxTabbableList>
+                <ux-radio-button
+                    uxTabbableListItem
+                    *ngFor="let option of filteredOptionList"
+                    name="group"
+                    [value]="selected"
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
```diff
@@ -0,0 +1,94 @@
...
+    isHidden(name: string): boolean {
+        return this.filter && (this.filter.length > 0) && (this.index(name) === -1);
+    }
+    
+    selectButton(event: KeyboardEvent, option: RadioOption): void {
+        this.selected = option;
+        event.preventDefault();
+    }
+
+    setFilter($event: string) {
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

src/components/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,49 @@
+<div class="ux-select-container">
+    <button #button type="button" class="form-control"
+            [uxMenuTriggerFor]="menu">
+        <ng-template #defaultDisplayContent>{{selectedString()}}</ng-template>
+        <ng-container [ngTemplateOutlet]="displayContentRef ? displayContentRef : defaultDisplayContent"></ng-container>
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

src/components/input-dropdown/input-dropdown.component.ts
```diff
@@ -0,0 +1,114 @@
...
+    @ViewChild('filterInput', { static: false }) filterInputElement: ElementRef;
+
+    filterText: string = '';
+
+    onChange: (_: T) => void = () => { };
+    onTouched: () => void = () => { };
+
+    private readonly _onDestroy$ = new Subject<void>();
+
+    constructor() { }
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

src/components/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,49 @@
+<div class="ux-select-container">
+    <button #button type="button" class="form-control"
+            [uxMenuTriggerFor]="menu">
+        <ng-template #defaultDisplayContent>{{selectedString()}}</ng-template>
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

src/components/input-dropdown/input-dropdown.component.less
```diff
@@ -0,0 +1,30 @@
...
+}
+
+.select-menu {
+    position: relative;
+    overflow-y: auto;
+
+    .filter-container {
+        position: relative;
+
+        .filter-button {
```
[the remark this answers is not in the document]
tobias-schreiner: I adapted the padding.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,119 @@
...
+
+<hr>
+
+<p>
+    The <code>input-dropdown</code> component is a drop-down with flexible content.
+    The following can be used to customize and interact with the component:
+</p>
+
+<uxd-api-properties tableTitle="Inputs">
+    <tr uxd-api-property name="<HTML content>" type="<HTML content>">
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
tobias-schreiner: Indeed.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
tobias-schreiner: I made it selectOption since it sets one of the radio options.

src/components/input-dropdown/input-dropdown.component.html
tobias-schreiner: That was a typo. It is fixed.

src/components/input-dropdown/input-dropdown.component.html
tobias-schreiner: I changed that. Setting a pure number does not work on the demo page, though, but when I set it manually.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
tobias-schreiner: Done.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
tobias-schreiner: On the demo page, it is not needed, but in case you passed a non-empty filter into the component. I am not sure if this hint for the consumers of the component is necessary.

What do you think?

ashley-hunter: I think it's probably best if we only include the code that is needed for the example to run. It's probably best to keep the example as simple as possible, as previously mentioned, people tend to just copy and paste code as is, so its probably best not to include code that isn't directly required and let them enhance the sample code as they need.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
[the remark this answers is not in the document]
ashley-hunter: Ah yes, you are correct, this is consistent with the select component, I think it was just the button behind it that made it look different. Thanks for moving the button, ignore the spacing issue.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.ts
tobias-schreiner: OK, I removed it.

src/components/input-dropdown/input-dropdown.component.ts
```diff
@@ -33,7 +33,7 @@ import { coerceCssPixelValue } from '@angular/cdk/coercion';
     ]
 })
 export class InputDropdownComponent<T> implements ControlValueAccessor, OnChanges, OnDestroy {
-    private _maxHeight: string;
+    maxHeightString: string;
```
ashley-hunter: Could we keep this as `_maxHeight` we have a convention that variables that should be `private` but can't be because they are used in the template should still have the underscore prefix.

This is because if someone accesses the component via ViewChild for example, auto complete will suggest there is a property `maxHeightString` that they can access which we don't really want them to use as it is internal and they should be interacting with the input `maxHeight` instead, so the underscore prefix helps indicate that this value is still technically private.

I also think the description needs updated as this now accepts a string or number and can use any CSS unit.
image

tobias-schreiner: I changed it.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -83,7 +83,7 @@
         If <code>true</code>, the filter field is not displayed.
     </tr>
     <tr uxd-api-property name="maxHeight" type="number">
```
ashley-hunter: Should `type` now be updated here to be `string | number`?

src/components/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,48 @@
...
+            <input #filterInput
+                    type="text"
+                    [placeholder]="placeholder"
+                    class="form-control"
+                    [(ngModel)]="filterText"
+                    (input)="filterChange.emit(filterText)"
+                    (click)="$event.stopPropagation()" />
+
+            <button type="button"
+                    class="btn btn-flat btn-icon-only filter-button"
```
donna-marijne: There is no `btn-icon-only` class. `btn-icon` perhaps? https://uxaspects.github.io/UXAspects/#/css/buttons#icon-buttons-documentation-component

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,116 @@
...
+                </div>
+            </ux-accordion-panel>
+        </ux-accordion>
+    </div>
+</div>
+
+<hr>
+
+<p>
+    The <code>input-dropdown</code> component is a drop-down with flexible content.
```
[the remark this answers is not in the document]
tobias-schreiner: Fixed. Also the two above.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,116 @@
...
+    The content of the drop-down can be transcluded, the display of the closed
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,116 @@
...
+    component should be inserted into a ng-template with name "displayContent".
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,116 @@
...
+    </div>
+</div>
+
+<hr>
+
+<p>
+    The <code>input-dropdown</code> component is a drop-down with flexible content.
+    The content of the drop-down can be transcluded, the display of the closed
+    component should be inserted into a ng-template with name "displayContent". 
+    The following can be used to customize and interact with the component:
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,116 @@
...
+        To programmatically modify the selection, assign a new array object to this property.
+    </tr>
+    <tr uxd-api-property name="hideFilter" type="boolean" defaultValue="false">
+        If <code>true</code>, the filter field is not displayed.
+    </tr>
+    <tr uxd-api-property name="maxHeight" type="number">
+        Maximum height of the drop-down in pixels.
+    </tr>
+    <tr uxd-api-property name="placeholder" type="string" defaultValue="Type to filter...">
+        Placeholder to display inside the filter field.<br>
```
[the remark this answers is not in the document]
tobias-schreiner: I removed the line break.

src/components/input-dropdown/input-dropdown.component.html
tobias-schreiner: Removed.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
tobias-schreiner: Done.

src/components/input-dropdown/input-dropdown.component.html
```diff
@@ -32,10 +34,11 @@
                     class="form-control"
                     [(ngModel)]="filterText"
                     (input)="filterChange.emit(filterText)"
-                    (click)="$event.stopPropagation()" />
+                    (click)="$event.stopPropagation()"
+                    aria-label="Filter"/>
```
ashley-hunter: This would prevent the aria label from being localized. Might I recommend the approach Angular Material uses for their [select](https://github.com/angular/components/blob/master/src/material/select/select.ts#L421) component adding an input on the component, eg:

```typescript
 /** Aria label of the input. If not specified, the placeholder will be used as label. */
@Input('aria-label') ariaLabel: string = '';
```

And by default in the HTML use the placeholder if no aria label has been specified:

```html
[attr.aria-label]="ariaLabel || placeholder"
```

tobias-schreiner: Very good. I changed it.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -0,0 +1,123 @@
...
+    <tr uxd-api-property name="aria-label" type="string">
+        Aria label of the filter field. If not specified, the placeholder will be used.
+    </tr>
+</uxd-api-properties>
+
+<uxd-api-properties tableTitle="Outputs">
+    <tr uxd-api-property name="filterChange" type="string">
+        Emits when the filter term changes. The event will contain the new filter string.
+    </tr>
+    <tr uxd-api-property name="selectedChange" type="T">
```
ashley-hunter: Should the type here be `T[]`?

tobias-schreiner: I changed it.

PR #1165, opened 2019-08-09
Fixing backwards compatibility

ashley-hunter:
#### Checklist

#### Ticket / Issue

N/A

1. TypeScript converted `ReadonlyArray<T>` to `readonly T[]` which breaks in versions lower than 3.4
2. Angular 8 renamed `defineInjectable` and `inject` (https://angular.io/api/core/defineInjectable) and instead the import should now be `ɵɵdefineInjectable` and `ɵɵinject`, there is no change in functionality, just a rename to indicate it is a private function. While the old imports are still available until Angular 11, as we are now using the Angular 8 compiler it has been updated to use the private imports.

Both of these can be solved with TypeScripts transform API, which I have added as as step after the library build.

After this is run the package now build successfully in Angular 6, 7 and 8 in production mode.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_Fixing-Backwards-Compatibility

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -19,7 +19,7 @@ module.exports = function(grunt) {
 
     // Register Tasks
     grunt.registerTask('lint', ['tslint:library', 'tslint:documentation', 'jshint:ng1', 'stylelint', 'tslint:e2e']);
-    grunt.registerTask('library', ['clean:library', 'run:angular_components_build', 'run:webpack_ng1']);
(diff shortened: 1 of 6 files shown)


ashley-hunter:
We would, yes, if you are happy with the changes I will add it to the MF repo as well

donna-marijne:
For the ux-aspects-micro-focus project, you can still use the develop-1.8.2 branch since it has not yet been merged.

ashley-hunter:
Great! Thanks, will make this change now

PR #1166, opened 2019-08-12
EL-3552 - Add size selector to the Icons documentation

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3552

- Added in size selector for ux-icon for 16px, 24px, 32px
- Default 24px
- Added in <code>size</code> to ux-icon clipboard copy

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3552-add-size-selector-icon-documentation

### DIFF
--- a/docs/app/pages/css/css-sections/icons/icons.module.ts
+++ b/docs/app/pages/css/css-sections/icons/icons.module.ts
@@ -16,6 +16,7 @@ import { CssUxIconsComponent } from './ux-icons/ux-icons.component';
 import { IconModule, TabsetModule, TooltipModule } from '@ux-aspects/ux-aspects';
 import { IconPreviewComponent } from './ux-icons/icon-preview/icon-preview.component';
 import { IconSnippetComponent } from './ux-icons/icon-preview/icon-snippet/icon-snippet.component';
+import { ButtonsModule } from 'ngx-bootstrap';
 
 
 const SECTIONS = [
@@ -51,7 +52,8 @@ const ROUTES = [
         TooltipModule,
         PopoverModule,
         TabsetModule,
-        RouterModule.forChild(ROUTES)
+        RouterModule.forChild(ROUTES),
+        ButtonsModule
     ],
     exports: SECTIONS,
     declarations: [
--- a/docs/app/pages/css/css-sections/icons/icons.module.ts
+++ b/docs/app/pages/css/css-sections/icons/icons.module.ts
@@ -16,7 +16,7 @@ import { CssUxIconsComponent } from './ux-icons/ux-icons.component';
 import { IconModule, TabsetModule, TooltipModule } from '@ux-aspects/ux-aspects';
 import { IconPreviewComponent } from './ux-icons/icon-preview/icon-preview.component';
 import { IconSnippetComponent } from './ux-icons/icon-preview/icon-snippet/icon-snippet.component';
-import { ButtonsModule } from 'ngx-bootstrap';
+import { ButtonsModule } from 'ngx-bootstrap/buttons';
 
 
 const SECTIONS = [
--- a/docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.html
+++ b/docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.html
@@ -11,6 +11,7 @@
 
     <ux-icon
         class="p-r-xs"
+        [size]="iconSizeValue"
         [name]="name">
     </ux-icon>
 
--- a/docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.html
+++ b/docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.html
@@ -11,7 +11,7 @@
 
     <ux-icon
         class="p-r-xs"
-        [size]="iconSizeValue"
+        [size]="size"
         [name]="name">
     </ux-icon>
 
--- a/docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.ts
+++ b/docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.ts
@@ -1,4 +1,4 @@
-import { ChangeDetectionStrategy, Component, ElementRef, HostListener, Input, OnInit, ViewChild } from '@angular/core';
+import { ChangeDetectionStrategy, Component, ElementRef, HostListener, Input, OnChanges, ViewChild } from '@angular/core';
 import { PopoverDirective } from '@ux-aspects/ux-aspects';
 
 @Component({
@@ -7,12 +7,12 @@ import { PopoverDirective } from '@ux-aspects/ux-aspects';
     styleUrls: ['./icon-preview.component.less'],
     changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class IconPreviewComponent implements OnInit {
+export class IconPreviewComponent implements OnChanges {
     @Input() name: string;
     @Input() classname: string;
     @Input() iconset: string;
     @Input() iconSetClass: string;
-    @Input() iconSizeValue: string;
+    @Input() size: string;
 
     uxComponentSnippet: string;
     iconSnippet: string;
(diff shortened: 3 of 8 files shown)


donna-marijne:
Could you title your pull requests as follows please:

"EL-1234 - Description of issue"

This helps me to identify the purpose of pull requests when reviewing the list.

docs/app/pages/css/css-sections/icons/icons.module.ts
```diff
@@ -16,6 +16,7 @@ import { CssUxIconsComponent } from './ux-icons/ux-icons.component';
 import { IconModule, TabsetModule, TooltipModule } from '@ux-aspects/ux-aspects';
 import { IconPreviewComponent } from './ux-icons/icon-preview/icon-preview.component';
 import { IconSnippetComponent } from './ux-icons/icon-preview/icon-snippet/icon-snippet.component';
+import { ButtonsModule } from 'ngx-bootstrap';
```
ashley-hunter: We should update this to import path:

```typescript
import { ButtonsModule } from 'ngx-bootstrap/buttons';
```

Importing directly from `ngx-bootstrap` imports the entire `ngx-bootstrap` library, and while a lot of it can be tree shaken away, there are some parts which cannot safely be removed by Webpack. So we can import directly from buttons to avoid unnecessary code being bundled.

As you can see the difference in the size of code being bundled even after tree shaking:
image

docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.ts
```diff
@@ -21,7 +22,7 @@ export class IconPreviewComponent implements OnInit {
     @ViewChild(PopoverDirective, { static: true }) popover: PopoverDirective;
 
     ngOnInit(): void {
-        this.uxComponentSnippet = `<ux-icon name="${this.name}"></ux-icon>`;
+        this.uxComponentSnippet = `<ux-icon name="${this.name}" size="${this.iconSizeValue}"></ux-icon>`;
```
ashley-hunter: Perhaps this size property should only appear whenever 16px is not selected, otherwise this is suggesting we add additional unused markup.

docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.ts
```diff
@@ -21,7 +22,7 @@ export class IconPreviewComponent implements OnInit {
...
```
ashley-hunter: There also seems to be an issue, the size in the code doesn't seem to get updated:

[Image from Gyazo](https://gyazo.com/c15719471d329ba127378b900d214bee)

docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.html
```diff
@@ -11,6 +11,7 @@
 
     <ux-icon
```
ashley-hunter: Should the text be vertically centered with the icon?

image

docs/app/pages/css/css-sections/icons/ux-icons/ux-icons.component.ts
```diff
@@ -34,6 +34,8 @@ export class CssUxIconsComponent {
 
     uxIconComponentRoute: string;
 
+    iconSizeValue: string = '24px';
```
ashley-hunter: Can we simplify the naming of this to just `size`?

docs/app/pages/css/css-sections/icons/ux-icons/ux-icons.component.less
```diff
@@ -0,0 +1,25 @@
...
+        }
+
+        .icon-resize-buttons {
+            margin-left: 8px;
+
+            .icon-resize-button {
+                text-transform: none;
+
+                &:first-child {
+                    margin-left: -1px;
```
ashley-hunter: Is this required? Having it doesn't seem to add much benefit? Or is there a purpose for this?

TaraTodd123:
Fixed the above issues, regarding naming and the size not changing in the popup

docs/app/pages/css/css-sections/icons/ux-icons/ux-icons.component.ts
```diff
@@ -34,6 +34,8 @@ export class CssUxIconsComponent {
 
     uxIconComponentRoute: string;
 
+    size: string = '24px';
```
ashley-hunter: I wonder here is 24px the best default here? Is this going to be the most commonly used size in an application?

I realise that the icons naturally are 24x24 in size and look best at this resolution, but now we have the copy and paste feature, by default if a consumer opens the popover and clicks the copy icon it will copy a 24px size icon. Is this likely to be the most common choice?

If not then we should consider the 16px variation as the default so they either don't have to switch sizes to get the correct sample code or manually have to edit the code after they copy and paste it in to their application.

Thoughts?

docs/app/pages/css/css-sections/icons/ux-icons/icon-preview/icon-preview.component.ts
donna-marijne: Adding `size` to the code snippets was not part of this ticket, could you remove it altogether please? It's a nice idea but I think it will cause more problems than help.

TaraTodd123: Yes, removed now.

docs/app/pages/css/css-sections/icons/ux-icons/ux-icons.component.ts
TaraTodd123: The size has now been removed, from the code snippet. I do agree with this as it nearly gives the impression that 24px is the default size, without having to add extra code.

@donna-marijne

TaraTodd123:
Updates to size selector made from mock-up

TaraTodd123:
- Added a responsive layout 
- Keep ux-select and ux-combobox as styles aren't right for readonly in OS site for ux-combobox

docs/app/pages/css/css-sections/icons/ux-icons/ux-icons.component.html
```diff
@@ -1,28 +1,79 @@
...
+                   type="text"
+                   class="form-control"
+                   placeholder="Search UX Aspects Icons..."
+                   [(ngModel)]="query"
+                   (ngModelChange)="search()">
+
+        </div>
+    </div>
+
+    <div class="col-lg-4 col-md-12 col-sm-12">
```
ashley-hunter: Just noticed one issue with text wrapping when the hpe-icon set is selected:

image

TaraTodd123: fixed

docs/app/pages/css/css-sections/icons/ux-icons/ux-icons.component.less
```diff
@@ -0,0 +1,21 @@
+.icon-search-items-container {
+    display: flex;
+    align-items: center;
+    padding: 5px 0;
+
+    .icon-selector-text {
+        justify-content: center;
+        flex-direction: column;
```
donna-marijne: These don't do anything without `display: flex` and therefore should be removed.

Also if you add `flex: none` in here you can fix the nasty line breaking issue.

docs/app/pages/css/css-sections/icons/ux-icons/ux-icons.component.html
```diff
@@ -1,28 +1,79 @@
...
+
+        <div class="icon-search-items-container">
+
+            <div class="icon-selector-text">Icon Set</div>
```
donna-marijne: This should be `label` rather than `div`. Same for the other label.

docs/app/pages/css/css-sections/icons/ux-icons/ux-icons.component.less
```diff
@@ -0,0 +1,21 @@
...
+        margin-right: 8px;
+    }
+
+    .icon-set-dropdown {
```
donna-marijne: Does this match anything or can it be removed?

TaraTodd123:
Updates to the above are done

PR #1167, opened 2019-08-12
EL-3521 - Protractor Screenshot Comparison

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3620

- Added new screenshot functionality to e2e tests
- Only added to tests that actually have a visual change to reduce number of screenshots and comparison overhead (although at least locally it seems to be pretty quick)
- Removed all use of chance in our tests as random data obviously changes the appearance of the tests each time. Also, the library is quite large so removing this should also speed up the tests a bit too.
- Updated some older tests to use the `async`/`await` approach, because otherwise timing could be inconsistent causing the screenshots to differ in various runs. We also get better type checking as previously we did a lot of `any` casting.
- I have to also check in the baseline screenshots against which all future tests will run against. If in future we change the appearance of a component, deleting the relevant screenshot will automatically generate a new baseline image when the tests are next run rather than performing a comparison.
- Screenshots generated in a Linux environment have different font rendering so this will always produce differences. I have added a check to disable image comparison when running in a non-linux environment. I have also produced a dockerfile that creates a linux environment with everything we need to run the tests. It mounts the `e2e` and `dist` folders so they will match the current dev environment. We cannot mount the entire dev environment as some node modules are built specifically for a windows environment so trying to run using them on a linux environment will cause errors. Also by mounting other folders adds quite a lot of overhead to the docker context. There are two scripts added to package.json, one for building the image (only really needs to be run once unless changes are made to things like webpack config etc..) and one to run the tests which will output baseline screenshots to the dev environment if new ones are added.
- Updated e2e webpack config to use `html-loader` instead of `raw-loader` to ensure image paths are correct. This now matches the way we do it in our docs configs etc..

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3521-Protractor-Image


ashley-hunter:
We might be able to use it as the base image, for installing node, chrome etc. but I'm not sure how we could do the rest without having a custom Dockerfile, as we need to have a full dev environment available, with the `e2e` and `dist` folders mounted so screenshots get saved outside of the container and a separate `node_modules` folder as there are modules that are built for windows that wont work in a linux container etc. I'm open to suggestions if you know how this might be possible as docker is not something I am hugely familiar with.

The error you are getting looks like a network issue, is your proxy set for docker?

donna-marijne:
I guess I'm asking this because the tests are going to run on the build image anyway when we do SEPG builds. So I am wondering what the difference would be in this case? Would we need to run the whole build inside the image to get the node modules installed correctly?

donna-marijne:
Proxy should be okay as far as I can tell?
image

ashley-hunter:
So from my understanding, the only thing the build image does, is set up a linux environment with node and chrome installed.

image

The build and tests are run in that container once it is created and not part of the image itself. So even if I use the build image, its not going to have any of the UX Aspects files or run any scripts etc, so I would still need to have all these additional steps somewhere? Unless I am misunderstanding?

You could try: http://web-proxy.eu.softwaregrp.net:8080, that's what I'm using and it seems to work.

donna-marijne:
No exactly, it's just a clean linux environment, and our project contains all the other needed dependencies.

The other thing is that we should keep the build/test stuff as similar as possible with the SEPG process, otherwise there's a risk of getting different results in each place.

I think the answer should be that if we need to generate screenshots or just run the screenshot comparison stuff on a dev environment, then we run the whole thing (build and test) inside the build env docker image. So then we mount the whole project directory. If it gets run on windows then we can either skip the image checks (I think you mentioned this) or just fail and say that e2e is only valid in a linux env.

Let me know your thoughts on this. I'm not 100% sure about any of this so far.

ashley-hunter:
Ok, no problem, that makes sense, I think that could work, we will need to try and exclude the node_modules folder from mounting and run `npm install` inside the container otherwise it will fail. I will look into that now

donna-marijne:
Thanks, I don't think we need to do anything special about node_modules other than document to run npm ci before building and running tests.

We'll also need to have an md document in the repo to describe the whole process.

ashley-hunter:
I have updated to use the build environment locally, you can just run `npm run docker:ci` and it should spin up a container with the same environment as the CI machine.

I have excluded the `node_modules` folder so you need to run `npm install` when you start the container. You can't run `npm ci` as that deletes the folder first and docker has a lock on the folder so it'll throw an error. I can change it to mount the `node_modules` if you'd prefer, in which case `npm ci` would work, however once you finish running the container you would need to do another `npm ci` to get things to work again on your dev environment again as the container will have overriden `node_modules` with the linux packages.

I have updated the protractor tests doc with instructions on this, feel free to suggest any changes or if you think it should be a separate doc etc..

donna-marijne:
I think we need to use npm ci, npm install ignores package-lock.json so we get different dependencies.

ashley-hunter:
`npm install` will use also the `package-lock.json` file the same as `npm ci` does:

image.

Its a very easy change and I'm happy to make it if you want me to, I was just trying to avoid breaking the dev environment when running the container.

donna-marijne:
It doesn't work like that anymore though, see https://github.com/npm/npm/issues/18103

ashley-hunter:
Ah, I was not aware that this had changed, thanks! Made the changes to support `npm ci`

donna-marijne:
Added Tara to help review since I'm still having network problems with docker.

scripts/run-ci-local.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
donna-marijne: I think both Tara and Tobias have been using the Windows CMD shell for development, in which case this script will not run.

We could document that bash is a prerequisite for UX Aspects contribution. But it might be a bit more friendly to reimplement this as a node script.

ashley-hunter:
Thanks for the feedback! I have made the following changes:

- Migrated to a Node.js script so bash should no longer be required
- Increased memory limit in the `docker run` arguments so this should fix the failing tests issues
- Updated documentation to be clearer

donna-marijne:
@TaraTodd123 After you merge, be sure to delete the branch as well. Thanks.

PR #1184, opened 2019-08-28
Input dropdown console error fix

tobias-schreiner:
#### Checklist

#### Ticket / Issue
Avoid console errors in input drop-down:

- When the component is loaded, MenuTriggerDirective.closeMenu() is called before view is initialized

- When opening the drop-down when filter is suppressed ([hideFilter]="true"), filterInputElement.nativeElement.focus() is called

Access view children only when available

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/microfocusbbn_CI_UXAspects_input-dropdown-console-error-fix/#/landing


src/components/input-dropdown/input-dropdown.component.ts
```diff
@@ -97,4 +102,10 @@ export class InputDropdownComponent<T> implements ControlValueAccessor, OnChange
         this.writeValue(undefined);
         event.stopPropagation();
     }
+
+    _focusFilter() {
```
ashley-hunter: Can we add missing return type?

donna-marijne:
Have you made a release build and verified that this resolves both problems in your production environment?

src/components/input-dropdown/input-dropdown.component.ts
```diff
@@ -60,14 +61,18 @@ export class InputDropdownComponent<T> implements ControlValueAccessor, OnChange
 
     ngOnChanges(changes: SimpleChanges): void {
         if (changes.selected) {
-            if (this.menuTrigger) {
+            if (this.menuTrigger && this._viewInitialized) {
```
ashley-hunter: Instead of having to do this view initialized thing, is it possible to use the `firstChanges` property to skip this initial close menu call, eg:

```typescript
if (this.menuTrigger && !changes.selected.firstChange)
```

This might not work, but it would be great to check if it did

src/components/input-dropdown/input-dropdown.component.ts
tobias-schreiner: I will fix that.

src/components/input-dropdown/input-dropdown.component.ts
tobias-schreiner: I tried it, but unfortunately, it does not always work. Sometimes the first selection click in the component does not close the menu.

tobias-schreiner: However, I liked the idea so much that I played around with it some more. I found and fixed a defect in our component and now it works. I just committed both fixes.

ashley-hunter: Great! Thanks for making the changes!

tobias-schreiner:
> Have you made a release build and verified that this resolves both problems in your production environment?

Yes, from the beginning.

PR #1185, opened 2019-08-28
EL-3626 - select subject

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3626

- inputChange and ValueChange no longer being called on initialization for select component
- Tests added for select component for inputChange, ValueChange and blurring the input
- InputChange and tagChange no longer being called on initialization in tags component
- Tests added for tag component for inputChange and tagsChange

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3626-select-subject

### DIFF
--- a/src/components/select/select.component.ts
+++ b/src/components/select/select.component.ts
@@ -180,6 +180,7 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
     private _value: T | ReadonlyArray<T>;
     private _input$ = new BehaviorSubject<string>('');
     private _dropdownOpen: boolean = false;
+    private _valueEmpty: boolean = false;
     private _onDestroy = new Subject<void>();
 
     constructor(
@@ -272,7 +273,7 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
             if (!this._element.nativeElement.contains(this._document.activeElement)) {
                 this.dropdownOpen = false;
                 if (!this.multiple) {
-                   this.input = this.getDisplay(this.value);
+                    this.input = this.getDisplay(this.value);
                 }
             }
         }, 200);
@@ -301,7 +302,8 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
         }
 
         // when the user types and the value is not empty then we should open the dropdown
-        if (event.keyCode !== ESCAPE) {
+        if (event.keyCode !== ESCAPE && !this._valueEmpty) {
+            this._valueEmpty = true;
             // open the dropdown once the filter debounce has elapsed
             this.filter$.pipe(take(1), takeUntil(this._onDestroy))
                 .subscribe(() => this.dropdownOpen = true);
--- a/src/components/select/select.component.ts
+++ b/src/components/select/select.component.ts
@@ -37,6 +37,10 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
         this._value$.next(value);
     }
 
+    get value(): T | ReadonlyArray<T> {
+        return this._value;
+    }
+
     /** The text in the input area. This is used to filter the options dropdown. */
     @Input()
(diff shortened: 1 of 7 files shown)


ashley-hunter:
I meant to tidy up some code that I added last week when we were working on the tag input, so I have pushed those changes

src/components/select/select.component.ts
```diff
@@ -270,7 +272,7 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
             if (!this._element.nativeElement.contains(this._document.activeElement)) {
                 this.dropdownOpen = false;
                 if (!this.multiple) {
-                    this.input = this.getDisplay(this.value);
+                   this.input = this.getDisplay(this.value);
```
donna-marijne: This was an incorrect indentation change, please re-do indentation for the file just to make sure.

src/components/select/select.component.ts
```diff
@@ -284,20 +286,25 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
...
 
-                // Update the input field. If dropdown isn't open then reset it to the previous value.
-                this.input = this.getDisplay(this.value);
-                event.preventDefault();
-                break;
+        // when the user types and the value is not empty then we should open the dropdown
+        if (event.keyCode !== ESCAPE) {
+            // open the dropdown once the filter debounce has elapsed
+            this.filter$.pipe(take(1), takeUntil(this._onDestroy))
+                .subscribe(() => this.dropdownOpen = true);
```
donna-marijne: This doesn't look right to me, could be a memory leak. Instead of subscribing every time esc is pressed, can you not just subscribe once and set a flag here to check in the subscription?

src/components/tag-input/tag-input.component.ts
```diff
@@ -51,19 +51,10 @@ export class TagInputComponent<T = any> implements AfterContentInit, OnChanges,
...
     /** The editable text appearing in the tag input. */
-    @Input()
-    get input() {
-        return this._input;
-    }
-    set input(value: string) {
-        this._input = value;
-        this.inputChange.emit(value);
-    }
+    @Input() input: string = '';
```
donna-marijne: Was there a reason not to use Subject here as well? To my understanding it's the same problem as with `value`.

donna-marijne: I can see it used in one place inside `keyHandler`:

````ts
const inputLength = this.input ? this.input.length : 0;
````

This is something that could be calculated in a subscription.

My issue is with code maintainability rather than whether this works or not.

Comment (ashley-hunter) [reply src/components/tag-input/tag-input.component.ts]: We also have the `commitInput` function which reads the `this.input` value are you happy to have this stored in a private variable as well?

Comment (donna-marijne) [reply src/components/tag-input/tag-input.component.ts]: Ah, you're right, I missed that. Let's just leave it as it is then.

**Comment by TaraTodd123** [comment]
Made the above updates

File: src/components/select/select.component.ts [hunk 4]
```diff
@@ -301,7 +302,8 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
         }
 
         // when the user types and the value is not empty then we should open the dropdown
-        if (event.keyCode !== ESCAPE) {
+        if (event.keyCode !== ESCAPE && !this._valueEmpty) {
+            this._valueEmpty = true;
             // open the dropdown once the filter debounce has elapsed
             this.filter$.pipe(take(1), takeUntil(this._onDestroy))
```
ashley-hunter: I think the issue is with the filter subscription in here. I added this in as a temporary fix yesterday to get it working, but meant to move this, I will help you make this fix after standup

TaraTodd123: updated

PR #1196, opened 2019-09-17
EL-3636 - Number picker validation fix and tests

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3636

- removed has-error styling when disabled
- Added styling for ng-invalid
- Added in tests for number picker

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3636-number-picker-validation

### DIFF
--- a/docs/app/pages/components/components-sections/input-controls/number-picker/number-picker.component.html
+++ b/docs/app/pages/components/components-sections/input-controls/number-picker/number-picker.component.html
@@ -41,6 +41,10 @@
         	Defines the maximum value the number picker can set.
     </tr>
     <tr uxd-api-property name="valid" type="boolean">
+        <p><b>Note:</b> This component has been deprecated. We recommend using <a
+            routerLink="https://angular.io/guide/form-validation#reactive-form-validation">reactive form validation</a>
+            instead.
+        </p>
         Can be used to show a red outline around the input to indicate an invalid value. By default the
         error state will appear if the user enters a number below the minimum value or above the maximum value.
     </tr>
--- a/docs/app/pages/components/components-sections/input-controls/number-picker/number-picker.component.html
+++ b/docs/app/pages/components/components-sections/input-controls/number-picker/number-picker.component.html
@@ -41,9 +41,9 @@
         	Defines the maximum value the number picker can set.
     </tr>
     <tr uxd-api-property name="valid" type="boolean">
-        <p><b>Note:</b> This component has been deprecated. We recommend using <a
-            routerLink="https://angular.io/guide/form-validation#reactive-form-validation">reactive form validation</a>
-            instead.
+        <p><b>Note:</b> This property has been deprecated. We recommend using <a
+            href="https://angular.io/guide/form-validation#reactive-form-validation">reactive form
+            validation</a> instead.
         </p>
         Can be used to show a red outline around the input to indicate an invalid value. By default the
         error state will appear if the user enters a number below the minimum value or above the maximum value.
--- a/src/components/number-picker/number-picker.component.less
+++ b/src/components/number-picker/number-picker.component.less
@@ -4,7 +4,7 @@ ux-number-picker-inline {
     flex-direction: row;
     height: 34px;
 
-    &.ux-number-picker-invalid:not([formControlName]) {
+    &.ux-number-picker-invalid:not(.ng-invalid) {
         > .number-picker-input {
             border-color: @critical;
         }
--- a/src/components/number-picker/number-picker.component.less
+++ b/src/components/number-picker/number-picker.component.less
@@ -4,6 +4,10 @@ ux-number-picker-inline {
     flex-direction: row;
     height: 34px;
 
+    &.ng-invalid {
+        border-color: @critical;
+    }
+
     &.has-error {
         > .number-picker-input:focus {
             border-color: @critical;
--- a/src/components/number-picker/number-picker.component.less
+++ b/src/components/number-picker/number-picker.component.less
@@ -4,12 +4,14 @@ ux-number-picker-inline {
     flex-direction: row;
     height: 34px;
 
-    &.ng-invalid {
-        border-color: @critical;
+    &.ux-number-picker-invalid:not([formControlName]) {
+        > .number-picker-input {
+            border-color: @critical;
+        }
     }
 
-    &.has-error {
-        > .number-picker-input:focus {
+    &.ng-invalid {
+        > .number-picker-input {
             border-color: @critical;
         }
     }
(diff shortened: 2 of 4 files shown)


src/components/number-picker/number-picker.component.ts
```diff
@@ -15,7 +15,7 @@ export const NUMBER_PICKER_VALUE_ACCESSOR: any = {
     templateUrl: './number-picker.component.html',
     providers: [NUMBER_PICKER_VALUE_ACCESSOR],
     host: {
-        '[class.has-error]': '!isValid()'
+        '[class.has-error]': '!_valid && !disabled'
```
donna-marijne: With the new `.ng-invalid` style, can this now be removed?

ashley-hunter: I don't think we can as we still support the `[(value)]` input/output, and `ng-invalid` will only be present when either `ngModel` or `ReactiveForms` are being used.

donna-marijne: Are there any recommendations or existing Angular controls which work this way? I feel like forcing validation appearance may continue to cause problems. Like we previously overlooked the disabled case, which is now baked into the component, but consumers may have other complex validation rules within their form layout in which case we would only want the .ng-invalid rule to apply.

ashley-hunter: I can't think of any other components of the top of my head that work this way. I agree, this was something that was overlooked at the time, and I would rather not have the component handle any validation itself but use Angular validation only.

Currently we have a `[valid]` input which a user can set to `true` if they want to show the red outline and we also have `[min]` and `[max]` inputs which are also used to validate the input. This isn't ideal as `ngModel` uses the `min` and `max` HTML attributes to also perform validation. 

I'm not sure how we can alter this without a breaking change as we have an input that can be specifically used to set the red outline,

donna-marijne: That's true. I actually think we ought to deprecate the `valid` input and advise to either use reactive forms, or apply `has-error` to the parent element (the Bootstrap way). Not sure what to do about min and max but with 2.0 coming up we do have an opportunity to make breaking changes.

In terms of current behavior, one idea is to change the name of the class from `has-error` to something specific like `ux-number-picker-invalid`. We can then make a style rule like

````css
ux-number-picker:not([formControlName]).ux-number-picker-invalid {
    border-color: red;
}
````

or something along those lines so that it only applies when the number picker is not part of a reactive form.

Comment (ashley-hunter) [reply src/components/number-picker/number-picker.component.ts]: Yeah, I agree, with 2.0 looming now is the time to deprecate things like this. Yeah, that makes sense, I think renaming the class is the best way rather than having the component update a class on something outside of it's template.

File: src/components/number-picker/number-picker.component.less [hunk 2]
```diff
@@ -4,12 +4,14 @@ ux-number-picker-inline {
     flex-direction: row;
     height: 34px;
 
-    &.ng-invalid {
-        border-color: @critical;
+    &.ux-number-picker-invalid:not([formControlName]) {
```
donna-marijne: I think this isn't sufficient because you can also bind a control with `[formControl]`. Possibly `:not(.ng-valid):not(.ng-invalid)`?

TaraTodd123: I cant have <code>:not(.ng-valid)<code> as this does not apply the border when using <code>ngModel</code> as it does not apply the <code>ng-invalid</code> class

donna-marijne: Do you mave a plunker example? I'm not sure I understand the scenario that you are describing.

ashley-hunter: I think this is the case where they are using `ngModel` to set the value but also using the built in `min` and `max` functionality. ngModel will apply the `ng-valid` class if none of its validation rules fail, but there is still the possibility of the components built in validation failing, so we could have both ng-valid and number-picker-invalid classes present.

TaraTodd123: Yes, that's what I mean ^^

TaraTodd123: So, this is using the current has-error styles but, if you change the value to 20 above the max value, the ng-invalid class doesn't apply and it still says ng-valid. https://plnkr.co/edit/Sri8u348U3JuhdlLu0Xj?p=preview

donna-marijne: Well you need to use `[min]` and `[max]` for number inputs: https://plnkr.co/edit/pkWyNVVsT2JmO2aTa8rG?p=preview

donna-marijne: Oh, that was with the old stylesheet, I see the point. What selector will work for this?

TaraTodd123: for that example <code>.ux-number-picker-invalid</code> will be applied

donna-marijne: `:not(.ng-valid):not(.ng-invalid):not([ngModel])`?

donna-marijne: Maybe you could do an optional inject of the parent FormGroupDirective, and set `ux-number-picker-valid` only if that is null?

TaraTodd123: I'm not sure if this will work as <code>ngModel</code> doesn't produce a <code>formGroup</code>. What I have currently applies either <code>ng-invalid</code> or the <code>.ux-number-picker-invalid</code>

donna-marijne: Yes, when `ngModel` is used then we will apply `ux-number-picker-invalid`. When a form group is present, we assume that validation is handled externally.

TaraTodd123: then would <code> &.ux-number-picker-invalid:not(.ng-invalid):not([formControlName])</code> work?

donna-marijne: The trouble with that is that you can alternatively use `[formControl]="myControl"` to bind a form control.

TaraTodd123: You could add <code>:not[(formControl)]</code> aswell? If not, what way do you want this to be done?

donna-marijne: I think that wouldn't match because object bindings aren't rendered to the DOM in the same way.

I'd suggest injecting the parent FormGroupDirective into the number picker, and only emitting the `ux-number-picker-validation` class if the directive is null. Does that make sense?

TaraTodd123: Yes, I will do that now.

docs/app/pages/components/components-sections/input-controls/number-picker/number-picker.component.html
```diff
@@ -41,6 +41,10 @@
         	Defines the maximum value the number picker can set.
     </tr>
     <tr uxd-api-property name="valid" type="boolean">
+        <p><b>Note:</b> This component has been deprecated. We recommend using <a
```
donna-marijne: This should be "property" rather than "component." Also, could you use `href` instead of `routerLink` for the external URL? Thanks.

TaraTodd123: Updated

PR #1198, opened 2019-09-20
EL-3637 - Float Label AutoFill Issue

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3637

- AutoFillMonitor added
- CSS for AutoFillMonitor Imported
- https://plnkr.co/edit/6k36pQ790o7r6MrmMP9t?p=preview - AutoFillMonitor plunker

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3637-float-label-autofill-issue

### DIFF
--- a/src/directives/float-label/float-label.directive.ts
+++ b/src/directives/float-label/float-label.directive.ts
@@ -32,7 +32,7 @@ export class FloatLabelDirective<T = string> implements OnInit, OnChanges, OnDes
             if (!this.raised && event.isAutofilled) {
                 this.raised = true;
             }
-            if (this.raised && !event.isAutofilled) {
+            if (this.raised && !event.isAutofilled && !this._input) {
                 this.raised = false;
             }
         });
--- a/src/directives/float-label/float-label.directive.ts
+++ b/src/directives/float-label/float-label.directive.ts
@@ -1,4 +1,6 @@
 import { Directive, ElementRef, HostBinding, Input, OnChanges, OnDestroy, OnInit, Renderer2 } from '@angular/core';
+import { AutofillMonitor } from '@angular/cdk/text-field';
+import { Subscription } from 'rxjs';
 
 @Directive({
     selector: '[uxFloatLabel]',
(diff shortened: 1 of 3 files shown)


src/directives/float-label/float-label.directive.ts
```diff
@@ -1,18 +1,49 @@
...
+        if (!input) {
+            return;
+        }
+
+        // create a new autofillMonitor subscription
+        this._subscription = this._autofillMonitor.monitor(input).subscribe(event => {
+            if (!this.raised && event.isAutofilled) {
+                this.raised = true;
+            }
+            if (this.raised && !event.isAutofilled) {
```
ashley-hunter: I wonder could you check this for me? This condition says that it the label is raised and we remove autofill we should set raised to false. 

What would happen if the user typed in `t` (this would raise the text), then auto fill offers a suggestion for `tara.todd@microfocus.com` which you hover over (isAutofilled now true), now if I hover out of the autofill (isAutofill now false), but there is text in the input so raised should remain true. I suspect we need to check if there is no text in the input before setting raised to false

src/directives/float-label/float-label.directive.ts
```diff
@@ -1,18 +1,49 @@
...
+        if (!input) {
+            return;
+        }
+
+        // create a new autofillMonitor subscription
+        this._subscription = this._autofillMonitor.monitor(input).subscribe(event => {
+            if (!this.raised && event.isAutofilled) {
+                this.raised = true;
+            }
+            if (this.raised && !event.isAutofilled && !this._input) {
```
ashley-hunter: Is `this._input` not the input element? Perhaps it should be the `hasText` function used to check?

TaraTodd123:
That should be the plunker fixed.

donna-marijne:
Tested with the following plunker in Chrome: https://plnkr.co/edit/1txJ1oNsY9GH5yBraiyD?p=preview

Labels were not raised following autofill:
image

Is the test case valid?

ashley-hunter:
I think that form is setup slightly wrong, think this should be it working:

https://plnkr.co/edit/LZ3PrK2ZD3cslpjW0DqY?p=preview

donna-marijne:
Gotcha, thank you!

PR #1210, opened 2019-10-07
EL-3651 - Iconset selector dropdown

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3651

- placeholder now defaulted to empty string if not assigned
- fixed issue when value binding was not working
- added karma test for placeholder in select component
- added e2e for having initial value set and updated screenshots for select component

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3651-iconset-selector-dropdown


src/components/select/select.component.spec.ts
```diff
@@ -43,6 +43,10 @@ describe('Select Component', () => {
         fixture.detectChanges();
     });
 
+    it('should be initially blank if no placeholder is set', () => {
+        expect(component.input).toBe('');
```
ashley-hunter: I think perhaps we should be testing the element `placeholder` property instead of the `input` property, perhaps testing both when empty and when populated? I don't believe the `input` value would ever change when the placeholder value changes?

src/components/select/select.component.spec.ts
```diff
@@ -43,6 +43,10 @@ describe('Select Component', () => {
...
+    });
+
```
ashley-hunter: Would it be possible to add tests to cover an initial value when using `value`, when using `ngModel` and when using `ReactiveForms`? Just so this issue won't regress again.

And possibly a test to ensure the value doesn't immediate get emitted when an initial value is set in each case?

src/components/select/select.component.spec.ts
```diff
@@ -43,6 +43,10 @@ describe('Select Component', () => {
         fixture.detectChanges();
     });
 
+    it('should be initially blank if no placeholder is set', () => {
+        expect(component.input).toBe('');
+    });
```
donna-marijne: This is more like a unit test. Could you use karma instead for this?

TaraTodd123:
Updates made for the above

donna-marijne:
Could you remove the e2e changes as requested, or is there a problem with doing so?

TaraTodd123:
Updates for the above have been made.

PR #1211, opened 2019-10-09
EL-3662-Fixing tag input error

dermot-hughes-mf:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3662

Fixing error that occurs in console when clicking on input field when tag limit is reached

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3662-Tag-Input-Error

### DIFF
--- a/e2e/tests/components/tags/tags.e2e-spec.ts
+++ b/e2e/tests/components/tags/tags.e2e-spec.ts
@@ -265,4 +265,18 @@ describe('TagsPage Tests', () => {
         expect(await page.getNumberOfTagsInTypeaheadList()).toEqual(24);
 
     });
+
+    /**
+     * Fixing https://portal.digitalsafe.net/browse/EL-3662
+     * Console error was thrown when clicking on the tag input after
+     * max tags had been reached and enforceTagLimits was true
+     */
+    it('should not focus input when max tag limit is reached and enforceTagLimits is true', async () => {
+
+        // set enforce tag limits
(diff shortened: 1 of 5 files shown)


e2e/tests/components/tags/tags.e2e-spec.ts
```diff
@@ -265,4 +265,18 @@ describe('TagsPage Tests', () => {
...
+     * Console error was thrown when clicking on the tag input after
+     * max tags had been reached and enforceTagLimits was true
+     */
+    it('should not focus input when max tag limit is reached and enforceTagLimits is true', async () => {
+
+        // set enforce tag limits
+        await page.enforceTagLimits.click();
+        await page.changeMaxTags('3');
+        // ensure there are there 3 tags
+        await page.clickOnTagsInput(); // console error would have been thrown here before fix
```
donna-marijne: This test failed because this is attempting to click on an element that doesn't exist. You could either click using an offset to where the input element would be, or just assert that `input.ux-tag-input` is not present (e.g. https://www.protractortest.org/#/api?view=ElementFinder.prototype.isPresent).

dermot-hughes-mf: The test was using the wrong element selector. Should be pointing at the correct one now!

PR #1216, opened 2019-10-10
Fix infinite scroll within tab

victor-ac:
#### Checklist

#### Ticket / Issue

* When an infinite scrolling element is placed inside a tab and that tab is selected and the deselected, the pageFunction will be continuously called, even though the tab is no longer selected. Depending on the approach adopted by the application consuming UX Aspects, this can either lead to:
  * a constant fetch of new records until the last page is fetched or the tab with infinite scroll is reselected; OR
  * a break in the functionality to fetch new records, even when the last page hasn't been returned yet, if the application doesn't return records is the tab with infinite scroll is not selected
* Here's a Plunkr to reproduce this issue: [link](https://next.plnkr.co/edit/mIn1b4FCcFe62bQS)
* This affects SCMOD-7844

* Include a check to see if the infinite scroll element is visible before loading the next page of records

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/victor-ac_CI_UXAspects_fix-infinite-scroll-within-tab/


src/ng1/directives/infiniteScroll/infiniteScroll.controller.js
```diff
@@ -89,8 +89,10 @@ export class InfiniteScrollController {
         // repeat until we have enough items
         this._ensureScrollableInterval = this.$interval.interval(() => {
 
+            let isElementHidden = this.$element.is(':hidden');
+
             // if we are currently loading or have loaded all pages then do nothing
-            if (this.loading || this.complete || document.hidden) {
+            if (this.loading || this.complete || document.hidden || isElementHidden) {
```
ashley-hunter: Hi Victor, thanks very much for the contribution, I wonder would we inline the `is` call. It can be quite an expensive check so adding it at the end of the if statement would allow us to only call it if all the other checks are false, eg:

```typescript
if (this.loading || this.complete || document.hidden || this.$element.is(':hidden')) {
```

ashley-hunter:
@donna-marijne Are you happy for this to be merged in with the current sprint content?

donna-marijne:
Yes, sorry, this should be fine to add. I'll make a placeholder ticket so that it doesn't get missed from the release notes. Thank you both.

ashley-hunter:
Great! Thanks!

PR #1227, opened 2019-10-23
EL-3672-slider-track-range

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3672

- Added in ability to update slider options after they have been initially set
- Default track colors now set in css
- Added in e2e test for changing track bar colours with screenshot
- Ability to take full screen screenshots

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3672-slider-track-range


e2e/tests/components/sliders/sliders.e2e-spec.ts
```diff
@@ -351,6 +351,15 @@ describe('SlidersPage Tests', () => {
 
     });
 
+    it('should allow track styling to be changed', async () => {
```
donna-marijne: I would like this test to go as follows:
1. Start with default colors (no `track.colors` defined).
2. Verify an initial screenshot.
3. Change to the updated colors.
4. Verify a second screenshot.

src/components/slider/slider.component.ts
```diff
@@ -330,8 +337,8 @@ export class SliderComponent implements OnInit, AfterViewInit, DoCheck {
         // get the element widths
         let thumbWidth: number;
 
-        if (this.options.handles.style === SliderStyle.Button) {
-            thumbWidth = this.options.track.height === SliderSize.Narrow ? 16 : 24;
+        if (this._options.handles.style === SliderStyle.Button) {
```
donna-marijne: I don't think this needed to change to `_options` because you defined an `options` getter.

ashley-hunter: There is a performance benefit from accessing the property instead of the getter. We access options a lot, so by accessing the property directly we skip the intermediate function call:

https://stackblitz.com/edit/typescript-womh4x

ashley-hunter: Actually, we also use `options` in the template, so ideally, `_options` would not be marked as `private` and the template should be updated to use `_options` instead. (https://www.reddit.com/r/Angular2/comments/6id6mf/low_performance_of_getters_binded_to_templates/)

TaraTodd123:
Updates for the above made

TaraTodd123:
We could possibly share them now, but I think there may be some additional work. So it would be probably best for the sharing of the tests to be done in a different ticket.

TaraTodd123:
Updates made for the above

PR #1244, opened 2019-11-12
Select / Typeahead: add recent options list

tobias-schreiner:
#### Checklist

#### Ticket / Issue
Enhancement: prepend a list of recently selected options to the drop-down list

Changed typeahead and select component as well as their demo pages and Plunkr snippets

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/microfocusbbn_CI_UXAspects_master


src/components/select/select.component.ts
```diff
@@ -151,6 +151,12 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
      */
     @Input() optionTemplate: TemplateRef<any>;
 
+    /** Container for saving the recently selected options. */
+    @Input() recentOptions: T[];
```
ashley-hunter: Can we use the `ReadonlyArray<T>` type for the new inputs and outputs you added instead of standard array types `T[]`. 

Enforcing immutability can prevent issues with change detection, as change detection performs a `===` comparison, so allow things like `push` to the array can prevent change detection from noticing the change, whereas the `ReadonlyArray` type removes any functions that will change the array without changing the array instance.

src/components/typeahead/typeahead.component.ts
```diff
@@ -1,10 +1,24 @@
-import { FocusOrigin } from '@angular/cdk/a11y';
-import { ChangeDetectionStrategy, ChangeDetectorRef, Component, ElementRef, EventEmitter, HostBinding, HostListener, Input, OnChanges, OnDestroy, Output, SimpleChanges, TemplateRef } from '@angular/core';
-import { BehaviorSubject, combineLatest, Subject } from 'rxjs';
-import { distinctUntilChanged, takeUntil } from 'rxjs/operators';
-import { InfiniteScrollLoadedEvent, InfiniteScrollLoadFunction } from '../../directives/infinite-scroll/index';
-import { TypeaheadOptionEvent } from './typeahead-event';
-import { TypeaheadService } from './typeahead.service';
+import {FocusOrigin} from '@angular/cdk/a11y';
```
ashley-hunter: Can we fix the formatting, all our scripts have spaces around the imports eg: `{ FocusOrigin }` instead of `{FocusOrigin}`

src/components/typeahead/typeahead.component.ts
```diff
@@ -99,13 +119,17 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
     /** Emit the highlighted element when it changes */
     @Output() highlightedElementChange = new EventEmitter<HTMLElement>();
 
+    /** Emits when recently selected options change.*/
+    @Output() recentOptionsChange =  new EventEmitter<T[]>();
```
ashley-hunter: Can we remove the additional space here after the `=`?

src/components/typeahead/typeahead.component.ts
```diff
@@ -99,13 +119,17 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
...
+
     activeKey: string = null;
     clicking = false;
     hasBeenOpened = false;
     highlighted$ = new BehaviorSubject<TypeaheadVisibleOption<T>>(null);
     highlightedKey: string = null;
     loadOptionsCallback: InfiniteScrollLoadFunction;
     visibleOptions$ = new BehaviorSubject<TypeaheadVisibleOption<T>[]>([]);
+    recentOptionsWrapped$ = new BehaviorSubject<TypeaheadVisibleOption<T>[]>(null);
```
ashley-hunter: Would it be safer here to default to a value of `[]` instead of null?

src/components/typeahead/typeahead.component.ts
```diff
@@ -191,6 +215,22 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
 
         // Re-filter visibleOptions
         this.updateOptions();
+
+        if (changes.recentOptions) {
```
ashley-hunter: Could we make `recentOptions` a setter instead of using `ngOnChanges`, as ngOnChanges will only get triggered when change detection runs, but if I was to set this array programmatically the observable would not get updated, eg:

```
@ViewChild(SelectComponent, { static: false ) select: SelectComponent;

...
ngAfterViewInit(): void {
    this.select.recentOptions = [...];
}
```

Whereas with a setter this would update the observable

src/components/typeahead/typeahead.component.ts
```diff
@@ -215,6 +255,20 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
 
     optionClickHandler(_event: MouseEvent, option: TypeaheadVisibleOption<T>): void {
         this.select(option, 'mouse');
+
+        if (Array.isArray(this.recentOptions)) {
+            // If the option is already in the array, remove it first.
+            let index = this.recentOptions.indexOf(option.value);
```
ashley-hunter: Can we use `const` here instead of `let` as they are never reassigned.

src/components/typeahead/typeahead.component.ts
```diff
@@ -215,6 +255,20 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
 
     optionClickHandler(_event: MouseEvent, option: TypeaheadVisibleOption<T>): void {
         this.select(option, 'mouse');
+
+        if (Array.isArray(this.recentOptions)) {
+            // If the option is already in the array, remove it first.
+            let index = this.recentOptions.indexOf(option.value);
+            if (index != -1) {
```
ashley-hunter: Can we use `!==` as we always want to do value and type checking

src/components/select/select.component.ts
tobias-schreiner: We have a use case where the options can be changed after initial setup. That is when two drop-down lists on the same time share the history. Therefore, the array should not be immutable.

src/components/typeahead/typeahead.component.ts
tobias-schreiner: The value null means that the feature is not used. The empty list means we want to have the history entries, but at the moment, there is no history yet.

src/components/typeahead/typeahead.component.ts
tobias-schreiner: The only time we change the recent options programmatically is in the optionClickHandler. We are using array functions there, though, like unshift() and splice(), which do not call the setter function. Therefore, we still need to call the update function for the recent options observable.

So, at the moment, it does not gain much, but adds complexity. You are right that if someone changes the code later, they will be able to reassign the recent options without taking care of the observable. If they manipulate the array, again, in a different way, it will not help.

I would therefore like to keep it as it is. Please tell me your opinion.

src/components/typeahead/typeahead.component.ts
tobias-schreiner: Done

src/components/typeahead/typeahead.component.ts
tobias-schreiner: Done

src/components/typeahead/typeahead.component.ts
tobias-schreiner: Done

src/components/typeahead/typeahead.component.ts
tobias-schreiner: Done.

src/components/select/select.component.ts
ashley-hunter: The `ReadonlyArray` type doesn't mean that it can't be changed be from the outside, it just means that the array reference needs changed.

So you can still pass it new values at any time, it just makes sure that any time you want to change the array contents you must also change the reference.

Arrays normally have functions like `push` which will change the contents of the array, however if we are to compare the old value with the new value:

```typescript
const oldValue = myArray;
const newValue = myArray;

newValue.push('Some Value');

newValue === oldValue // this will result in true, even though we have pushed to the array
```

Change detection in Angular uses a simple `===` compare to detect changes, so by using `push` to mutate the array the change might not be detected.

If we use a readonly array it enforces us to create a new array each time with the updated content, eg:

```typescript
const oldValue = myArray;
const newValue = [...myArray];

newValue = [...newValue, 'new value'];

newValue === oldValue // results in false
```

Therefore change detection can easily and very performantly detect the change.

src/components/typeahead/typeahead.component.ts
ashley-hunter: The problem you mention about the setter not being called by `unshift` and `splice`, this is another reason for using `ReadonlyArray`. If using a `ReadonlyArray` then we must create a new array with the updated content each time we change it resulting in the setter being called every time we change it.

In fact by changing the array to be readonly we may be able to get rid of the observable all together

src/components/select/select.component.ts
kingmohammas: Done

src/components/typeahead/typeahead.component.ts
kingmohammas: Done

src/components/select/select.component.ts
tobias-schreiner: Thank you for your detailed explanations!

tobias-schreiner:
> Hover state is broken:
> Also doesn't work for multiple select.

Both are fixed.

ashley-hunter:
Thanks for the fixes, I had a few small suggestions, I don't have push permission to your branch, but you can see the diff here:

https://github.com/microfocusbbn/UXAspects/compare/master...UXAspects:suggested-changes

I also noticed a few small issues:

1. Keyboard Navigation does not work to access the recent items
2. On multiple select when I select new recent items they all appear disabled:

image

While I suppose this is technically correct, is this desired? I means the user is further from actionable items. Perhaps we should only show recent items that aren't selected? Any thoughts?

donna-marijne:
Yes, I think that selected options should be removed from the recent list in both single and multiple select modes. I assumed that this was the point of passing in both an array and a separate limit.

tobias-schreiner:
> I had a few small suggestions

I have incorporated most of them. I have a few questions:

1. We did not use EventEmitter<ReadonlyArray<T>>, since IntelliJ complained ReadonlyArray was not a class. Are you sure this does not create any issues?

2. We used recentOptionsWrapped$ instead of recentOptions$, since the data type is different (see the map() call in the setter), and that should also show in the name. Do you have a suggestion for a better name than "-Wrapper"?

3. What is the status of the elvis operator in typescript? I thought is was not yet supported.

4. The round brackets around the async operator makes it more readable for me - is there a reason to avoid them?

>     1. Keyboard Navigation does not work to access the recent items

We will take a look.

>     2. On multiple select when I select new recent items they all appear disabled:
> 
> While I suppose this is technically correct, is this desired? I means the user is further from actionable items. Perhaps we should only show recent items that aren't selected? Any thoughts?

We noticed this and where not sure how to handle it, either. The only option is to remove the disabled options, which shortens the list below the maximum count, even if more non-disabled options had been selected before - they have just dropped off the list. That could be perceived as a defect by the user as well.

Example: Select A, B, C, D, E, and remove them again. Now the recent options show A-E. Select f. In the current implementation, you would see a disabled F and clickable B-E. In the alternative implementation, the recent options would just show B-E and the user will ask: where is A gone?

What do you think would be the lesser evil?

By the way, because of this effect, I would expect that most multiple selection use cases would not enable the recent options feature.

ashley-hunter:
Hi Tobias,

Thanks for the feedback, here are my responses to your questions:

1. Yes, this is fine, we do this in many components, and the type `ReadonlyArray` so this is more correct than a standard array. In terms of runtime execution a `ReadonlyArray` is a normal javascript array anyway, readonly is only for TypeScript to enforce immutability.

2. Yeah, I mainly thought `wrapped` didn't add any value as a description as it didn't make it any clearer what the difference was, in which case we might as well just omit it. If you can think of a better name then by all means rename it, if not then `recentOptions$` is fine.

3. You are correct, the elvis operator is being introduced in TypeScript 3.7 (we target 3.6 or lower currently). However, in Angular templates you can use it as the Angular compiler transforms it into the appropriate expression, so it is not actually the TypeScript implementation being used here.

4. I mainly just removed them as they are unnecessary unless we are directly interacting with the unwrapped variable, which we are not in the case of the `ngFor`. On first look when I see brackets in a template expression it generally indicates that they are there because a certain part of the expression needs evaluated before other parts

tobias-schreiner:
Hi Ashley,

thank you for your explanations! I learn a lot from working with Donna and you.

I have implemented the keyboard navigation now.

Are there any remaining open points?

donna-marijne:
FYI upstream master is now up to date with our 1.8.9 release, so you might want to merge that in to your branch.

tobias-schreiner:
I did that one hour ago. But thank you for the hint.

tobias-schreiner:
The tests are done, please take a look.

e2e/tests/components/select/standard/select.e2e-spec.ts
```diff
@@ -561,4 +561,71 @@ describe('Select Tests', () => {
         expect(await imageCompare('select-custom-icon-multiple-clear-btn')).toEqual(0);
     });
 
+    async function checkRecentOptions(multi: boolean, expectedOptions: string[]) {
```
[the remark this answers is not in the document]
tobias-schreiner: Done.

tobias-schreiner:
> ... the highlight effect has a transition on it in the recent items ...
> The standard items do not have this highlight transition ...

The standard items actually do have the same effect, but only the currently selected one. There is a this rule in navigation.less:

.nav-header,
li.active {
    transition: all 0.5s;
}

which is responsible for this effect.

I can simply suppress all transitions for all the typeahead list items, if you want, but this CSS rule could also cause trouble in other areas. Do you want to investigate some more and figure out a better solution? Or I will do just the quick fix.

ashley-hunter:
@donna-marijne That looks like an old AngularJS style? Perhaps the safest way to handle this is just to disable transitions for all select items? Any thoughts?

ashley-hunter:
Also I notice there is a conflict that will need resolved

donna-marijne:
Yes, that rule has been removed for 2.0. It certainly should not be applied to ux-select. I'd suggest overriding the transition in the select stylesheet.

tobias-schreiner:
I merged and fixed the transition defect. Please take another look.

donna-marijne:
1.8.10 will be released today, so this will be deferred to the `develop-1.8.11` branch. Thanks all for your collaboration.

PR #1245, opened 2019-11-18
EL-3612 - contributing

donna-marijne:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3612

* Updated CONTRIBUTING.md with up to date details including using forks, making a PR, and a full list of the build scripts.
* Updated Getting Started to remove unsupported techniques. AngularJS is no longer mentioned since noone should be starting an AngularJS app at this point. Angular JS information is still available at https://uxaspects.github.io/UXAspects-ng1/#/gettingstarted.
* Updated the Developer Standard with some more guidance on documentation. Also moved information about writing automated tests here, and removed the old Protractor guide.
* Updated peer dependencies to exclude Angular 9 when it comes out until we verify compatibility.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3612-contributing


docs/app/pages/getting-started/getting-started.component.html
```diff
@@ -1,104 +1,132 @@
...
+        <h2 id="add-to-an-existing-project">Add to an Existing Project</h2>
+
+        <p>If you want to consume UX Aspects in an existing application follow the steps below:</p>
+
+        <h3>Install UX Aspects</h3>
+
+        <pre>npm install --save @ux-aspects/ux-aspects</pre>
+
+        <p>
+            Angular 6 applications should also install <code>rxjs-compat</code>.
```
ashley-hunter: Just noticed this should also be changed. `rxjs-compat` is required all all versions of Angular we support (until we remove our dependency on `angular-split`

donna-marijne: Should it then be a dependency of the library, or a peer dependency?

ashley-hunter: It would need to be a peer dependency as the `rxjs-compat` version would need to match the version of `rxjs` that is installed, which would vary based on the version of Angular that the application is using

donna-marijne:
I've added rxjs-compat to the peerDependencies and removed the note about Angular 6. Also added some OnPush info in the dev standard.

PR #1256, opened 2019-11-28
EL-3705 - date range picker localization

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3705

- Added inputs to documentation for date/time localization
- custom pipe created for the time date/time localization
- Added an e2e test
- Updated screenshots

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3705-date-range-picker-localization


docs/app/pages/components/components-sections/date-time-picker/date-range-picker/date-range-picker.component.html
```diff
@@ -172,6 +172,12 @@
     <tr uxd-api-property name="endPickerAriaLabel" type="string" defaultValue="Selecting the end date">
         Defines the aria-live announcement when a date in the end picker is focused.
     </tr>
+    <tr uxd-api-property name="dateLocalization" type="string | Function">
```
ashley-hunter: Could be be more specific with the `Function` type, so instead do: `(date: Date) => string` or whatever it should be

docs/app/pages/components/components-sections/date-time-picker/date-range-picker/date-range-picker.component.html
```diff
@@ -172,6 +172,12 @@
...
+        Allows localization of the start/end date strings.
+    </tr>
+    <tr uxd-api-property name="timeLocalization" type="string">
```
ashley-hunter: Can't this also be a function. We should maybe also tell them what format the string can be? Perhaps even link to the Angular Date Pipe formats section?

e2e/pages/app/date-range-picker/date-range-picker.testpage.component.ts
```diff
@@ -44,6 +44,12 @@ export class DateRangePickerTestPageComponent {
     /** Indicate if the show now should be visible */
     showNowBtn: boolean = false;
 
+    /** Time localization */
+    timeLocalization: string;
```
ashley-hunter: Should we also add a test where we test the functions?

src/components/date-range-picker/date-range-picker.component.html
```diff
@@ -2,9 +2,21 @@
 
   <div class="header-section">
     <div *ngIf="!rangeService.start" class="select-header">{{ selectStartTitle }}</div>
-    <div *ngIf="rangeService.start" class="date-header">{{ rangeService.start | date: 'd MMMM y' }}</div>
+      <div *ngIf="rangeService.start" class="date-header">
+          <ng-container *ngIf="!dateLocalization">
```
ashley-hunter: I think we can get rid of these `ngIf`s, instead we could just set the default `dateLocalization` value in the typescript to be:

```
dateLocalization = 'd MMMM y';
```
And then we only need to have:

```html
{{ rangeService.start | dateLocale: dateLocalization }}
```

As it handles dates strings anyway

src/components/date-range-picker/date-range-picker.component.html
```diff
@@ -2,9 +2,21 @@
...
+              {{ rangeService.start | date: 'd MMMM y' }}
+          </ng-container>
+          <ng-container *ngIf="dateLocalization">
+              {{ rangeService.start | dateLocale: dateLocalization }}
+          </ng-container>
+      </div>
     <div [style.visibility]="rangeService.start && showTime ? 'visible' : 'hidden'" class="time-header">
-        {{ rangeService.start | date: showMeridian ? 'shortTime' : 'HH:mm' }}
+        <ng-container *ngIf="!timeLocalization">
```
ashley-hunter: We can also probably remove the `ngIf` here too, eg:

```html
{{ rangeService.start | dateLocale: timeLocalization || showMeridian ? 'shortTime' : 'HH:mm' }}
```

donna-marijne:
Do you think it would be appropriate to use `toLocaleDateString` and `toLocaleTimeString` as defaults instead of the current fixed strings?

docs/app/pages/components/components-sections/date-time-picker/date-range-picker/date-range-picker.component.html
```diff
@@ -172,6 +172,12 @@
...
+        DAllows localization of the start/end time strings.
```
donna-marijne: Typo here

src/components/date-range-picker/date-localization.pipe.ts
```diff
@@ -0,0 +1,16 @@
+import { Inject, LOCALE_ID, Pipe, PipeTransform } from '@angular/core';
+import { formatDate } from '@angular/common';
+import { DateLocalizationFormatter } from './date-localization-formatter.type';
+
+@Pipe({
+    name: 'dateLocale'
```
donna-marijne: This should probably be something like `formatDate` or `formattedDate`, i.e. describing the outcome of the pipe. Also we have a separate subdirectory for custom pipe components under `src` which could be useful in case we need this functionality in other date-based components in the future.

src/components/date-range-picker/date-range-picker.component.html
donna-marijne: Indentation is inconsistent here. Looks like the file is using 2 spaces, so you could reformat the rest to match the 4 spaces that you have used.

src/components/date-range-picker/date-localization.pipe.ts
```diff
@@ -0,0 +1,16 @@
...
+})
+export class DateLocalizationPipe implements PipeTransform {
+
+    constructor(@Inject(LOCALE_ID) private locale: string) {}
```
donna-marijne: @ashley-hunter Does this depend on the consumer using Angular i18n? In the documentation site  on my environment this has the value "en-US" which is apparently the default if not configured.

TaraTodd123:
I tired to use toLocaleDateString as the defaults but it changed the current appearance of the dates to 
image

src/components/date-range-picker/date-localization.pipe.ts
ashley-hunter: There is a `navigator.language` property available in the browser perhaps we should use instead (this is fairly new so I'm not sure if it works in IE 11 or not). But we could use the `LOCALE_ID` as fallback if IE is not supported

donna-marijne:
Yes I see, let's just leave it as it is then. Thanks.

src/components/date-range-picker/date-localization.pipe.ts
donna-marijne: Yes that sounds good. If that's not suitable then they still have the option of providing a function.

TaraTodd123:
Changes for the above all made

TaraTodd123:
screenshots now updated for build to pass

PR #1262, opened 2019-12-10
EL-3715-FormTouched

JonathanToddMF:
#### Checklist

* [X] The contents of this PR are mine to submit. No third party code or other material is being committed.

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3715

Marking the component as touched when using ngModel when the user clicks on the component. Also added an event for the multiple option by emitting and event to the parent component. 

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3715-FormTouched

### DIFF
--- a/src/components/select/select.component.spec.ts
+++ b/src/components/select/select.component.spec.ts
@@ -362,26 +362,28 @@ describe('Select Component - NgModel Input', () => {
         expect(component.onValueChange).not.toHaveBeenCalled();
     });
 
-    it('should change from untouched to touched when clicking the input in single mode', async () => {
+    fit('should change from untouched to touched when clicking the input in single mode', async () => {
         component.multiple = false;
         component.value = component.options[0];
-        fixture.autoDetectChanges();
+        fixture.detectChanges();
         await fixture.whenStable();
         expect(nativeElement.querySelector('ux-select').classList).toContain('ng-untouched');
 
         getSelect(component.multiple).click();
+        fixture.detectChanges();
         expect(nativeElement.querySelector('ux-select').classList).toContain('ng-touched');
 
     });
 
-    it('should change from untouched to touched when clicking the input in multiple mode', async () => {
+    fit('should change from untouched to touched when clicking the input in multiple mode', async () => {
         component.multiple = true;
         component.value = [component.options[0]];
-        fixture.autoDetectChanges();
+        fixture.detectChanges();
         await fixture.whenStable();
         expect(nativeElement.querySelector('ux-select').classList).toContain('ng-untouched');
 
         getSelect(component.multiple).click();
+        fixture.detectChanges();
         expect(nativeElement.querySelector('ux-select').classList).toContain('ng-touched');
 
     });
--- a/src/components/select/select.component.spec.ts
+++ b/src/components/select/select.component.spec.ts
@@ -1,7 +1,7 @@
 import { Component } from '@angular/core';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
-import { SelectModule } from './select.module';
 import { FormControl, FormGroup, FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { SelectModule } from './select.module';
 
 @Component({
     selector: 'app-select-test',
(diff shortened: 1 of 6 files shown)


src/components/tag-input/tag-input.component.html
```diff
@@ -31,7 +31,7 @@
                [placeholder]="disabled ? '' : (placeholder || '')"
                [disabled]="disabled"
                [focusIf]="isSelected(tags.length)"
-               (click)="toggle(); $event.stopPropagation()"
+               (click)="toggle(); $event.stopPropagation(); onTouched.emit()"
```
ashley-hunter: I don't think this will work if we navigate to the component using the keyboard. I think it might be more useful to have an `inputFocus` event that fires when the input gets focused, would it work in there?

JonathanToddMF: I added the onTouched.emit() to the (focus) input instead of the click

src/components/tag-input/tag-input.component.ts
```diff
@@ -192,6 +192,9 @@ export class TagInputComponent<T = any> implements AfterContentInit, OnChanges,
     /** Raised when a tag has been clicked. The `tag` property of the event contains the clicked tag. Call `preventDefault()` on the event to prevent the default behaviour of selecting the tag. */
     @Output() tagClick = new EventEmitter<TagInputEvent>();
 
+    // When clicking on the input during mutliple mode it will send a on touched event to the parent component
+    @Output() onTouched = new EventEmitter<TagInputEvent>();
```
ashley-hunter: It might be good to make this output called `inputFocus` instead. The `tag-input` component is a component that an app could use on it's own (not in the select) so `onTouched` is not really something that normally gets an event whereas the input focus could be something an app could be interested in.

As a side note, for naming outputs we generally want to avoid prefixing them with `on` (see Angular Styleguide https://angular.io/guide/styleguide#dont-prefix-output-properties).

We also try to avoid using `-ed` on the end of event names, for example if you look at the built in browser events Angular provides, the are `(click)`, `(focus)`, `(blur)` instead of `(clicked)`, `(focused)`, `(blurred)`

ashley-hunter: It may also be good to add the `inputFocus` event to the doc? https://uxaspects.github.io/UXAspects/#/components/input-controls#tags

src/components/tag-input/tag-input.component.html
```diff
@@ -32,7 +32,7 @@
                [disabled]="disabled"
                [focusIf]="isSelected(tags.length)"
                (click)="toggle(); $event.stopPropagation()"
-               (focus)="inputFocusHandler()"
+               (focus)="inputFocusHandler(); inputFocus.emit()"
```
ashley-hunter: It would be good here to emit the native focus event using `$event`

src/components/tag-input/tag-input.component.ts
```diff
@@ -192,6 +192,9 @@ export class TagInputComponent<T = any> implements AfterContentInit, OnChanges,
     /** Raised when a tag has been clicked. The `tag` property of the event contains the clicked tag. Call `preventDefault()` on the event to prevent the default behaviour of selecting the tag. */
     @Output() tagClick = new EventEmitter<TagInputEvent>();
 
+    // When clicking on the input during mutliple mode it will send a on touched event to the parent component
+    @Output() inputFocus = new EventEmitter<TagInputEvent>();
```
ashley-hunter: Currently we are not emitting anything, so `TagInputEvent` would be incorrect, this would be `void`, however with my previous comment of emitting the native focus event we should have a type of `FocusEvent` I believe.

The docs would also need updated.

JonathanToddMF: Thinks I've updated this correctly, let me know if there is any further changes!

JonathanToddMF:
> Thanks for making those changes, just spotted what I think is an unused function

Whoops, removed

src/components/select/select.component.spec.ts
```diff
@@ -361,6 +361,34 @@ describe('Select Component - NgModel Input', () => {
 
         expect(component.onValueChange).not.toHaveBeenCalled();
     });
+
+    it('should change from untouched to touched when clicking the input in single mode', async () => {
+        component.multiple = false;
+        component.value = component.options[0];
+        fixture.autoDetectChanges();
```
ashley-hunter: `autoDetectChanges` is designed to configure how change detection should be run rather than actually running change detection (it does trigger change detection which is why it is working, however this is not what is should be used for, we have `fixture.detectChanges` for that instead). If your tests need this enabled then we should add this to the `beforeEach`, eg:

image

Then in our tests we should be able to use:

`fixture.detectChanges()` as this is the function that actually does the change detection., eg:

image

PR #1267, opened 2020-01-13
El 3707 media player

JonathanToddMF:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3707

Added input to overwrite filename of audio file
If service detects base64 string it will hide the filename
updated docs and added tests
Ashley authored some changes to prevent a change detection error preventing the Karma tests from working

#### Documentation CI URL
https://jenkins.swinfra.net/job/SEPG/job/UXAspects/view/Developer/job/UXAspects~UXAspects~EL-3707-MediaPlayer~CI/

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3707-MediaPlayer/

#### Downstream Build(s)
https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3707-MediaPlayer~CI/


docs/app/pages/components/components-sections/media-player/media-player/media-player.component.html
```diff
@@ -46,6 +52,9 @@
     <tr uxd-api-property name="source" type="string">
         The url to the media file to be loaded by the media player.
     </tr>
+    <tr uxd-api-property name="displayName" type="string">
```
ashley-hunter: @donna-marijne Would this be better as `filename`? Or do you think `displayName` describes what this is better?

I think the description could also be re-worded to follow the phrasing we use in our other component inputs. It might also be worth mentioning that this only applies to audio files? And that by default the file name is retrieved from the source file?

docs/app/pages/components/components-sections/media-player/media-player/media-player.component.html
```diff
@@ -23,6 +22,13 @@
                         <ux-radio-button [(value)]="mode" option="quiet" [disabled]="type === 'audio'">Quiet</ux-radio-button>
                     </div>
                 </div>
+
+                <div class="row uxd-customize-row">
```
ashley-hunter: I'm not sure this requires an customize input as by default it will display the filename. This is only really relevant if we are using a base64 file as the `src`.

docs/app/pages/components/components-sections/media-player/media-player/media-player.component.ts
```diff
@@ -26,7 +27,7 @@ export class ComponentsMediaPlayerComponent extends BaseDocumentationSection imp
             'app.component.css': this.snippets.raw.appCss
         },
         modules: [{
-            imports: ['RadioButtonModule', 'MediaPlayerModule', 'AccordionModule'],
+            imports: ['RadioButtonModule', 'MediaPlayerModule', 'AccordionModule', 'SelectModule', 'FormsModule', 'ReactiveFormsModule'],
```
ashley-hunter: Are these additional imports required? None of the changes to the template use a select component or reactive forms?

src/components/media-player/media-player.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+    mode: string = 'standard';
+    displayName: string = null;
+
+    audioSource = '';
+
+    constructor() {
+    }
+}
+
+fdescribe('Media Player Component - Audio', () => {
```
ashley-hunter: Remove `f` as this will only run these tests skipping all others

src/components/media-player/media-player.component.spec.ts
```diff
@@ -0,0 +1,76 @@
+import { Component } from '@angular/core';
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { MediaPlayerModule } from './media-player.module';
+
+@Component({
+    selector: 'ux-media-player-audio',
+    template: `<ux-media-player *ngIf="type === 'audio'" class="media-player" type="audio" [displayName]="displayName" [source]="audioSource"></ux-media-player>
```
ashley-hunter: I don't think we need the `ngIf` here? also we have a `type` and `mode` variable in the component class which is never used.

src/components/media-player/media-player.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+        fixture.detectChanges();
+
+        setTimeout(() => {
+            fixture.detectChanges();
+            expect(getTitle().innerText).toBe('Ocean-Waves.mp3');
+            done();
+        }, 2000);
+    });
+
+    function getTitle(): any {
```
ashley-hunter: Can we give this a type, `any` should only be used as a last resort.

src/components/media-player/media-player.component.ts
```diff
@@ -35,6 +35,9 @@ export class MediaPlayerComponent implements AfterViewInit, OnDestroy {
     /** The `anonymous` keyword means that there will be no exchange of user credentials when the media source is fetched. */
     @Input() crossorigin: 'use-credentials' | 'anonymous' = 'use-credentials';
 
+    @Input()
```
ashley-hunter: Can we add a comment above the input describing the purpose. This allows editors to show documentation about the input without requiring the user to view the docs site.

src/services/audio/audio.service.ts
```diff
@@ -17,10 +17,12 @@ export class AudioService {
...
             this._http.get(mediaElement.src, { responseType: 'blob' }).subscribe(response => {
 
-                const filename = mediaElement.src.substring(mediaElement.src.lastIndexOf('/') + 1);
                 const extension = mediaElement.src.substring(mediaElement.src.lastIndexOf('.') + 1).toLowerCase();
-
+                let filename;
                 let description;
+                let sourceString: string = mediaElement.src;
+
+                sourceString.includes('base64') ? filename = '' : filename = sourceString.substring(mediaElement.src.lastIndexOf('/') + 1);
```
ashley-hunter: This can be simplified, we shouldn't need to duplicate the `filename = ` part and can get rid of `sourceString`.
We should also avoid using `includes` as this in not supported in all browsers and is one of the functions that is not always polyfilled by default:

```typescript
const filename = mediaElement.src.indexOf('base64') !== -1 ? '' : mediaElement.src.substring(mediaElement.src.lastIndexOf('/') + 1)
```

docs/app/pages/components/components-sections/media-player/media-player/media-player.component.html
donna-marijne: Yes, either `filename` or `title` would be fine.

ashley-hunter: Great, I think it is best to avoid `title` as if someone uses this as a string attribute this will also add a browser tooltip to the element.

donna-marijne: Good point!

docs/app/pages/components/components-sections/media-player/media-player/media-player.component.html
donna-marijne: Yes, there should be no change to the documentation example for this ticket, since it uses a named file.

src/components/media-player/extensions/controls/controls.component.ts
```diff
@@ -45,8 +45,8 @@ export class MediaPlayerControlsExtensionComponent extends MediaPlayerBaseExtens
         }
     };
 
-    private _volume: number = 50;
-    private _previousVolume = 50;
+    private _volume: number = 100;
+    private _previousVolume = 100;
```
donna-marijne: I'm not clear why this was changed. Please update the "proposed changes" section in the PR if this was intentional.

src/components/media-player/media-player.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+        component.displayName = '';
+        fixture.detectChanges();
+        expect(getTitle().innerText).toBeFalsy();
+    });
+
+    it('should display displayName when one is added to the markup', () => {
+        component.displayName = 'Display Name';
+        fixture.detectChanges();
+        expect(getTitle().innerText).toBeTruthy();
```
donna-marijne: Shouldn't this be something like `expect(getTitle().innerText).toBe('Display Name')`?

src/components/media-player/media-player.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+        setTimeout(() => {
+            fixture.detectChanges();
+            expect(getTitle().innerText).toBe('Ocean-Waves.mp3');
+            done();
+        }, 2000);
+    });
+
+    function getTitle(): any {
+        return nativeElement.querySelector('.player-container .audio-player .audio-file-name');
```
donna-marijne: This should probably return `innerText` rather than appending it to each function call.

src/components/media-player/media-player.component.spec.ts
```diff
@@ -0,0 +1,76 @@
...
...e9d3rJlADvHSxK6Mpo0EYgjT+lN4DLlZ4t1YUZJwBdgogOevvBVmv29EsXHcYEAIb4qOJcu5TvtVOAaYiBALGE0ykWcDNs+VALJFMcp9F1Fk+wwJmi2BpdTNSsUUrBItgBxJ6CAOCx8ID52yQZjpnXdV0cwlCG2DCmqaY9hnRcJ6cWSr8wmTurPgl89cRIzKBFcO739eUTHFGDJuXU77q8qhII6hA6APgEAp6Qo7xFV495hVEj1zVWI5KLtbUbJY/jlQdd0/cyGKjgGIbvhVNcCwBs+VDDUEANE7nG1o6SDAnUMLFTeVERQTLhCAcQMpQPwmGZLNoaoa4IXSj0kbpjcmOkR7ggzG8dydAbPgixyHjNx8yaACT9fEbDNTnZ7ZPPa0FWTGfW00NVR9TbZtbUtiluzGZGAQTQv7AkEoSmiIebhzhAIABlEHq60XRvRYQoY0q0EDOtDqfMqPamA79tpsPWCccc88ehQ6ti2Nk4cGTnt8UUtbWO9j5WHEdmHh+7C9wIRcbudjS9CXjM7kaocWm3CfsmIKfZpvtHUVaURZ6uy8pxxLmzR5SEHNsXspeo+DDzw4x9MaFSHm6tIyWiFempO9P2dtOMiWnTObTzx19O+Msvi/htb6NqsZ/O3s7eMiJWj9dTFq5AEM9w6Lh/X076mm6aA9sCSIoJAHo13PxwXzBABvcU8c/u7iQHUUTgzNM0Wo1dlmJRzDKZSAAAAAAwJwLO5a/4yl+5v++H209u/94v13Pn59nDcvtSL21xLUpndfvy7VA5//w4f27LOL795PYo+f7974+SXupFFhk591zTOL79eH98nts4eItR07Oe9axnPeuZE0CZ27tPW1u3OzOKit91rOe+1K7jFmVbZRyPx0AxaDGNyYnJp/7+eXL7yaB/9H/Pku1wvejpgenJsPvz+/+mrSgjkevxerzVlbPImoxrYbKeY8BJsnrRY89KMmG12IoUvs3b7eanpycFiVHdouYWg8lsNmt1FS+33+9zIiFVOq5jbU9nZ1MABYA2AQAAAAAAg0MAAAkAAACfd/skAWoWAKqz3++H+/T0pGA6hrVD9XUrrc4zc3Nzc/r9sOZyIGq6bDnpYwKensDQ9y/K37+cAFtbW3TudhWOtwZzlstb/X5/a6vf72/x5+fm5nB6slBlZ3Fcha363d5ut7u3ni1rLoPf728l3KcK';
```
donna-marijne: Try this to avoid adding a huge string to this file: https://github.com/mathiasbynens/small

src/components/media-player/media-player.service.ts
```diff
@@ -138,7 +138,7 @@ export class MediaPlayerService {
     }
 
     get duration(): number {
-        return this._mediaPlayer ? this._mediaPlayer.duration : 0;
+        return this._mediaPlayer && !isNaN(this.mediaPlayer.duration) ? this._mediaPlayer.duration : 0;
```
donna-marijne: Is this change related to this ticket?

src/components/media-player/media-player.component.spec.ts
JonathanToddMF: Thank you for this, I had a hard time finding such a small audio file!

src/components/media-player/media-player.service.ts
ashley-hunter: There was an expression changed error happening, this was to prevent it I believe

src/components/media-player/extensions/controls/controls.component.ts
ashley-hunter: The default browser volume is 100 so by having 50 was causing an expression changed error. But as Donna mentioned it is good to put things like this in the Proposed Changes section

src/components/media-player/media-player.component.spec.ts
```diff
@@ -4,23 +4,20 @@ import { MediaPlayerModule } from './media-player.module';
 
 @Component({
     selector: 'ux-media-player-audio',
-    template: `<ux-media-player *ngIf="type === 'audio'" class="media-player" type="audio" [displayName]="displayName" [source]="audioSource"></ux-media-player>
+    template: `<ux-media-player class="media-player" [type]="'audio'" [filename]="filename" [source]="audioSource"></ux-media-player>
```
ashley-hunter: Whenever the value for an input is a constant string, you can omit the the `[` `]`, so in this instance you can write:

```
type="audio"
```

If you have square brackets then Angular has to check if the property has changed every time change detection is run, if we know the value will never change then providing a string literal mean Angular has one less thing to check.

src/components/media-player/extensions/controls/controls.component.ts
JonathanToddMF: Apologies, there was a ExpressionChangedAfterItHasBeenChecked error when running fixture.detectChanges() in the karma tests. 

These changes authored by @ashley-hunter are to prevent this error from occurring.

src/components/media-player/media-player.service.ts
JonathanToddMF: Yeah there was a ExpressionChangedAfterItHasBeenChecked error when running fixture.detectChanges() in the karma tests.

src/components/media-player/media-player.component.spec.ts
```diff
@@ -46,31 +43,31 @@ fdescribe('Media Player Component - Audio', () => {
...
-        component.displayName = '';
+        component.audioSource = 'data:audio/mpeg;base64,/+MYxAAAAANIAAAAAExBTUUzLjk4LjIAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA';
+        component.filename = '';
         fixture.detectChanges();
-        expect(getTitle().innerText).toBeFalsy();
+        expect(getTitle()).toBeFalsy();
```
ashley-hunter: Probably best to check strings against an actual value rather than true/false. JavaScript has a lot of falsy values, so if the string was '0', this could be reported as falsy, but not what we are expecting the value to be.

src/components/media-player/media-player.component.spec.ts
```diff
@@ -46,31 +43,31 @@ fdescribe('Media Player Component - Audio', () => {
...
-            expect(getTitle().innerText).toBe('Ocean-Waves.mp3');
+            expect(getTitle()).toBe('Ocean-Waves.mp3');
             done();
         }, 2000);
     });
 
-    function getTitle(): any {
-        return nativeElement.querySelector('.player-container .audio-player .audio-file-name');
+    function getTitle(): string {
+        return (<HTMLElement>nativeElement.querySelector('.player-container .audio-player .audio-file-name')).innerText;
```
ashley-hunter: The `querySelector` function accepts a generic to specify it's type, which means we don't need to cast it, for example, this could be:

```typescript
return nativeElement.querySelector<HTMLParagraphElement>('...').innerText;
```

src/services/audio/audio.service.ts
```diff
@@ -17,12 +17,11 @@ export class AudioService {
         return Observable.create((observer: Observer<AudioMetadata>) => {
             this._http.get(mediaElement.src, { responseType: 'blob' }).subscribe(response => {
 
-                const extension = mediaElement.src.substring(mediaElement.src.lastIndexOf('.') + 1).toLowerCase();
-                let filename;
                 let description;
                 let sourceString: string = mediaElement.src;
```
ashley-hunter: Is this variable used?

src/components/media-player/media-player.component.ts
```diff
@@ -35,8 +35,9 @@ export class MediaPlayerComponent implements AfterViewInit, OnDestroy {
     /** The `anonymous` keyword means that there will be no exchange of user credentials when the media source is fetched. */
     @Input() crossorigin: 'use-credentials' | 'anonymous' = 'use-credentials';
 
+    /** Overwrite the filename displayed in the audio meida player */
```
ashley-hunter: Typo

src/components/media-player/extensions/controls/controls.component.ts
donna-marijne: No problem, could you also add repro details to the ticket to make sure that this gets proper QA?

docs/app/pages/components/components-sections/media-player/media-player/media-player.component.html
```diff
@@ -1,13 +1,12 @@
 <ux-media-player #player *ngIf="type === 'video'" class="media-player" type="video" [source]="videoSource" [quietMode]="mode === 'quiet'">
     <track label="English" lang="en" default [src]="subtitles">
 </ux-media-player>
-<ux-media-player *ngIf="type === 'audio'" class="media-player" type="audio" [source]="audioSource"></ux-media-player>
+<ux-media-player *ngIf="type === 'audio'" class="media-player" type="audio"  [source]="audioSource" [filename]="filename"></ux-media-player>
```
ashley-hunter: I think we decided not to include this in the documentation example, in which case we can remove the filename attribute from here and the variable from the component typescript file

src/components/media-player/extensions/controls/controls.component.ts
JonathanToddMF: sure thing

docs/app/pages/components/components-sections/media-player/media-player.module.ts
```diff
@@ -27,6 +28,7 @@ const ROUTES = [
         MediaPlayerModule,
         RadioButtonModule,
         AccordionModule,
+        FormsModule,
```
donna-marijne: Presumably FormsModule no longer needs to be imported at this point.

docs/app/pages/components/components-sections/media-player/media-player/media-player.component.html
```diff
@@ -1,13 +1,12 @@
 <ux-media-player #player *ngIf="type === 'video'" class="media-player" type="video" [source]="videoSource" [quietMode]="mode === 'quiet'">
     <track label="English" lang="en" default [src]="subtitles">
 </ux-media-player>
-<ux-media-player *ngIf="type === 'audio'" class="media-player" type="audio" [source]="audioSource"></ux-media-player>
+<ux-media-player *ngIf="type === 'audio'" class="media-player" type="audio"  [source]="audioSource"></ux-media-player>
```
donna-marijne: Revert added whitespace.

PR #1268, opened 2020-01-15
EL-3737-Wizard-Marquee

cmallon367:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3737

An If statement was implemented so that when the check box has been selected the next button will  grey out and will not respond to clicks.
The documentation and screenshots were updated and e2e testing was completed successfully.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3737-Wizard-Marquee

#### Downstream Build(s)
https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3737-Wizard-Marquee~CI/


docs/app/pages/components/components-sections/wizard/marquee-wizard/marquee-wizard.component.html
```diff
@@ -262,6 +262,10 @@ <h1 class="m-t-nil">Sixth step title</h1>
         If set to
         <code>true</code> the 'Next' button will appear disabled and will not respond to clicks.
     </tr>
+    <tr uxd-api-property name="isDisabled" type="boolean">
```
ashley-hunter: I might be missing something but this doesn't seem to be used?

src/components/marquee-wizard/marquee-wizard.component.ts
```diff
@@ -89,6 +89,23 @@ export class MarqueeWizardComponent extends WizardComponent implements OnDestroy
...
     }
 
+
+    /**
+     * If the validation is true then disable the next button
+     * and do not allow progression onto the next step
+     */
+    isDisabled(): boolean {
+
+         const step = this.getCurrentStep() as MarqueeWizardStepComponent;
```
ashley-hunter: This could probably be simplified to:

```typescript
return !this.getCurrentStep().valid;
```

It might also be good to rename this function something slightly more descriptive. `isDisabled` doesn't give me any indication that this is to do with the next button for example

e2e/pages/app/marquee-wizard/marquee-wizard.testpage.component.html
```diff
@@ -29,6 +29,7 @@
 
 <button id="resizeable" class="btn button-primary" (click)="resizable = true">Enable resizable</button>
 <button id="footerTemplate" class="btn button-secondary" (click)="hasFooterTemplate = true">Add footerTemplate</button>
+<button id="validation" class="btn button-secondary" (click)="validation = true">Add Validation</button>
```
ashley-hunter: "validation" implies that it is enabling checking to see if it is valid or not rather than marking the step as invalid. Could this be changed to something like "Set Step Invalid" and perhaps perhaps we could flip the variable to be `invalid` or `stepInvalid` to make it clear.

donna-marijne:
The PR has not been filled correctly. Could you fix the checklist and re-add the Downstream Build(s) section with the appropriate information? If you need help, let me know.

docs/app/pages/components/components-sections/wizard/marquee-wizard/marquee-wizard.component.html
```diff
@@ -262,6 +262,10 @@ <h1 class="m-t-nil">Sixth step title</h1>
         If set to
         <code>true</code> the 'Next' button will appear disabled and will not respond to clicks.
     </tr>
+    <tr uxd-api-property name="stepInvalid" type="boolean">
```
ashley-hunter: It looks like this may be the wrong property you are documenting, your test is using `nextInvalid` as the input.

It might be good to update the description to be something like:

"If set to `true` the 'Next' button will become disabled when the current step is invalid."

src/components/marquee-wizard/marquee-wizard.component.ts
```diff
@@ -89,6 +92,17 @@ export class MarqueeWizardComponent extends WizardComponent implements OnDestroy
         }
     }
 
+
+    /**
+     * If the validation is true then disable the next button
+     * and do not allow progression onto the next step
+     */
+    stepDisabled(): boolean {
```
ashley-hunter: Perhaps `isNextDisabled()` might be a more self explanatory name, just to clarify it is the next button that is being checked not the step?

Also could you review the whitespace here, we could remove the empty line and the double space after the `&&`.

Consider adding the following to your VS Code settings:

image

image

```json
"[typescript]": {
        "editor.formatOnSave": true,
        "editor.codeActionsOnSave": {
            "source.organizeImports": true
        },
        "editor.defaultFormatter": "vscode.typescript-language-features"
    },
```

This will automatically format your code correctly when you save a file.

src/components/marquee-wizard/marquee-wizard.component.ts
```diff
@@ -28,6 +28,9 @@ export class MarqueeWizardComponent extends WizardComponent implements OnDestroy
     /** If set to true the resizable splitter will be enabled and set to the default width **/
     @Input() resizable: boolean = false;
 
+    /** If set to true the nextinvalid will be enabled and current behavior will enable async validation. **/
```
ashley-hunter: Consider using the same description you write in the docs for the property here. This is designed to be documentation for consumers without having to leave their editor, so it's good if the description here is on par with the docs

docs/app/pages/components/components-sections/wizard/marquee-wizard/marquee-wizard.component.html
```diff
@@ -262,6 +262,10 @@ <h1 class="m-t-nil">Sixth step title</h1>
         If set to
```
ashley-hunter: @donna-marijne should this change be moved to the subclass (WizardComponent) to make this option available to both the wizard and the marquee wizard?

donna-marijne: Yes, I believe so. Good idea.

cmallon367: Just to confirm do you want me to move the out of the marquee wizard and into the wizard component?

donna-marijne: Yes, push it up to the base class so that it's available in both components.
```ts
export class MarqueeWizardComponent extends WizardComponent implements OnDestroy {
```
Update documentation and tests for both.

cmallon367: No problem

donna-marijne:
Could you fix the PR please?

src/components/marquee-wizard/marquee-wizard.component.ts
```diff
@@ -28,6 +28,9 @@ export class MarqueeWizardComponent extends WizardComponent implements OnDestroy
     /** If set to true the resizable splitter will be enabled and set to the default width **/
     @Input() resizable: boolean = false;
 
+    /** If set to true the 'Next' button will become disabled when the current step is invalid **/
+    @Input() disableNextWhenInvalid: boolean = false;
```
donna-marijne: Don't think this is needed because it's defined in the superclass.

src/components/marquee-wizard/marquee-wizard.component.ts
```diff
@@ -89,6 +92,20 @@ export class MarqueeWizardComponent extends WizardComponent implements OnDestroy
...
 
+
+    /**
+     * If the validation is false then disable the next button
+     * and do not allow progression onto the next step however
+     * if the validation is set to true then the next button
+     * disable the next button and give a greyed out
+     * appearance as an indicator
+     */
+    isNextDisabled(): boolean {
```
donna-marijne: As above, this is defined in the superclass already.

src/components/wizard/wizard.component.ts
```diff
@@ -69,6 +69,9 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
     /** If set to `false` the 'Next' button will be hidden. */
     @Input() nextVisible: boolean = true;
 
+      /** If set to true the 'Next' button will become disabled when the current step is invalid **/
+      @Input() disableNextWhenInvalid: boolean = false;
```
donna-marijne: Wrong indentation.

src/components/wizard/wizard.component.ts
```diff
@@ -193,6 +196,18 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
 
+     /**
+     * If the validation is false then disable the next button
+     * and do not allow progression onto the next step however
+     * if the validation is set to true then the next button
+     * disable the next button and give a greyed out
+     * appearance as an indicator
+     */
+    isNextDisabled(): boolean {
+        const { valid, disableNextWhenInvalid } = this.getCurrentStep() as WizardStepComponent;
```
donna-marijne: You don't need to cast the return value, the return type of `getCurrentStep` is `WizardStepComponent`.

docs/app/pages/components/components-sections/wizard/marquee-wizard/marquee-wizard.component.html
```diff
@@ -258,9 +258,9 @@ <h1 class="m-t-nil">Sixth step title</h1>
     <tr uxd-api-property name="finishAriaLabel" type="string">
         Defines the text for the aria label on the 'Finish' button.
     </tr>
-    <tr uxd-api-property name="nextDisabled" type="boolean">
+    <tr uxd-api-property name="disableNextWhenInvalid" type="boolean">
```
donna-marijne: Did you forget to define this on the `WizardComponent`? The recommendation was to use the `disableNextWhenInvalid` value from `WizardComponent`, optionally overridden by the value on `WizardStepComponent`.

e2e/tests/components/marquee-wizard/marquee-wizard.e2e-spec.ts
```diff
@@ -58,7 +58,7 @@ describe('Marquee Wizard Tests', () => {
 
         // only the first one should actually have any content
         page.stepContents.each(async (step, idx) =>
-            expect(await step.$$('*').count()).toBe(idx === 0 ? 1 : 0));
+            expect(await step.$$('*').count()).toBe(idx === 0 ? 6 : 0));
```
donna-marijne: Don't really like the idea of having to update this if we happen to add another element inside the checkbox for example. Can we just assert that it's greater than zero for the first step?

e2e/pages/app/wizard/wizard.testpage.component.html
```diff
@@ -12,8 +13,9 @@
         <p>Step 3 Content</p>
     </ux-wizard-step>
 
-    <ux-wizard-step header="Step 4">
+    <ux-wizard-step header="Step 4" [disableNextWhenInvalid]="disableNextWhenInvalid" [valid]="!stepInvalid">
         <p>Step 4 Content</p>
+        <ux-checkbox [(value)]="stepInvalid" id= "stepInvalid">Set this step to error state</ux-checkbox>
```
donna-marijne: I'd recommend not having two components with the same `id` value. I think these two checkboxes can have separate ids and values, e.g. `step1Invalid`, `step4Invalid`.

Also there's a random extra space in here and above which should be removed.

e2e/tests/components/marquee-wizard/marquee-wizard.e2e-spec.ts
```diff
@@ -85,6 +85,40 @@ describe('Marquee Wizard Tests', () => {
         expect(attr).toBeNull();
     });
 
+    it('should disable the next button when the step is invalid and disableNextWhenInvalid = true', async () => {
+        // enable validation
```
donna-marijne: I think this comment is misleading, you're setting the step to be invalid with the following line.

e2e/tests/components/marquee-wizard/marquee-wizard.e2e-spec.ts
```diff
@@ -85,6 +85,40 @@ describe('Marquee Wizard Tests', () => {
...
+        await page.stepInvalidButton.click();
+        await page.disableNextWhenInvalidButton.click();
+
+        let next: ElementFinder = await page.getNextButton();
+        let attr = await next.getAttribute('disabled');
+        expect(attr).not.toBeNull();
+
+        // enable validation
```
donna-marijne: Also misleading, you are setting the step to be valid by clicking it a second time.

e2e/tests/components/marquee-wizard/marquee-wizard.e2e-spec.ts
```diff
@@ -85,6 +85,40 @@ describe('Marquee Wizard Tests', () => {
         expect(attr).toBeNull();
     });
 
+    it('should disable the next button when the step is invalid and disableNextWhenInvalid = true', async () => {
+        // enable validation
+        await page.stepInvalidButton.click();
+        await page.disableNextWhenInvalidButton.click();
+
+        let next: ElementFinder = await page.getNextButton();
+        let attr = await next.getAttribute('disabled');
```
donna-marijne: Both of these variables can be `const` since they do not get reassigned.

cmallon367:
@donna-marijne Ready for review

docs/app/pages/components/components-sections/wizard/wizard/wizard.component.html
```diff
@@ -78,6 +78,9 @@
     <tr uxd-api-property name="finishAriaLabel" type="string">
         Defines the text for the aria label on the 'Finish' button.
     </tr>
+    <tr uxd-api-property name="disableNextWhenInvalid" type="boolean">
```
ashley-hunter: I could be wrong but looking at the code adding this to the wizard/marquee wizard component doesn't look like it does anything? It looks like it only takes into account this value when it is applied to the step element:

```typescript
isNextDisabled(): boolean {
    const { valid, disableNextWhenInvalid } = this.getCurrentStep();
    return disableNextWhenInvalid && !valid;
}
```

So we should either remove it from the wizard and marquee wizard component and only have it on the step component, or support both, but we would need to make it works when used on the wizard components and we would need to add tests. @donna-marijne do you think it is beneficial to have on both a wizard and step component?

donna-marijne: Yes, I suggested a two-tier approach above, citing the use case of a wizard which contains only one step requiring async validation (i.e. when you hit next it makes an async call to update the valid status). This should not be too complicated to implement, but if you don't think it's worthwhile we can just define it at the step level. The issue with that is potentially repetitive configuration since the default is false (for backward compatibility reasons), but for someone who wants to use this feature they probably want it enabled on either all or most of the wizard steps.

ashley-hunter: That sounds fine to me, if we could then update the code to support both step and wizard level options (step level taking precedence) and also add a test to cover wizard level and a test to cover that step level overrides the wizard level

donna-marijne: So something like:
```ts
disableNextWhenInvalid: boolean | undefined = undefined;
```

And then
```ts
return (step.disableNextWhenInvalid === true || this.disableNextWhenInvalid) && !valid;
```

I'm just freehand typing this but hopefully you get the idea?

donna-marijne: Actually I guess it would be more like
```ts
return (step.disableNextWhenInvalid === undefined ? this.disableNextWhenInvalid : step.disableNextWhenInvalid) && !valid;
```

cmallon367:
Tara is going to complete the rest of the testing.

TaraTodd123:
Updates for tests, running build to get the screenshots now.

src/components/wizard/wizard-step.component.ts
```diff
@@ -14,6 +14,9 @@ export class WizardStepComponent {
     /** The text to be displayed in the wizard step tab. */
     @Input() header: string;
 
+    /** If set to true it will become disabled when the current step is invalid **/
+    @Input() disableNextWhenInvalid: boolean | undefined = undefined;
```
ashley-hunter: I am thinking this might be better not as `undefined` but with a default value of `false`? By default all types are `nullable`/`undefined` in typescript so the `| undefined = undefined` is adding code that is doing the same as the browser will do by default anyway.

That way the value will always be a `boolean` and we maintain existing behavior, and can removed the undefined checks in the `isNextDisabled` function (and same in step component)

donna-marijne: Doesn't this prevent you from doing `[isNextDisabled]="false"` on the step to override the parent value of true?

ashley-hunter: Ah yes, good point actually, scrap the default value of false, but you can still just leave the type as:

```typescript
@Input() disableNextWhenInvalid: boolean;
```

TaraTodd123: Updated

ashley-hunter: Thanks, apologies for the unnecessary change, my misunderstanding

donna-marijne:
To save time I have pushed my final review feedback to the branch:

Documentation updates
5b1998c

* Added `disableNextWhenInvalid` to the Marquee Wizard documentation example.
* Adjusted wording of the documentation.
* Updated the Marquee Wizard snippets which were missing several changes.
* Added missing alt text.

Changed WizardComponent.disableNextWhenInvalid to be boolean
b2031ed

* `disableNextWhenInvalid` is only true/false on WizardComponent.
* Fixed comments.

Reworked e2e tests
7efae5b

* Converted unit tests into more e2e-orientated tests.
* Added test coverage of inheriting `disableNextWhenInvalid` value from the wizard component.
* Added test for backward compatibility.

Let me know what you think.

docs/app/pages/components/components-sections/wizard/marquee-wizard/marquee-wizard.component.html
```diff
@@ -258,10 +259,6 @@ <h1 class="m-t-nil">Sixth step title</h1>
     <tr uxd-api-property name="finishAriaLabel" type="string">
         Defines the text for the aria label on the 'Finish' button.
     </tr>
-    <tr uxd-api-property name="nextDisabled" type="boolean">
```
ashley-hunter: Is this supposed to have been removed? As this is still an Input on the component?

TaraTodd123: I will add in now

TaraTodd123:
Updated the removed input

PR #1269, opened 2020-01-15
EL-3721-Tag-Input-White-Label-Focus-State-For-Tags

andrewstevenson-microfocus:
#### Checklist

#### Ticket / Issue

- The Tag Input component is styled the same for the Micro Focus theme and the White Label theme.
- The White Label Tag Input contains the Micro Focus blue and should be changed.
- https://portal.digitalsafe.net/browse/EL-3721.

- Added Focus Indicator to Tag Input component.

#### Documentation CI URL

[Build Status](https://jenkins.swinfra.net/job/SEPG/job/UXAspects/job/UXAspects~UXAspects~EL-3721-Tag-Input-White-Label-Focus-State-For-Tags~CI/)

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3721-Tag-Input-White-Label-Focus-State-For-Tags

#### Downstream Build(s)

Quantum
[Build Status](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~quantum~EL-3721-Tag-Input-White-Label-Focus-State-For-Tags~CI/)

Quantum UXAspects
[Build Status](https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~quantum-ux-aspects~EL-3721-Tag-Input-White-Label-Focus-State-For-Tags~CI/)

UXAspects MicroFocus
[Build Status](https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3721-Tag-Input-White-Label-Focus-State-For-Tags~CI/)


src/components/tag-input/tag-input.component.html
```diff
@@ -4,7 +4,10 @@
     [class.ux-tag-input-icon-inset]="icon"
     (click)="toggle()">
 
-    <li *ngFor="let tag of tags; let i = index" class="ux-tag"
+    <li uxFocusIndicator
+        mouseFocusIndicator="true"
+        programmaticFocusIndicator="true"
+        *ngFor="let tag of tags; let i = index" class="ux-tag"
```
ashley-hunter: Just a slight formatting suggestion, but here every attribute is on a separate line, with the exception of `class`, could it be moved to its own line to be consistent.

src/components/tag-input/tag-input.component.html
```diff
@@ -4,7 +4,11 @@
     [class.ux-tag-input-icon-inset]="icon"
     (click)="toggle()">
 
-    <li *ngFor="let tag of tags; let i = index" class="ux-tag"
+    <li uxFocusIndicator
+        mouseFocusIndicator="true"
```
ashley-hunter: Is `mouseFocusIndicator` correct here? When I try this clicking on the tag it shows the focus outline which I don't believe it should:

image

andrewstevenson-microfocus: If I dont have mouseFocusIndicator there would be no styling difference between a mouse selected tag and an unselected tag on the WL theme (except when using the keyboard)?

ashley-hunter: In the white label could we not use the `:focus` selector to know when it is clicked and the `ux-focus-indicator-active` class to know if it was focused using the keyboard?

donna-marijne: Andi is correct but I don't remember why we originally kept the purple background focus appearance on click after the accessibility changes. It was meant so that keyboard users could select a tag and hit delete to remove it. Mouse users can click the 'x' button though.

Would it be weird if we didn't change appearance at all when clicking a tag? What would be lost from a UX perspective? Let me know what you both think.

ashley-hunter: I think its fine to show the purple background as hitting backspace when the tag is clicked on allows it to be deleted if I remember correctly.

I was mainly just thinking about the green outline, we could use the standard `:focus` selector to set the background color when clicked (or even the `cdk-mouse-focused` class) and then use the `ux-focus-indicator-active` selector to show the outline? Or am I misunderstanding, I haven't read into all the details about this ticket, it just seems wrong to show the focus outline whenever the user clicks on the tag?

donna-marijne: Yes you are right. I think the problem comes from the fact that we are mixing up the concepts of focus and selection in this component. A selected tag can be thought of in the same way selected piece of text in the input field. However for the tags this is implemented using browser focus.

I think this means that we should perhaps not add the focus ring, and instead just invert colors (white on black/dark gray) in the WL theme.

andrewstevenson-microfocus: You're suggesting using css to apply to focus ring specific to the WL theme Ashley? Its upto you guys what way you think it would look best. I ran it past both Donna and Roland that the focus ring would appear on all themes before I put the initial PR up hence why I went forward with that solution but I can do it using CSS if that would be preferable. I would think that the WL theme should probably show the focus ring when selecting the tag input as that was the initial idea behind the ticket and Rolands chosen design but again, its upto you guys.

ashley-hunter: I haven't been involved in the conversation so I don't know the background to all of this so I'm perhaps talking nonsense and perhaps this is completely the expected behavior.

When I click on a tag right now, I get the green outline. Nowhere else in our UI in any theme do we show an outline when clicking on something, we only show the outline whenever the keyboard is used. Perhaps this is the expected change that it is visible on click, however my concern was this is the opposite of what we do elsewhere.

If it is expected that we show the outline on click then ignore my ramblings, I just wanted to clarify as this is not how we present on clicks.

donna-marijne: I'll check it Roland. I thought the focus ring looked good in this component based on the mockups, but I didn't really think it through in terms of the mouse interaction.

andrewstevenson-microfocus: No worries Donna, let me know what Roland thinks and I can implement whatever he decides on.

donna-marijne: I checked with Roland, in light of Ashley's feedback we'll do away with the focus ring idea, and just go with the following color scheme in the WL stylesheet. Are you okay with that?

donna-marijne: https://plnkr.co/edit/wU9q145lU8peMBalrrY8?p=preview

andrewstevenson-microfocus: No worries at all, i'll implement that design now.

donna-marijne:
Can this PR be closed now?

andrewstevenson-microfocus:
> Can this PR be closed now?

I have committed the changes and am just waiting on the builds to pass before I move the ticket back into 'In Build' for review.

PR #1270, opened 2020-01-15
El 3724 select dropdown

JonathanToddMF:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3724

- Removed one line of code causing issue
- Added in test

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3724-Select-Dropdown

#### Downstream Build(s)
* caf/ux-aspects-micro-focus: https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf%7Eux-aspects-micro-focus%7EEL-3724-Select-Dropdown%7ECI/

### DIFF
--- a/src/components/select/select.component.spec.ts
+++ b/src/components/select/select.component.spec.ts
@@ -407,12 +407,12 @@ describe('Select Component - NgModel Input', () => {
 
         const input = getInput();
         input.focus();
+
+        dispatchKeyboardEvent(input, 'keydown', O, null, 'o');
         component.value = 'O';
-        dispatchKeyboardEvent(input, 'keydown', O);
+        fixture.detectChanges();
 
         await fixture.whenStable();
-
-        fixture.detectChanges();
         expect(getTypeahead()).toBeTruthy();
     });
 
--- a/src/components/select/select.component.spec.ts
+++ b/src/components/select/select.component.spec.ts
@@ -1,4 +1,4 @@
-import { O, TAB } from '@angular/cdk/keycodes';
+import { O, SHIFT, TAB } from '@angular/cdk/keycodes';
 import { Component } from '@angular/core';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { FormControl, FormGroup, FormsModule, ReactiveFormsModule } from '@angular/forms';
@@ -394,7 +394,20 @@ describe('Select Component - NgModel Input', () => {
 
         const input = getInput();
         input.focus();
-        dispatchKeyboardEvent(input, 'keydown', TAB);
+        dispatchKeyboardEvent(input, 'keydown', TAB, null, 'Tab');
+
+        await fixture.whenStable();
+
+        fixture.detectChanges();
+        expect(getTypeahead()).toBeFalsy();
+    });
+
+    it('should not open dropdown when pressing a not printable key such as shift', async () => {
+        fixture.detectChanges();
+
+        const input = getInput();
+        input.focus();
+        dispatchKeyboardEvent(input, 'keydown', SHIFT, null, 'Shift');
 
         await fixture.whenStable();
 
--- a/src/components/select/select.component.spec.ts
+++ b/src/components/select/select.component.spec.ts
@@ -229,6 +229,24 @@ describe('Select Component', () => {
         expect(component.value).toBe('One');
     });
 
+    it('should not open dropdown when tabbing past select', () => {
+        fixture.detectChanges();
+
+        const tabEvent: Event = new KeyboardEvent('keydown', {
+            'code': '9'
+        });
+        for (let i = 0; i < 5; i++) {
+            window.dispatchEvent(tabEvent);
+            fixture.detectChanges();
+        }
+        fixture.detectChanges();
+        expect(getTypeahead()).toBeFalsy();
+    });
+
+    function getTypeahead(): HTMLElement | null {
+        return nativeElement.querySelector('ux-typeahead.open');
+    }
+
     function getClearButton(isMultiple: boolean = false): HTMLElement | null {
         return nativeElement.querySelector(`.${isMultiple ? 'ux-tag-icon' : 'ux-select-icon'}.ux-icon-close`);
     }
--- a/src/components/select/select.component.spec.ts
+++ b/src/components/select/select.component.spec.ts
@@ -1,6 +1,7 @@
 import { Component } from '@angular/core';
 import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { FormControl, FormGroup, FormsModule, ReactiveFormsModule } from '@angular/forms';
+import { dispatchKeyboardEvent } from '../../common/testing/dispatch-event';
 import { SelectModule } from './select.module';
 
 @Component({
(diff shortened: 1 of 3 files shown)


JonathanToddMF:
Not overly happy with my approach for the Jasmine tests for this. If there is any feedback that would be appreciated greatly

ashley-hunter:
In regards to testing, you could take a look under `src/common/testing` we have some utility functions for emitting keyboard events if that is any help.

src/components/select/select.component.spec.ts
```diff
@@ -229,18 +230,16 @@ describe('Select Component', () => {
...
-        fixture.detectChanges();
-
-        const tabEvent: Event = new KeyboardEvent('keydown', {
-            'code': '9'
-        });
-        for (let i = 0; i < 5; i++) {
-            window.dispatchEvent(tabEvent);
+    fit('should not open dropdown when tabbing past select', (done) => {
+        for (let i = 0; i < 10; i++) {
+            dispatchKeyboardEvent(document, 'keydown', 9);
```
ashley-hunter: Whenever you need to use keycodes, the `@angular/cdk` packages exports all the keycodes for us to use so we don't have to use hard coded numbers.

src/components/select/select.component.spec.ts
```diff
@@ -229,6 +231,22 @@ describe('Select Component', () => {
...
+    it('should not open dropdown when tabbing past select', (done) => {
+        for (let i = 0; i < 10; i++) {
+            dispatchKeyboardEvent(document, 'keydown', TAB);
+            fixture.detectChanges();
+        }
+
+        setTimeout(() => {
+            expect(getTypeahead()).toBeFalsy();
+            done();
+        }, 1000);
```
ashley-hunter: Just curious, does this need to be `1000`? I understand we may need to delay slightly, but would a smaller timeout work? eg. `1`, or perhaps something even quicker like `Promise.resolve().then(() => ...)` or `requestAnimationFrame`.

While this seems like a small timeout, the more and more tests we add, and every time we delay by a second that can really increase the total time it takes to run the tests on our already slow build process.

src/components/select/select.component.ts
```diff
@@ -233,7 +233,6 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
             take(1),
             takeUntil(this._onDestroy))
             .subscribe(() => {
-                this.dropdownOpen = true;
```
ashley-hunter: Hmm, I think there is another case to consider here. If I tab to the select input and type 'U' for example, the list will appear.

If I reload the page, and I tab past the select, then using tab return to the select and type 'U' the list does not appear.

src/components/select/select.component.spec.ts
```diff
@@ -229,6 +231,25 @@ describe('Select Component', () => {
         expect(component.value).toBe('One');
     });
 
+    it('should not open dropdown when tabbing past select', (done) => {
+
+        Promise.resolve(dispatchTabs()).then(() => {
```
ashley-hunter: Can you verify that this test is working as expected? If I revert your fix the test still passes, whereas this test should fail without your fix and pass after your fix

JonathanToddMF:
Re-done my initial fix factoring in the requirements from @ashley-hunter. 
I also added in additional test to ensure the functionality works as it does now except when using the Tab and Escape key.

@ashley-hunter I attempted using the promise but still required a small 200 millisecond timeout. Maybe I have done this incorrectly.

src/components/select/select.component.ts
```diff
@@ -324,8 +324,8 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
             event.preventDefault();
         }
 
-        // when the user types and the value is not empty then we should open the dropdown
-        if (event.keyCode !== ESCAPE) {
+        // when the user types and the value is not empty then we should open the dropdown except for the tab and escape keys.
+        if (!(event.keyCode === ESCAPE || event.keyCode === TAB)) {
```
ashley-hunter: That has partially fixed the issue, however if I tab to the select an type 'U' the list appears. If I tab out and tab back in and type 'U' the list does not reappear, eg:

[Image from Gyazo](https://gyazo.com/bdb73b034049047bec3671952700698d)

src/components/select/select.component.spec.ts
```diff
@@ -387,6 +408,54 @@ describe('Select Component - NgModel Input', () => {
 
     });
 
+    fit('should not open dropdown when tabbing past select', (done) => {
```
ashley-hunter: Need to remove these `fit`s

src/components/select/select.component.spec.ts
```diff
@@ -387,6 +408,54 @@ describe('Select Component - NgModel Input', () => {
...
+        fixture.detectChanges();
+
+        function focusInput() {
```
ashley-hunter: `Promise.resolve()` is an alternative to `setTimeout`, it is almost an equivalent of a timeout with a time of 0, only quicker.

So you wouldn't need both. In the case where you need to wait for a period of time, eg: 200ms, there are other approaches to avoid actually having to wait that amount of time.

The first one is `whenStable` which returns a promise that completes when Javascripts task queue is empty: https://angular.io/guide/testing#whenstable

Here is how you could use it in your test for example:
```typescript
it('should open dropdown when entering value', async () => {
        fixture.detectChanges();

        const input = getInput();
        input.focus();
        dispatchKeyboardEvent(input, 'keydown', O);
        component.value = 'O';

        await fixture.whenStable();

        fixture.detectChanges();
        expect(getTypeahead()).toBeTruthy();
    });
```

There are other cases where you want to simulate the passing of time you can use the `fakeAsync` and `tick` testing functions Angular provides: https://angular.io/guide/testing#async-test-with-fakeasync

src/components/select/select.component.ts
```diff
@@ -219,7 +220,10 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
             filter(() => this.allowNull),
             filter(value => !this.multiple && value !== this.getDisplay(this.value)),
             takeUntil(this._onDestroy)
-        ).subscribe(() => this.value = null);
+        ).subscribe(() => {
+            this.value = null
```
ashley-hunter: Missing semi-colon and need to remove all the console.logs

JonathanToddMF: my bad, will try to remember to check this in future

donna-marijne:
I've fixed your pull request. Please take care to populate this correctly in future since it contains valuable information to help reviewers check your code. If anything is unclear, do ask.

ashley-hunter:
Thats a very good point! I never considered that, definitely something we should handle!

JonathanToddMF:
Updated the code to ignore, ctrl, shift, alt, escape, tab and windows/command keys. The reason I haven't used charCode is it is deprecated. This option allows us to add any key. Although I couldn't find anyway to make this shorter.

src/components/select/select.component.ts
```diff
@@ -324,9 +324,10 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
             event.preventDefault();
         }
 
-        // when the user types and the value is not empty then we should open the dropdown
-        if (event.keyCode !== ESCAPE) {
+        // when the user types and the value is not empty then we should open the dropdown except for non printable keys.
+        if (event.ctrlKey === false && event.shiftKey === false && event.altKey === false && event.metaKey === false && !(event.keyCode === TAB) && !(event.keyCode === ESCAPE)) {
```
ashley-hunter: There is an `event.key` property that is available, for printable characters it returns the letter/number, for not printable keys it returns the name, eg: `control` etc.. If the key length is 1 then we can tell if it is a printable key or not

src/components/select/select.component.ts
```diff
@@ -325,7 +325,7 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
         }
 
         // when the user types and the value is not empty then we should open the dropdown except for non printable keys.
-        if (event.ctrlKey === false && event.shiftKey === false && event.altKey === false && event.metaKey === false && !(event.keyCode === TAB) && !(event.keyCode === ESCAPE)) {
+        if (event.key.length <= 1) {
```
ashley-hunter: Is there a case where we would want to open the dropdown etc, when the length is less than 1?

JonathanToddMF: None that I can think of. The drop down will open from pressing the Enter key and arrow keys. These are handled in a separate event.

ashley-hunter: Ok, then could we change this to be `event.key.length === 1` instead? Also the branch is out of date from eh base branch which needs to be resolved:

image

JonathanToddMF: sure thing

JonathanToddMF: Rebased

PR #1273, opened 2020-01-27
El-3749: Date Range

JonathanToddMF:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3749

1. Problem was caused by incorrect parsing of the date in the component.ts file in the documentation. This was not noticed as the below error was preventing this code from being called.
2. An error was being thrown by the if statement. Timezone was being equated to null based on the above match method. This caused a error to be thrown when timezone.length was called. I changed this to null, which fixes the if statement. **Note**, a value for timezone is given if there is a timezone modifier on the string so this will still correctly fire the else below.
3. Had some discussions with @ashley-hunter about whether to use a Karma test for this or e2e. End to end seemed more applicable here as it was a mixture of both the input component and the date picker. 

#### Builds
https://jenkins.swinfra.net/job/SEPG/job/UXAspects/view/Developer/job/UXAspects~UXAspects~EL-3749-DateRange~CI/

#### Downstream Build(s)
caf/ux-aspects-mirco-focus: https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3749-DateRange~CI/

#### Documentation CI URL
Will update once built: https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3749-DateRange


docs/app/pages/components/components-sections/date-time-picker/date-range-picker/date-range-picker.component.ts
```diff
@@ -78,7 +78,7 @@ export class ComponentsDateRangePickerComponent extends BaseDocumentationSection
         // check if the date contains a hyphen
         const parts = (date.indexOf('—') ? date.split('—') : date.split('-'));
         const startDate = Date.parse(parts[0].trim());
-        const endDate = Date.parse(parts[0].trim());
+        const endDate = Date.parse(parts[1].trim());
```
ashley-hunter: Does the code snippet/plunker not also need updated?

JonathanToddMF: Updated now and added as a comment to the original ticket - https://next.plnkr.co/edit/fGOJRoh1RXUfGAAl

src/components/date-range-picker/date-range-picker.spec.ts
```diff
@@ -0,0 +1,73 @@
...
+import { FormsModule } from '@angular/forms';
+import { PopoverModule } from '../../components/popover';
+import { IconModule } from '../icon';
+import { DateRangePickerModule } from './date-range-picker.module';
+
+
+@Component({
+    selector: 'ux-components-date-range-picker',
+    template: `
+        <ng-template #popoverTemplate>
```
TaraTodd123: You will need to remove <code><ng-template></code> from this. Regarding the karma tests should maybe add a few more tests to test the inputs eg <code>start</code>, <code>end</code> and the outputs eg <code>startChange</code>, <code>endChange</code>

ashley-hunter:
Thanks for the changes, I tried out the plunker with your changes, but when I click the change button the timepicker still does not get updated? Any ideas?

image

src/components/date-range-picker/date-range-picker.spec.ts
ashley-hunter: Agreed, the e2e tests are good for testing based on input to the text area as this is not part of the component, however it would be good to add karma tests to actually test the component inputs and outputs are functioning as expected.

JonathanToddMF: updated tests and added tests for start, end and onRangeChange

JonathanToddMF:
> Thanks for the changes, I tried out the plunker with your changes, but when I click the change button the timepicker still does not get updated? I think there maybe a further issue with the component itself, not just the example code.

Thanks for your help with this one!

TaraTodd123:
The build seems to be failing on one of the tests @JonathanToddMF

src/components/date-range-picker/date-range-picker.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    showSeconds: boolean = false;
+    showMeridian: boolean = true;
+    showSpinners: boolean = true;
+    showNowBtn: boolean = false;
+    startTimezone: any = { name: 'GMT', offset: 0 };
+    endTimezone: any = { name: 'GMT', offset: 0 };
+}
+
+fdescribe('Date Range Picker', () => {
+    let component: DateRangePickerComponent;
```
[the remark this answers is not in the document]
ashley-hunter: It might be good to add a rule to our linting to prevent this:
https://www.npmjs.com/package/tslint-defocus

Its easy to forget to remove the `f` while testing and its a common occurrence which could easily go unnoticed. At least with the linting rule in place the build will fail if we use `describe` or `fit` whereas we are currently only relying on the review to notice this.

src/components/date-range-picker/date-range-picker.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+            [showTimezone]="showTimezone"
+            [showSeconds]="showSeconds"
+            [showMeridian]="showMeridian"
+            [showSpinners]="showSpinners"
+            [showNowBtn]="showNowBtn"
+            [startTimezone]="startTimezone"
+            [endTimezone]="endTimezone"
+            (startChange)="onRangeChange()"
+            (endChange)="onRangeChange()"
+            (keydown.escape)="popover.hide(); input.focus()"
```
TaraTodd123: Had another look at these karma tests, `startChange` and `endChange` outputs are both calling the same function possibly it would be better to call them `startChange` and `endChange` so there is no confusion?

src/components/date-range-picker/date-range-picker.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+        spyOn(component, 'onRangeChange');
+        component.end = new Date('Thu Jan 23 2020 23:59:59 GMT+0000 (Greenwich Mean Time)');
+
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        expect(getDate().innerHTML).toBe(' 23 January 2020 ');
+        expect(component.onRangeChange).toHaveBeenCalled();
+    });
+
```
TaraTodd123: Above tests would need updates to `spyOn` correct functions. Also might be better to call the tests something like should call startChange when the end date is changed and should call startChange when the start date is changed ?

e2e/tests/components/date-range-picker/date-range-picker.po.spec.ts
```diff
@@ -9,6 +9,7 @@ export class DateRangePickerPage {
     disableMinAndMaxBtn = $('#disable-min-max');
     dateFormatBtn = $('#date-format');
     timeFormatBtn = $('#time-format');
+    dateHeader = $('.date-header');
```
TaraTodd123: Use <code>.range-header</code> instead

tslint.json
```diff
@@ -2,8 +2,12 @@
   "rulesDirectory": [
     "codelyzer"
   ],
+  "extends": [
```
ashley-hunter: This will require the npm package to be installed, but I don't see updates to package.json or package-lock.json in these changes?

JonathanToddMF:
Builds are failing doe to the following error
"ERROR InternalError: "too much recursion"

The errors being thrown in the time picker component in the e2e tests. Seemed to be related to the code we added into the date-time-picker.component.ts causing a infinite loop? Removing the line "this.datepicker.date$.next(new Date(value));" fixes the build but this obviously will not fix the issue in the ticket. This will require further investigation.

TaraTodd123:
This is passing now.

docs/app/pages/components/components-sections/date-time-picker/date-range-picker/date-range-picker.component.ts
```diff
@@ -78,7 +78,7 @@ export class ComponentsDateRangePickerComponent extends BaseDocumentationSection
         // check if the date contains a hyphen
```
pamadei: In Digital Safe we are currently getting a call stack exceeded error when using the Date Time Picker component with time view enabled. I notice that there is a incorrect `if` expression:

image

The comment says emit only if they are different, whereas the code is emitting only if they are the same, which is causing an infinite loop.

Might be good if we could fix this as part of this ticket? (It's Ashley btw 😁)

donna-marijne: Fine by me, if you have a repro case for us to add to the ticket for QA?

ashley-hunter: I'll see if I can come up with one, thanks

ashley-hunter: https://plnkr.co/edit/tKHxoVhZvKFJrrMzqdPb?p=preview

ashley-hunter: If `date` is undefined before the popover is opened the error seems to occur

src/components/date-time-picker/date-time-picker.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+        fixture.detectChanges();
+    });
+
+    it('should initialise correctly', () => {
+        expect(component).toBeTruthy();
+    });
+
+    it('should update date and call onDateChange when start date is changed ', async () => {
+        spyOn(component, 'onDateChange');
+        component.date = new Date('Tue Jan 07 2020 00:00:00 GMT+0000 (Greenwich Mean Time)');
```
ashley-hunter: Can I suggest changing this to use the alternate format for setting dates as it is less prone to mistakes:

```typescript
new Date(2020, 0, 7)
```

docs/app/pages/components/components-sections/date-time-picker/date-range-picker/date-range-picker.component.ts
pamadei: Thank you @TaraTodd123 and @ashley-hunter for looking at this 😃

docs/app/pages/components/components-sections/date-time-picker/date-range-picker/date-range-picker.component.ts
```diff
@@ -78,7 +78,7 @@ export class ComponentsDateRangePickerComponent extends BaseDocumentationSection
...
         const parts = (date.indexOf('—') ? date.split('—') : date.split('-'));
         const startDate = Date.parse(parts[0].trim());
-        const endDate = Date.parse(parts[0].trim());
+        const endDate = Date.parse(parts[1].trim());
 
         if (parts.length >= 1 && !isNaN(startDate)) {
```
donna-marijne: Checking parts.length here and below is redundant after already doing `parts[i].trim()` above. Array bounds ought to be checked before accessing the items.

src/components/date-time-picker/time-view/time-view.component.ts
```diff
@@ -58,11 +58,16 @@ export class TimeViewComponent implements OnInit, OnDestroy {
             this.value.setFullYear(date.getFullYear());
             this.value.setMonth(date.getMonth());
             this.value.setDate(date.getDate());
+            this.value.setHours(date.getHours());
+            this.value.setMinutes(date.getMinutes());
```
donna-marijne: I think seconds should be handled here too, in case `showSeconds` is enabled?

e2e/tests/components/date-range-picker/date-range-picker.e2e-spec.ts
```diff
@@ -298,6 +298,23 @@ describe('Date Range Picker Tests', () => {
...
+        expect(await page.getInputText()).toBe('1 January 2020  09:00 AM GMT — 3 January 2020  10:59 PM GMT');
+
+        // remove focus from input
+        await page.clickOnRangeHeader();
+
+        // Check date picker has been updated
+        expect(await page.getPickerDateHeader(Picker.Start)).toBe('1 January 2020');
+        expect(await page.getPickerTimeHeader(Picker.Start)).toBe('9:00 AM');
+        expect(await page.getPickerDateHeader(Picker.End)).toBe('3 January 2020');
+        expect(await page.getPickerTimeHeader(Picker.End)).toBe('10:59 PM');
```
donna-marijne: Can we check that the time spinners (hours, minutes, seconds) also update when the input is changed, since there was a fix for that in this PR? The above is mainly testing the parsing code in the test page component, which doesn't really give us any coverage.

TaraTodd123:
Updates but still need to get screenshots

TaraTodd123:
Updates made and build passing

PR #1274, opened 2020-01-31
EL-3714 - Select recent options (contribution from Tobias)

donna-marijne:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3714

Continuation of https://github.com/UXAspects/UXAspects/pull/1244 to clean up some additional issues:
* Updated documentation and examples.
  * Reset `recentOptions` when switching between strings and objects.
  * Added demo of `recentOptions` binding to the example.
* Passthrough `recentOptions` to allow 2-way binding on `ux-select`.
* Removed an unnecessary function on typeahead key service.
* Extracted the `ol` part of the typeahead to `TypeaheadOptionsListComponent` (internal component) to allow reuse of template logic.
  * Corrected stylesheet to account for the new element in the hierarchy.
  * Corrected some selectors in the e2e tests to work with the updated hierarchy.
  * Refactored other types into their own files to protect against circular dependencies.
  * `active` state does not apply to recent options.
* Accessibility: Fixed the `id` sequencing via `startIndex` on the new component.
* Typeahead refactoring:
  * Added `getVisibleOptions` to eliminate some duplicated code.
  * Added `allVisibleOptions` to replace certain uses of `visibleOptions$`.
  * Renamed/removed a couple of properties.
  * Added `isDisabled` to `TypeaheadVisibleOption` to avoid constantly recalculating this.
  * Removed redundant `highlightedKey` property.
* Filter expression applies to recent options.
* Added classes `ux-typeahead-recent-options` and `ux-typeahead-all-options` to help e2e tests target the correct options list.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3714-select-recent-options

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3714-select-recent-options~CI/)


TaraTodd123:
Thanks, Looks to be working really well! Just noticed that the plunker for the typeahead doesn't work with the new additions, should there be a customizable example for this? or is the code even needed?

src/components/select/select.component.spec.ts
```diff
@@ -604,4 +604,108 @@ describe('Select Component - With custom Icon', () => {
...
+    allowNull: boolean = false;
+    dropdownOpen: boolean = true;
+    recentOptions: string[] = [];
+    recentOptionsMaxCount = 2;
+
+    onRecentOptionsChange(): void {}
+}
+
+fdescribe('Select with recent options', () => {
+
```
TaraTodd123: Just noticed there is an f left in-front of the describe.

donna-marijne: Sorry, still making changes on this.

donna-marijne:
Should be fixed now, I also added a tslint rule to help with accidentally committing fdescribe, xdescribe etc.

Regarding the plunkers, it should work if you pass in an empty array per the documentation. I did not enable it by default for either select or typeahead because I don't think it will be used by the majority of people consuming the component.

src/components/typeahead/typeahead.component.ts
```diff
@@ -221,6 +217,20 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
...
 
+        // Cut off recentOptions at recentOptionsMaxCount
+        if (changes.recentOptions || changes.recentOptionsMaxCount) {
+            this._recentOptions = this.recentOptions
+                ? this.recentOptions.slice(0, this.recentOptionsMaxCount)
+                : undefined;
+
+            if (changes.recentOptionsMaxCount) {
+                // Avoid ExpressionChangedAfterChecked error
+                setTimeout(() => {
```
ashley-hunter: Event emitters can be asynchronous which should prevent the expression changed error without us having to use timeouts etc.

You could try:

```typescript
@Ouput() recentOptionsChange = new EventEmitter<ReadonlyArray<T>>(true);
```

And that should hopefully negate the need for the timeout

donna-marijne: I don't really know why, because isAsync seems to do basically the same thing, but changing this breaks the karma tests that I had added.

ashley-hunter: Ok, no worries. Then leave it as is, thanks!

PR #1280, opened 2020-02-12
EL-3660 - typeahead dynamic dropdown

TaraTodd123:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3660

- added a dropDirection 'auto' setting
- Added tests e2e/karma for the typeahead and the dropDirection settings
- Added screenshot tests for typeahead.
- Created an injectable service for the functionality of checking the component in the viewport and how close it is to the bottom of the window.
- Updated the documentation, examples and plunker for the typeahead component and select component.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3600-typeahead-dynamic-dropdown

#### Downstream Build(s)

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/caf_CI_ux-aspects-micro-focus_EL-3600-typeahead-dynamic-dropdown


TaraTodd123:
Moved functionality into an injectable service. Moved some of the tests to karma tests but I have had to keep the dropDirection auto functionality in a e2e test as due to drop direction changing when on scrolling to the bottom of the page or being near the bottom of the page I could not get this functionality to show in a karma test.

src/services/viewport-listerner/viewport-listener.module.ts
```diff
@@ -0,0 +1,9 @@
+import { NgModule } from '@angular/core';
+import { ViewportListenerService } from './viewport-listener.service';
+import { ResizeModule } from '../../../dist/directives/resize';
```
donna-marijne: I think this is the problem, you've accidentally imported from the dist directory.

TaraTodd123: Thanks!

e2e/tests/components/typeahead/typeahead.e2e-spec.ts
```diff
@@ -0,0 +1,27 @@
...
+
+        await page.typeaheadInput.click();
+        expect(await page.getTypeaheadOptionListClass()).toContain('ux-typeahead-all-options');
+        expect(await page.getTypeaheadClass()).toContain('drop-up');
+
+        await page.clickOnMaxHeightDecrease();
+
+        await page.typeaheadInput.click();
+        expect(await page.getTypeaheadOptionListClass()).toContain('ux-typeahead-all-options');
+        expect(page.getTypeaheadClass()).not.toContain('drop-up');
```
donna-marijne: Should we perhaps include screenshots for the two dropdown positions?

src/services/viewport-listerner/viewport-listener.service.ts
```diff
@@ -0,0 +1,77 @@
+import { ElementRef, Injectable, OnDestroy, Renderer2 } from '@angular/core';
+import { ViewportRuler } from '@angular/cdk/scrolling';
+import { ResizeService } from '../../directives/resize/index';
+import { takeUntil } from 'rxjs/operators';
+import { fromEvent, Subject } from 'rxjs';
+
+@Injectable()
+/** Rename  */
```
donna-marijne: Possibly something like `PopoverOrientationService` which would indicate its usage?

src/services/viewport-listerner/viewport-listener.service.ts
```diff
@@ -0,0 +1,77 @@
...
+
+        return new ViewportListener(nativeElement, nativeElementParent, this._resizeService, this._viewportRuler, this._renderer);
+    }
+
+}
+
+export class ViewportListener implements OnDestroy {
+
+    /** Allow subscribing to state changes */
+    change$ = new Subject<ViewportDirection>();
```
donna-marijne: This might be better named `orientation$`.

src/services/viewport-listerner/viewport-listener.service.ts
```diff
@@ -0,0 +1,77 @@
...
+                this.onScrollOrResize();
+            });
+
+        // watch for changes to the typeahead position when scrolling
+        fromEvent(window, 'scroll', { passive: true }).pipe(takeUntil(this._onDestroy))
+            .subscribe(() => this.onScrollOrResize());
+
+    }
+
+    ngOnDestroy(): void {
```
donna-marijne: Looking at the current code, I don't think this will be invoked. `ngOnDestroy` would be called if defined on the service, but not a plain object like this. You could rename this to `destroy` and call it in the `ngOnDestroy` hook of the typeahead component. However, since the code is not really making use of the `ViewportListener` object, you could just move the implementation into the service and eliminate `ViewportListener`. In that case `ngOnDestroy` would be called automatically when the component that injected it gets destroyed.

src/services/viewport-listerner/viewport-listener.service.ts
```diff
@@ -0,0 +1,77 @@
...
+        this._rect = this._elementParent ? this._elementParent.parentElement.getBoundingClientRect() : this._element.parentElement.getBoundingClientRect();
+        const itemHeight = this._element.offsetHeight;
+        const viewportSize = this._viewportRuler.getViewportSize();
+        const bottomSpaceAvailable = viewportSize.height - this._rect.bottom - itemHeight;
+
+        this.change$.next(bottomSpaceAvailable <= 0 ? ViewportDirection.Up : ViewportDirection.Down);
+    }
+}
+
+export const enum ViewportDirection {
```
donna-marijne: Type name may need to update in line with the other changes.

TaraTodd123:
This is ready for review both branches have built with my changes, i've just updated the branch with changes from the develop branch

src/components/typeahead/typeahead.spec.ts
```diff
@@ -0,0 +1,143 @@
...
+
+        // return the values after a delay to simulate server response time
+        return of(values).pipe(delay(1000)).toPromise();
+    }
+
+    constructor(public typeaheadKeyService: TypeaheadKeyService<string>) {
+
+        /* Adding values to typeahead list */
+        for (let index = 0; index < 200; index++) {
+            this.values = [...this.values, chance.name()];
```
donna-marijne: I know this doesn't affect the tests as they currently stand, but we should never use chance in a testing scenario. If you could replace it with something deterministic here that'd be great.

TaraTodd123:
Updates made for above

src/components/typeahead/typeahead.component.ts
```diff
@@ -255,6 +255,7 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
     ngOnDestroy(): void {
         this._onDestroy.next();
         this._onDestroy.complete();
+        this._popoverOrientation.createPopoverOrientationListener(this.typeaheadElement.nativeElement, this.typeaheadElement.nativeElement.parentElement).destroy();
```
donna-marijne: You'll need to save the original instance as a private property and destroy that, this is just creating a second set of resources and cleaning those up.

TaraTodd123:
Update made.

src/components/typeahead/typeahead.component.ts
```diff
@@ -138,6 +138,8 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
 
     private _onDestroy = new Subject<void>();
 
+    private _popoverOrientationDirection: PopoverOrientationListener;
```
donna-marijne: Presumably this should be named `_popoverOrientationListener` now.

src/components/typeahead/typeahead.component.ts
```diff
@@ -144,7 +149,11 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
     constructor(
         public typeaheadElement: ElementRef,
         private _changeDetector: ChangeDetectorRef,
-        private _service: TypeaheadService
+        private _popoverOrientation: PopoverOrientationService,
```
donna-marijne: With the change to store the object, this no longer needs to be persisted, so you can remove the `private` and the underscore.

src/components/typeahead/typeahead.component.ts
```diff
@@ -174,6 +183,17 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
                 }
             });
 
+        this._popoverOrientationDirection = this._popoverOrientation.createPopoverOrientationListener(this.typeaheadElement.nativeElement, this.typeaheadElement.nativeElement.parentElement);
+
+        this._popoverOrientationDirection.orientation$.pipe(takeUntil(this._onDestroy))
+            .subscribe(direction => {
+                if (this.dropDirection === 'auto' && direction === PopoverOrientationDirection.Up) {
```
donna-marijne: Just noticed that dropDirection = 'up' no longer works, I think this code will need to change. I'd suggest switching away from using renderer and instead computing a property with a hostbinding, e.g.:
```ts
@HostBinding('class.drop-up')
private _dropUp: boolean;
```

You can then update the property from both sources, e.g. set up ngOnChanges to watch `dropDirection`:
```ts
if (changes.dropDirection.currentValue === 'auto') {
  this._dropUp = this._popoverOrientationDirection.orientation$.getValue() === PopoverOrientationDirection.Up;
} else {
  this._dropUp = changes.dropDirection.currentValue === 'up';
}
```

This subscription should then only update `_dropUp` if `dropDirection` is auto.

src/services/popover-orientation/popover-orientation.service.ts
```diff
@@ -0,0 +1,76 @@
...
+        this._rect = this._elementParent ? this._elementParent.parentElement.getBoundingClientRect() : this._element.parentElement.getBoundingClientRect();
+        const itemHeight = this._element.offsetHeight;
+        const viewportSize = this._viewportRuler.getViewportSize();
+        const bottomSpaceAvailable = viewportSize.height - this._rect.bottom - itemHeight;
+
+        this.orientation$.next(bottomSpaceAvailable <= 0 ? PopoverOrientationDirection.Up : PopoverOrientationDirection.Down);
+    }
+}
+
+export const enum PopoverOrientationDirection {
```
donna-marijne: Name got a bit messed up, should probably be `PopoverOrientation`?

src/components/typeahead/typeahead.component.ts
TaraTodd123: I tried doing the above but because I am using orientation$ is a Subject I don't have access to valueOf() and some of the times the current value was coming back as undefined. I've pushed a fix that has fixed the 'up' dropDirection.

TaraTodd123:
I think adding an e2e test with a screenshot will help with making sure the dropDirection 'up' and 'down' works as per our functionality is in the docs. As I already have karma tests in for it as well, and they are passing or failing when I say expect the class list to contain the drop-up class and vice versa.

src/components/typeahead/typeahead.component.ts
donna-marijne: You can convert orientation$ to a BehaviorSubject to resolve that.

src/services/popover-orientation/popover-orientation.service.ts
```diff
@@ -28,7 +28,7 @@ export class PopoverOrientationService {
 export class PopoverOrientationListener {
 
     /** Allow subscribing to state changes */
-    orientation$ = new Subject<PopoverOrientation>();
+    orientation$ = new BehaviorSubject<PopoverOrientation>(null);
```
donna-marijne: Default orientation is down, so that would make a good initial value here.

src/services/popover-orientation/popover-orientation.service.ts
```diff
@@ -1,8 +1,8 @@
-import { ElementRef, Injectable, OnDestroy, Renderer2 } from '@angular/core';
+import { ElementRef, Injectable, Renderer2 } from '@angular/core';
```
donna-marijne: Also I think Renderer2 is not being used for anything in this file.

src/components/typeahead/typeahead.component.ts
```diff
@@ -187,10 +187,10 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
 
         this._popoverOrientationListener.orientation$.pipe(takeUntil(this._onDestroy))
             .subscribe(direction => {
-                if (this.dropDirection === 'auto' && direction === PopoverOrientation.Up || this.dropDirection === 'up') {
-                    this._renderer.addClass(this.typeaheadElement.nativeElement, 'drop-up');
+                if (this.dropDirection === 'auto') {
+                    this._dropUp = direction === PopoverOrientation.Up;
                 } else {
-                    this._renderer.removeClass(this.typeaheadElement.nativeElement, 'drop-up');
+                    this._dropUp = this.dropDirection === 'up';
```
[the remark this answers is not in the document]
TaraTodd123: removed

src/components/typeahead/typeahead.component.ts
```diff
@@ -271,6 +279,9 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
         this.clicking = false;
     }
 
+    @HostBinding('class.drop-up')
+    _dropUp: boolean;
```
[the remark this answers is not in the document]
TaraTodd123: Its not liking the private flag but ive moved it up to above the constructor

src/services/popover-orientation/popover-orientation.service.ts
```diff
@@ -1,8 +1,8 @@
...
 import { ViewportRuler } from '@angular/cdk/scrolling';
 import { ResizeService } from '../../directives/resize/index';
 import { takeUntil } from 'rxjs/operators';
-import { fromEvent, Subject } from 'rxjs';
+import { BehaviorSubject, fromEvent, Subject } from 'rxjs';
```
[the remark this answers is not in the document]
TaraTodd123: Subject needs to remain for the _onDestroy

src/services/popover-orientation/popover-orientation.service.ts
TaraTodd123: updated to down as the initial value

src/services/popover-orientation/popover-orientation.service.ts
TaraTodd123: removed Renderer2

TaraTodd123:
That's all the changes up to date now. Just waiting on the build to pass.

src/components/typeahead/typeahead.component.ts
```diff
@@ -144,7 +152,11 @@ export class TypeaheadComponent<T = any> implements OnChanges, OnDestroy {
     constructor(
         public typeaheadElement: ElementRef,
         private _changeDetector: ChangeDetectorRef,
-        private _service: TypeaheadService
+        public popoverOrientation: PopoverOrientationService,
```
donna-marijne: This should not be public, it is only used in the constructor body.

TaraTodd123:
Build has finally passed!!

PR #1290, opened 2020-02-21
EL-3770 - dashboard options update

andrew-Jolly:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3770

- dashboard now renders when options updated
- karma tests added to check that options update
- e2e test added to prove widgets react to changes
- dashboard screenshots updated

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3770-Dashboard-Options-Update

#### Downstream Build(s)

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/caf_CI_ux-aspects-micro-focus_EL-3770-Dashboard-Options-Update

https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3770-Dashboard-Options-Update~CI/


src/components/dashboard/dashboard.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    });
+
+    it('should change columns option if columns option adjusted', async() => {
+        expect(component.options.columns).toBe(3);
+
+        component.options.columns = 6;
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        expect(component.options.columns).toBe(6);
```
donna-marijne: I believe the original plunker proved that the options object did not revert back to the original value after being updated. The problem was the dashboard itself not updating, which is what these tests should aim to verify by inspecting the rendered DOM. This one for example might be done by checking the dashboard's aria-label.

src/components/dashboard/dashboard.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    });
+
+    it('should initialise correctly', () => {
+        expect(component).toBeTruthy();
+    });
+
+    it('should change columns option if columns option adjusted', async() => {
+        expect(component.options.columns).toBe(3);
+
+        component.options.columns = 6;
```
donna-marijne: The dashboard component is `OnPush` so I'm not even sure this is triggering change detection.

src/components/dashboard/dashboard.spec.ts
andrew-Jolly: I agree, I don't think these test really prove much. This is why I decided to do the e2e tests. The problem I've got with this is that I need to inspect the widgets rather than the dashboard itself. I had a few goes at introducing widgets to the tests but I couldn't find a way of actually getting them to behave correctly.

donna-marijne: What was the problem?

andrew-Jolly: The widgets just stacked on top of each other. I tried just using an individual widget but still no css was taking effect. Maybe I was looking at this from the wrong angle and I should have been testing the widget-component. The widget gets all its style attributes via @HostBinding and from what I can see in plunker, it was getting the option changes, its just the dashboard wasn't rendering on change. This is why I think the e2e test is the way to go to prove this fix. Since there wasn't a dashboard testbed, is it worth keeping it just in case its required in the future?

andrew-Jolly: Actually on looking back at it again, the widgets were not getting the updated options until the window was resized causing the dashboard to re render.

donna-marijne: I think some CSS must be taking effect if they are stacking on top of each other, since that would require `position: absolute`. Bear in mind that a LOT of the styling is custom on that dashboard example, you can check by opening the plunker.

However, we don't really care what it looks like for unit tests. Things like the row height and min width should be programmatically applied to the respective elements, so should be possible to confirm the effects of all of these options regardless of other styling or animation state.
image

donna-marijne: > 
> 
> Actually on looking back at it again, the widgets were not getting the updated options until the window was resized causing the dashboard to re render.

Yes, as noted below you are not triggering change detection because you are not changing the object reference of `options`.

src/components/dashboard/dashboard.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+                            </div>
+                        </div>
+                    </ux-dashboard-widget>
+
+                </ux-dashboard>
+    `
+})
+export class DashboardComponent {
+
+    options: DashboardOptions;
```
donna-marijne: Set `options = this.initialOptions`, to save you doing it in the test body and running change detection again.

src/components/dashboard/dashboard.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+        let adjustedOptions: DashboardOptions  = {
+            columns: 3,
+            padding: 5,
+            rowHeight: 400,
+            emptyRow: false,
+            minWidth: 187
+        };
+        component.options = component.initialOptions;
+        fixture.detectChanges();
+        await fixture.whenStable();
```
donna-marijne: As noted above, these 3 lines can be removed if the test component is set up correctly.

src/components/dashboard/dashboard.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+            rowHeight: 200,
+            emptyRow: false,
+            minWidth: 400
+        };
+
+        component.options = adjustedOptions;
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        expect(widget.style.height).toBe('400px');
```
donna-marijne: The purpose of minWidth is to switch the dashboard to stacked mode, so if you set the test up to have e.g. two widgets on the first row and one on the next, you can test that increasing the minWidth causes the dashboard height to increase from 2*rowHeight to 3*rowHeight.

PR #1296, opened 2020-02-28
EL-3805 - infinite scroll duplicate load request

andrew-Jolly:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3805

- filter Types that are either null or undefined are converted to empty strings.
- karma tests added to confirm the above and check that the filter still works with all Types.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3805-Infinite-Scroll-duplicate-load-request

#### Downstream Build(s)

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/caf_CI_ux-aspects-micro-focus_EL-3805-Infinite-Scroll-duplicate-load-request

https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3805-Infinite-Scroll-duplicate-load-request~CI/


src/directives/infinite-scroll/infinite-scroll.directive.ts
```diff
@@ -314,7 +320,13 @@ export class InfiniteScrollDirective<T = any> implements OnInit, AfterContentIni
...
             const observable = Array.isArray(loadResult) ? of(loadResult) : from(loadResult);
 
-            const subscription = observable.pipe(first()).subscribe(
+            let subscription: Subscription;
+
+            // subscription needs to be a let here if subscription completes right away the complete function can be called
+            // before the assignment. While browsers will ignore this currently as we transpile
+            // all const/let statements to var, when this is no longer the case (or we are running in a test environment)
+            // const will be used and this can throw an error if we try to access a const variable before it is assigned
+            subscription = observable.pipe(first()).subscribe(
```
ashley-hunter: Just looking here we have a case when the subscription complete function is fired right away, however below this we are still adding the subscription to the list of subscriptions: 

```
this._subscriptions.push(subscription);
```

We should probably only add it to the list if it has not already completed

src/directives/infinite-scroll/infinite-scroll.component.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+            declarations: [InfiniteScrollTestComponent],
+        })
+            .compileComponents();
+    }));
+
+    beforeEach(() => {
+        fixture = TestBed.createComponent(InfiniteScrollTestComponent);
+        component = fixture.componentInstance;
+        loadSpy = spyOn(component, 'load').and.callThrough();
+        component.filterText = undefined;
```
ashley-hunter: I don't think this is required as the default value will already be `undefined`

src/directives/infinite-scroll/infinite-scroll.component.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+                </div>
+    `
+})
+export class InfiniteScrollTestComponent {
+
+    filterText: any;
+    pageSize = 20;
+    loadOnScroll: boolean = false;
+    loading: boolean = false;
+    exhausted: boolean = false;
```
ashley-hunter: I think the only property we need here is `filterText`. `loading` and `exhausted` are never used and `pageSize` and `loadOnScroll` never change so can just be put in the html, eg: `[loadOnScroll]="true"`

src/directives/infinite-scroll/infinite-scroll.component.spec.ts
```diff
@@ -0,0 +1,127 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+import { InfiniteScrollModule } from './infinite-scroll.module';
+import { Component } from '@angular/core';
+import { FormsModule } from '@angular/forms';
+
+
+@Component({
+    template: `<div class="row">
```
ashley-hunter: I think we can probably simplify this template to just:

```html
<div [uxInfiniteScroll]="load" [filter]="filterText" [pageSize]="pageSize" [loadOnScroll]="loadOnScroll"></div>
```

I don't think we need the input or anything as we are just updating the `filterText` variable in code

src/directives/infinite-scroll/infinite-scroll.component.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    beforeEach(() => {
+        fixture = TestBed.createComponent(InfiniteScrollTestComponent);
+        component = fixture.componentInstance;
+        loadSpy = spyOn(component, 'load').and.callThrough();
+        component.filterText = undefined;
+        fixture.detectChanges();
+    });
+
+    it ('should initialise correctly', () => {
```
ashley-hunter: We can probably remove this test as none of the tests will pass if the fixture doesn't get initialized

src/directives/infinite-scroll/infinite-scroll.component.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+        expect(component).toBeTruthy();
+    });
+
+    it ('should initially call load with filter value of "" if filter input value is undefined', async() => {
+
+        expect(component.filterText).toBe(undefined);
+
+        fixture.detectChanges();
+        await fixture.whenStable();
```
ashley-hunter: Is this needed? The `expect` isn't needed as this is the default value, we aren't changing it. The detectChanges bit may also not be needed

e2e/pages/app/infinite-scroll/standard/infinite-scroll.testpage.component.ts
```diff
@@ -34,7 +34,7 @@ export class InfiniteScrollTestPageComponent {
                     .filter((e) => this.isFilterMatch(e))
                     .slice(pageStart, pageStart + pageSize);
                 resolve(newItems);
-            }, 200);
+            }, 2000);
```
ashley-hunter: What was the reason for needing this change? We generally want to avoid timeouts etc.. in e2e tests, as if we have 10 tests, waiting 200ms this adds a 2 second delay, but bumping them up to 2 seconds delay could increase the test time to 20 seconds for 10 tests which we want to avoid as our e2e tests are already slow

src/directives/infinite-scroll/infinite-scroll.directive.ts
ashley-hunter: So to fix this perhaps (I haven't tested this). But you could create something a variable like this:

```typescript
let completed: boolean = false;
let subscription: Subscription;

...

() => {
    // remove this request from the list
    this._subscriptions = this._subscriptions.filter(s => s !== subscription);

    completed = true;
}

...

if (!completed) {
    this._subscriptions.push(subscription);
}
```

e2e/pages/app/infinite-scroll/standard/infinite-scroll.testpage.component.ts
andrew-Jolly: I just noticed that the set timeout was 2000 in the fullscreen test and 200 here. I can try reduce this on both tests to 500. I think the problem might have been linked to Line17 `debouncedFilterText = this.filterText.pipe(debounceTime(500));` Because the load request isnt called twice now, having a timeout of less than 500 may have caused it to fail?

ashley-hunter: I think in this case we should reduce the debounce time to a really low value (in both fullscreen and non fullscreen).

We are programmatically setting the input value so we won't need much of a debounce as that is to prevent unnecessary work if the person is slow at typing, which protractor will not be. Any e2e test speed improvements is a great.

src/directives/infinite-scroll/infinite-scroll.directive.ts
andrew-Jolly: Ill try the above and run the tests again.

src/directives/infinite-scroll/infinite-scroll.component.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+    beforeEach(() => {
+        fixture = TestBed.createComponent(InfiniteScrollTestComponent);
+        component = fixture.componentInstance;
+        loadSpy = spyOn(component, 'load').and.callThrough();
+        fixture.detectChanges();
+    });
+
+    it ('should initially call load with filter value of "" if filter input value is undefined', async() => {
+
+        expect(loadSpy).toHaveBeenCalledWith(0, 20, '');
```
ashley-hunter: I wonder would it also be good to add some checks for the number of times the load function is called to ensure we don't try to load pages multiple times, eg:

```typescript
expect(loadSpy).toHaveBeenCalledTimes(1);
```

PR #1303, opened 2020-03-09
EL-3794 - Column Sorting Order

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3794
https://portal.digitalsafe.net/browse/EL-3813

- Updating the column sorting section to make it re-usable across both the doc sites
- Adding `order` and `orderChange` property to allow column sorting state to be saved and restored
- Added `allowNoSort` input to only allow toggling between Ascending and Descending
- Updated example code to use `ReadonlyArray`
- Adding unit tests for bother column sorting directive and component

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3794-Column-Sorting-Order

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3794-Column-Sorting-Order~CI/)


ashley-hunter:
Hi Donna, thank you for the great feedback!

My only query is regarding the issue you mention about the sort function being called when programmatically either on start up or changing the sort order by input. 

Currently this is all handled by the consuming component, when the user clicks or uses the keyboard on the th:

image


My only concern with that is that that would require us having to emit the `stateChange` or `orderChange` event when:
- On load emit a change (which seems wrong as nothing has changed)
- On Input change, emit which is what we have been fixing in a lot of components this sprint as we should only emit when the component changes the value.

Or had you something else in mind?

donna-marijne:
Yes those are good points, what I had in mind was trying to reduce the number of integration points with the component, but I realise that that does not really fit with the API that we currently have.

Possibly it's something to be discussed at another time, or maybe a generic grid component which integrates this and other components can take care of those concerns.

Maybe we can just note something in the documentation to the effect of "it is the application's responsibility to ensure that the initial sort state of the table matches the initial column sorting state"?

ashley-hunter:
That sounds good to me, I will make the suggested changes, thanks!

ashley-hunter:
The sort order issue should now be resolved, and documentation should now be updated.

docs/app/pages/components/components-sections/tables/column-sorting/column-sorting.component.html
```diff
@@ -1,133 +1,137 @@
...
-        <tr uxColumnSorting>
-            <th></th>
-            <th class="clickable"
-                uxFocusIndicator
-                tabindex="0"
-                [attr.aria-label]="getColumnAriaLabel('Name', nameColumn)"
-                [attr.aria-sort]="nameColumn.state"
-                (click)="changeState('Name', nameColumn)"
-                (keydown.enter)="changeState('Name', nameColumn)">
+    <tr uxColumnSorting>
```
donna-marijne: Seems like the indentation got a bit messed up here.

ashley-hunter: Thanks! Should now be fixed

PR #1306, opened 2020-03-10
EL-3812 - Value-change

cmallon367:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3812

Added a workaround to valueChange output to make sure it does not  get emitted twice
Added a karma test

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3812-Value-change

#### Downstream Build(s)

 
https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3812-Value-change~CI


src/components/number-picker/number-picker.component.ts
```diff
@@ -49,10 +49,12 @@ export class NumberPickerComponent implements ControlValueAccessor, OnDestroy, O
     }
 
     set value(value: number) {
+        if (this._value !== value) {
```
donna-marijne: Indentation...

src/components/number-picker/number-picker.component.ts
```diff
@@ -49,10 +49,12 @@ export class NumberPickerComponent implements ControlValueAccessor, OnDestroy, O
     }
 
     set value(value: number) {
-        this._value = value;
-        this.valueChange.emit(value);
-        this._propagateChange(value);
-        this._valid = this.isValid();
+        if (this._value !== value) {
+            this._value = value;
+            this.valueChange.emit(value);
```
ashley-hunter: I don't think this fixes the issue. The code that was here before was wrong, so it's more that adding a check to see if it has changed.

Currently as it stands, if the user changes the value using an input, the valueChange event still gets emitted, see the plunker: https://plnkr.co/edit/fe7IycWw0zDlpRcq

Really we should never be emitting from a setter, as this emits any time the value changes which is wrong, we only want to emit whenever the component changes the value, in other words typing in the input, or clicking the buttons, so we should use those events to emit and not any time the value changes.

We also include tests to cover that changing the input does not emit a value.

src/components/number-picker/number-picker.component.ts
```diff
@@ -127,7 +127,10 @@ export class NumberPickerComponent implements ControlValueAccessor, OnDestroy, O
         }
 
         if (!this.disabled) {
-            this.value = Math.max(Math.min(this.value + this.step, this.max), this.min);
+            this.value = Math.max(
```
ashley-hunter: I think you should maybe increase the line length of your code formatter, its seems a bit over keen to wrap lines. Lines this short should probably remain all on one line to make them easier to read.

src/components/number-picker/number-picker.component.ts
ashley-hunter: If you need any help with this just let me know

donna-marijne:
Seems like some test failures on this one?

cmallon367:
@donna-marijne Yeah i just seen that, fails on e2e tests, I will look into it when I've set up the table poc

donna-marijne:
Hi Christina, if you could give this priority over the PoC that would be great. This one is an internal request.

src/components/number-picker/number-picker.component.ts
```diff
@@ -49,10 +49,11 @@ export class NumberPickerComponent implements ControlValueAccessor, OnDestroy, O
     }
 
     set value(value: number) {
-        this._value = value;
-        this.valueChange.emit(value);
-        this._propagateChange(value);
-        this._valid = this.isValid();
+        if (this._value !== value) {
+            this._value = value;
+            this._propagateChange(value);
```
ashley-hunter: I think this should also be in the increment and decrement functions as this is essentially the equivalent of an event emitter for Angular Forms and could mark a field as `dirty` whenever it hasn't actually been changed

src/components/number-picker/number-picker.component.spec.ts
```diff
@@ -430,4 +436,45 @@ describe('Number Picker Component - value', () => {
...
+    });
+
+    it('should not emit valueChange when the value input changes', async () => {
+        component.value = 1;
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        spyOn(component, 'onValueChange');
+
+        let controlUp = nativeElement.querySelector<HTMLButtonElement>('input').value;
```
ashley-hunter: We can make this `const`

src/components/number-picker/number-picker.component.spec.ts
```diff
@@ -430,4 +436,45 @@ describe('Number Picker Component - value', () => {
...
+
+    it('should emit valueChange when entering value' , () => {
+        fixture.detectChanges();
+
+        const inputChange = getInput();
+        inputChange.focus();
+
+        dispatchKeyboardEvent(inputChange, 'keydown', 0, null, '0');
+        component.value = 0;
+        fixture.detectChanges();
```
ashley-hunter: As we discussed, this test has no assertions so we should add some in. Additionally we should also test that ngModelChange is emitted here as well as valueChange.

ashley-hunter:
Well spotted, yeah that shouldn't be happening!

src/components/number-picker/number-picker.component.spec.ts
```diff
@@ -430,4 +437,55 @@ describe('Number Picker Component - value', () => {
...
+        spyOn(component, 'onValueChange');
+
+        const controlUp = nativeElement.querySelector<HTMLButtonElement>('input').value;
+        expect(controlUp).toEqual('1');
+
+        expect(component.onValueChange).not.toHaveBeenCalled();
+    });
+
+    it('should emit valueChange event when a value is entered', async () => {
+        fixture.detectChanges();
```
ashley-hunter: I don't think this detectChanges/whenStable should be required here?

src/components/number-picker/number-picker.component.ts
```diff
@@ -191,9 +192,15 @@ export class NumberPickerComponent implements ControlValueAccessor, OnDestroy, O
...
+     * This is a workaround for angular bug https://github.com/angular/angular/issues/12540 */
     _emitValueChange(value: number): void {
+        if (value === this._lastValue) {
+          return;
+        }
+        this._lastValue = value;
         this.valueChange.emit(value);
         this._propagateChange(value);
-    }
+      }
```
ashley-hunter: Indentation need fixed

src/components/number-picker/number-picker.component.ts
```diff
@@ -192,9 +193,9 @@ export class NumberPickerComponent implements ControlValueAccessor, OnDestroy, O
         this.disabled = isDisabled;
     }
 
+    /** Set the value and emit the change to the output and Angular forms.
+    ** This is a workaround for angular bug https://github.com/angular/angular/issues/12540 */
```
donna-marijne: Could you remove this and make it a normal comment please? I had done this for you, but you overwrote it apparently.

cmallon367:
Both builds are now passing

PR #1307, opened 2020-03-13
El-3734: Select Component

JonathanToddMF:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3734

1. Initial fixes provided by @TaraTodd123 
2. Removed latest commit from Tara and that seemed to resolve any issues

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3734-select-component

#### Downstream Build(s)

https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3734-select-component~CI/


ashley-hunter:
Hi @JonathanToddMF I have pushed a few changes to point you in the right direction, the issue is slightly more widespread than the original changes covered.

Essentially, we subscribed to both the value and input observables and emitted any time they changed, which is wrong, as these observables were changed whenever the user changed the value via an input, or if the component changed it.

So the change needed to be only emit whenever the component changes the values. We also were calling the `_onChange` form callback when we shouldn't either. This callback should be treated the same as an output, and should only be called whenever the component changes the value.

I have made a few changes which seem to work as expected, but I suggest you check in greater detail, and more importantly add test cases to cover the emits and form callbacks

JonathanToddMF:
> Hi @JonathanToddMF I have pushed a few changes to point you in the right direction, the issue is slightly more widespread than the original changes covered.
> 
> Essentially, we subscribed to both the value and input observables and emitted any time they changed, which is wrong, as these observables were changed whenever the user changed the value via an input, or if the component changed it.
> 
> So the change needed to be only emit whenever the component changes the values. We also were calling the `_onChange` form callback when we shouldn't either. This callback should be treated the same as an output, and should only be called whenever the component changes the value.
> 
> I have made a few changes which seem to work as expected, but I suggest you check in greater detail, and more importantly add test cases to cover the emits and form callbacks

Thanks for this Ashley, I will look into it

ashley-hunter:
Added additional fixes, improved some flaky tests and updated the existing tests. Should now be ready for review.

Still one outstanding issue then it'll be good for review. (Removed the draft status before encountering the additional issue)

ashley-hunter:
Tests are now passing

PR #1308, opened 2020-03-18
EL-3815-Wizard-Async-Validation

andrewstevenson-microfocus:
#### Checklist

#### Ticket / Issue
- https://portal.digitalsafe.net/browse/EL-3815
- Need To Implement And Document A Method Of Applying Async Capable Validation To Wizard Steps.

- Add Validator Input to Wizard-Step Component For Passing in Async Validation Functions.
- Handle Validation in Wizard Component
- Documention of Input.
- Add Karma Tests For The Wizard Component

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3815-Wizard-Async-Validation

#### Downstream Build(s)

[Build Status](https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3815-Wizard-Async-Validation~CI/)

* https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3815-Wizard-Async-Validation~CI/
[Build Status](https://jenkins.swinfra.net/job/SEPG/job/UXAspects/view/Developer/job/UXAspects~UXAspects~EL-3815-Wizard-Async-Validation~CI/)


src/components/wizard/wizard.component.ts
```diff
@@ -178,27 +183,66 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
+            if (!this.completed) {
+                this._subscriptions.push(this.subscription);
+            }
+
+        } else {
+            this.wizardValidationCompleted();
+        }
+    }
+
+    wizardValidationCompleted() {
```
ashley-hunter: This function is really responsible for the next logic and it should probably remain `next`. We also might want to run validation on `finish`, in which case this would involve duplicating all the validation logic. It would be better to have validation in a separate function that the next function calls, and if it returns a promise then we can use `await` to wait until any observable or promise has reolved, eg:

```typescript
async next(): void {

   if (!await this.isStepValid()) {
        return;
   }
...

```

docs/app/pages/components/components-sections/wizard/wizard/wizard.component.html
```diff
@@ -1,6 +1,6 @@
 <ux-wizard [orientation]="orientation" (stepChange)="onStepChange($event)">
 
-    <ux-wizard-step *ngFor="let step of steps" [header]="step.header">
+    <ux-wizard-step *ngFor="let step of steps" [header]="step.header" [validator]="userValidator">
```
donna-marijne: I think we discussed not modifying the documentation example, so let's drop this change.

docs/app/pages/components/components-sections/wizard/wizard/wizard.component.ts
```diff
@@ -58,6 +58,14 @@ export class ComponentsWizardComponent extends BaseDocumentationSection implemen
         this._announcer.announce(`${this.steps[index].header} activated`);
     }
 
+    userValidator = function (): boolean {
+
+        // users logic
+        const something: boolean = null;
+
+        return something;
+    };
```
donna-marijne: We'll need to remove this as well.

src/components/wizard/wizard.component.ts
```diff
@@ -178,27 +183,66 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
      */
     next(): void {
 
+        this.currentStep = this.getCurrentStep();
```
donna-marijne: This doesn't really need to be a property of the class since we can call `getCurrentStep()` at any time. Let's change this to
```ts
const currentStep = this.getCurrentStep();
```

and remove the `currentStep: any;` above.

src/components/wizard/wizard.component.ts
```diff
@@ -178,27 +183,66 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
+
+        if (this.currentStep.validator) {
+
+            let validationResult = this.currentStep.validator();
```
ashley-hunter: Can this be a const

karma.conf.js
```diff
@@ -23,7 +23,7 @@ module.exports = function (config) {
         colors: true,
         logLevel: config.LOG_INFO,
         autoWatch: true,
-        browsers: ['ChromeHeadlessNoSandbox'],
+        browsers: ['ChromeHead'],
```
donna-marijne: Remove this change.

docs/app/pages/components/components-sections/wizard/wizard/wizard.component.ts
```diff
@@ -58,6 +58,14 @@ export class ComponentsWizardComponent extends BaseDocumentationSection implemen
         this._announcer.announce(`${this.steps[index].header} activated`);
     }
 
+    userValidator = function (): boolean {
```
ashley-hunter: This should be:

```typescript
userValidator(): boolean {
    ...
}```

We should also remove this `something` code, either just return `true` or put some logic in here,

File: karma.conf.js (hunk 6 above)
Comment (ashley-hunter) [inline karma.conf.js]: This shouldn't be committed

File: src/components/wizard/wizard.component.ts [hunk 8]
```diff
@@ -115,8 +115,13 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
 
     @ContentChild('footerTemplate', { static: false }) footerTemplate: TemplateRef<WizardFooterContext>;
 
+    private _subscriptions: Subscription[] = [];
+
     id: string = `ux-wizard-${uniqueId++}`;
     invalidIndicator: boolean = false;
+    currentStep: any;
+    completed: boolean = false;
+    subscription: Subscription;
```
donna-marijne: Not sure if you need both `subscription` and `subscriptions` here? I think only one can be active at a time because the user can only click next on one step at a time, so probably drop the array.

src/components/wizard/wizard.component.ts
```diff
@@ -115,8 +115,13 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
 
     @ContentChild('footerTemplate', { static: false }) footerTemplate: TemplateRef<WizardFooterContext>;
 
+    private _subscriptions: Subscription[] = [];
+
     id: string = `ux-wizard-${uniqueId++}`;
     invalidIndicator: boolean = false;
+    currentStep: any;
```
ashley-hunter: We shouldn't use `any`

src/components/wizard/wizard.component.ts
```diff
@@ -178,27 +183,66 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
 
+        this.currentStep = this.getCurrentStep();
+
+        if (this.currentStep.validator) {
+
+            let validationResult = this.currentStep.validator();
+
+            const observable = typeof validationResult === 'boolean' || validationResult === null ? of(validationResult) : from(validationResult); // doesnt accept null
+
+            this.subscription = observable.pipe(first()).subscribe(
```
ashley-hunter: I'm not really sure what all this `subscription` and `subscriptions` stuff is needed for? We have no logic in `ngOnDestroy` to tear down these subscriptions?

As per my suggestion below this could all be simplified to something like this:

```typescript
isStepValid(): Promise<boolean> {

    // get the current activer step
    this.currentStep = this.getCurrentStep();

    // if there is no validator then return the valid state
    if (!this.currentStep.validator) {
        return this.currentStep.valid;
    }

    // get the validator result
    const validatorResult = this.currentStep.validator();

    // if the function returns a promise
    return validatorResult instanceof Promise ? validatorResult : validatorResult.toPromise();
}
```

donna-marijne: I'm not sure, but if the wizard was destroyed while a request was pending, it might cause issues when it completes? So perhaps the subscription should be cancelled in ngOnDestroy?

ashley-hunter:
Looks like we are getting some screenshot failures?

PR #1314, opened 2020-04-01
EL-3717 - Badge directive

darrenmcc-microfocus:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3717

Add in a badge directive that can be added to any element.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3717/#/components/notifications#badge

#### Downstream Build(s)

https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3717~CI/


donna-marijne:
+Ashley, in case you want to give feedback on the API for this one.

docs/app/pages/components/components-sections/badge/badge/badge.component.html
```diff
@@ -0,0 +1,69 @@
...
+    The <code>ux-badge</code> directive can be used to easily attached a badge to any DOM element to be used as a label/tag, or to show (x) number of new items relating to it etc.
+</p>
+
+<p>The appearance and behavior of the directive can be configured using the following attributes:</p>
+
+<uxd-api-properties tableTitle="Inputs">
+    <tr uxd-api-property name="badgeBackgroundColor" type="string">
+        The color you want to use for the background color. This should be a string value that matches one of the colors available in the <code>ColorService</code>.
+    </tr>
+    <tr uxd-api-property name="badgeTextColor" type="string">
```
ashley-hunter, line 15: I'm not sure if this input is required. In all of Roland's designs the text color is either black or white, whichever has the best contrast ratio. In which case we can use our contrast ratio directive to automatically set the colour. I notice this is what Angular Material do, they just have a `matBadgeColor` input which sets the background colour and the text is automatically coloured to the most accessible colour. Thoughts?

docs/app/pages/components/components-sections/badge/badge/badge.component.html
```diff
@@ -0,0 +1,69 @@
...
+<hr>
+
+<p>
+    The <code>ux-badge</code> directive can be used to easily attached a badge to any DOM element to be used as a label/tag, or to show (x) number of new items relating to it etc.
+</p>
+
+<p>The appearance and behavior of the directive can be configured using the following attributes:</p>
+
+<uxd-api-properties tableTitle="Inputs">
+    <tr uxd-api-property name="badgeBackgroundColor" type="string">
```
ashley-hunter, line 12: Can I also suggest adding a `badgeHidden` property. There could be occasions whenever we want to conditionally hide the badge

docs/app/pages/components/components-sections/badge/badge/badge.component.html
```diff
@@ -0,0 +1,69 @@
...
+    <tr uxd-api-property name="badgeText" type="string">
+        The text you wish to display on the badge. This property being set will trigger a check to detect if it is a number or text.
+    </tr>
+    <tr uxd-api-property name="badgeMaxLength" type="number">
+        This allows you to set a character / number limit on the text within the badge. If the content is number and larger than the specified size it will append a <strong>+</strong> otherwise it will the content as text and truncate the characters using ellipsis <strong>&hellip;</strong>.
+    </tr>
+    <tr uxd-api-property name="badgeAriaLabel" type="string">
+        The text you wish to used for accessibility devices, such as screen readers etc. If not set this will default to the <code>badgeText</code> value.
+    </tr>
+    <tr uxd-api-property name="badgeVerticalPosition" type="string" defaultValue="top">
```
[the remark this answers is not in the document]
ashley-hunter: I don't see a 'center' variant in any of Roland's designs? Would it just be top and bottom? Alternatively we may want to follow Materials naming of 'above'/'below' and 'before' and 'after'?

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,170 @@
...
+        this._fontColor = color ? ThemeColor.parse(this._colorService.resolve(color)) : this._determineContrastColor();
+    }
+
+    @Input() public set badgeText(s: string) {
+        const subject = s.trim();
+        this._isNumber =  /^\d+$/.test(subject);
+        this._text = subject;
+    }
+
+    @Input() public set badgeAriaLabel(s: string) {
```
ashley-hunter, line 45: Normally we just omit the `public` property as by default TypeScript properties are public. Also if there is no "transforming the value" its best just to use `@Input() badgeAriaLabel: string;` instead of a setter, as a setter is essentially a function call and is slower than an assigment

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,170 @@
+import { Directive, ElementRef, Input, Renderer, OnInit, OnChanges, SimpleChanges } from '@angular/core';
+import { ColorService } from 'src/services/color/color.service';
+import { ThemeColor } from 'src/services/color/theme-color';
+import { ContrastService } from './../accessibility/contrast-ratio/contrast.service';
+
+export enum BadgeVerticalPosition {
+    Top = 'Top',
```
ashley-hunter, line 7: Probably best to make all the enum string values lower case to match all our other enums

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,170 @@
...
+            ThemeColor.parse(
+                this._contrastService.getContrastColor(
+                    this._backgroundColor,
+                    this._lightColor,
+                    this._darkColor
+                ).toRgba())
+            : this._lightColor;
+    }
+
+    private _determineVPos(position: string): BadgeVerticalPosition {
```
ashley-hunter, line 151: Might be good to consider how Angular Material does it's badge component: https://github.com/angular/components/blob/master/src/material/badge/badge.ts

Essentially it just adds classes to the host element: https://github.com/angular/components/blob/master/src/material/badge/badge.ts#L56

We should be able to skip most of these checks as Angular's template checking in production mode (especially in Angular 9) will throw an exception if the user tries to pass an invalid type to the property

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,170 @@
...
+    selector: '[uxBadge]'
+})
+export class BadgeDirective implements OnInit, OnChanges {
+    /**
+     * Define the badge background color
+     */
+    @Input() public set badgeBackgroundColor(color: string) {
+        this._backgroundColor = ThemeColor.parse(this._colorService.resolve(color));
+    }
+
```
ashley-hunter, line 34: We include comments above all our inputs and outputs to allow the code editor to provide documentation to the consumer, eg: https://github.com/UXAspects/UXAspects/blob/develop-1.8.17/src/components/alert/alert.component.ts#L22

src/directives/badge/badge.directive.ts
ashley-hunter: I notice, Material are using an aria-description instead of an aria label: https://github.com/angular/components/blob/master/src/material/badge/badge.ts#L99
We should verify the correct aria attribute to use here

docs/app/pages/components/components-sections/badge/badge/badge.component.html
donna-marijne: Yeah we were talking about this primarily as a rarely-used override in case someone really wanted red text or something, but that might be me overthinking it. I'll link this to Roland and see if he can think of any valid use case of text color other than black or white.

docs/app/pages/components/components-sections/badge/badge/badge.component.html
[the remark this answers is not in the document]
donna-marijne: I'm guessing it was this:
image

However I think that might be better as a separate setting or maybe a standalone inline component, since it has completely different spacing than the other alignments.

src/directives/badge/badge.directive.ts
donna-marijne: It should indeed be aria-description, thanks Ashley.

docs/app/pages/components/components-sections/badge/badge/badge.component.html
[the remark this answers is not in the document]
ashley-hunter: Ah, I missed that. Yeah, I agree, might be good to be separate in the sense it is a inline element rather than something that is absolutely positioned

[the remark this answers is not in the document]
donna-marijne: I think I'll raise a separate ticket for that, we can discuss how to deliver it during story time.

docs/app/pages/components/components-sections/badge/badge/badge.component.html
donna-marijne: Just heard from Roland, he thinks it's unlikely that someone would need to override the text color in any situation, so we can go ahead and drop this input.

docs/app/pages/components/components-sections/badge/badge/badge.component.html
```diff
@@ -0,0 +1,69 @@
...
+        The color you want to use for the background color. This should be a string value that matches one of the colors available in the <code>ColorService</code>.
+    </tr>
+    <tr uxd-api-property name="badgeTextColor" type="string">
+        The color you want to use for the text/number on the badge. This should be a string value that matches one of the colors available in the <code>ColorService</code>. If you do not set this value the <code>ContrastService</code> will automatically pick which text color is suitable based on the <code>badgeBackgroundColor</code>.
+    </tr>
+    <tr uxd-api-property name="badgeText" type="string">
+        The text you wish to display on the badge. This property being set will trigger a check to detect if it is a number or text.
+    </tr>
+    <tr uxd-api-property name="badgeMaxLength" type="number">
+        This allows you to set a character / number limit on the text within the badge. If the content is number and larger than the specified size it will append a <strong>+</strong> otherwise it will the content as text and truncate the characters using ellipsis <strong>&hellip;</strong>.
```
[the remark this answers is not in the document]
darrenmcc-microfocus: I'll change it to `badgeValueLimit` and the behaviour will work like so: if a number is passed it will show this number if the value exceeds it followed by a '+' (as you mentioned above), if text this number is used as a char length for the string and truncates.

docs/app/pages/components/components-sections/badge/badge/badge.component.html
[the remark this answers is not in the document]
darrenmcc-microfocus: @ashley-hunter I added this myself - I thought there may be uses where you'd want it horizontally centered but I'll remove that. I'll follow the angular-material naming convention also.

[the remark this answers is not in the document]
darrenmcc-microfocus: @donna-marijne yes I was trying to keep things cleaner by having inline as part of the vertical position options as they are related - I'll separate them as you have mentioned the difference in spacing.

docs/app/pages/components/components-sections/badge/badge/badge.component.html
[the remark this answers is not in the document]
donna-marijne: Yes, that sound fine, thank you.

docs/app/pages/components/components-sections/badge/badge/badge.component.html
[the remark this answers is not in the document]
donna-marijne: Here's the follow on ticket, for reference: https://portal.digitalsafe.net/browse/EL-3863

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,286 @@
...
+        while (count--) {
+            const current = parentElement.children[count];
+
+            if (current.classList.contains(className)) {
+                parentElement.removeChild(current);
+            }
+        }
+    }
+
+    private _determineContrastColor(): ThemeColor {
```
ashley-hunter, line 269: Just to make you aware of our function naming conventions as they may not be totally obvious at first. Function names only get an underscore if they should be private but can't be marked private because they are used in an Angular template. This is our way to indicate that a public function should not be used. Otherwise we can omit the underscore

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,286 @@
...
+            }
+        }
+
+        badgeElement.classList.remove('hidden');
+
+        return this._badgeElement;
+    }
+
+    private clearExisting(className: string): void {
+        const parentElement: HTMLElement = this._element.nativeElement;
```
ashley-hunter, line 257: I'm assuming this is removing any child element that is a badge? Could this be simplified to the following?

```typescript
if (this._badgeElement) {
    this._element.nativeElement.removeChild(this._badgeElement);
}
```

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,286 @@
...
+
+export type BadgeVerticalPosition = 'above' | 'below' | 'top' | 'bottom';
+
+export type BadgeHorizontalPosition = 'before' | 'after' | 'left' | 'right';
+
+export type BadgeSize = 'small' | 'medium' | 'large';
+
+@Directive({
+    selector: '[uxBadge]',
+    host: {
```
ashley-hunter, line 14: We should add an `exportAs: 'ux-badge'` here to allow easy programmatic access.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,276 @@
+import { AfterContentChecked, Directive, ElementRef, Input, OnChanges, OnDestroy, Renderer2, SimpleChanges } from '@angular/core';
+import { ColorService } from './../../services/color/color.service';
+import { ThemeColor } from './../../services/color/theme-color';
+import { ContrastService } from './../accessibility/contrast-ratio/contrast.service';
+
+export type BadgeVerticalPosition = 'above' | 'below' | 'top' | 'bottom';
+
+export type BadgeHorizontalPosition = 'before' | 'after' | 'left' | 'right';
```
ashley-hunter, line 8: @donna-marijne Do you think we need both 'above/below' and 'top/bottom' or can we just have one set as they both mean the same thing?

As for `before` and `after` and `left` and `right` we might want only one of these, however there is an additional future difference here that we should consider if we do decide to have both. Left and right are explicit directional values, however before and after are not, they are dependent on the language directionality, eg. in ltr language before is left, but in rtl language before is right. Just a thought.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,276 @@
...
+
+@Directive({
+    selector: '[uxBadge]',
+    exportAs: 'ux-badge',
+    host: {
+        class: 'ux-badge-container',
+        '[class.ux-badge-small]': '_badgeSize === "small"',
+        '[class.ux-badge-medium]': '_badgeSize === "medium"',
+        '[class.ux-badge-large]': '_badgeSize === "large"',
+        '[class.ux-badge-above]': 'isAbove()',
```
ashley-hunter, line 20: We can only reference public properties and functions in an Angular production build, otherwise it will throw errors.

Ideally we want to also just use equality checks here rather than function calls. Function calls are really slow in templates (and host bindings) and they get called every time change detection runs. Its often not a noticeable different in our docs site, however some of the apps our components are used in are really heavy enterprise pages that are not well optimized, so making our code efficient can make a big difference overall. Something like `'[class.ux-badge-above]': 'badgeVerticalPosition === "above" || badgeVerticalPosition === "top"` is much more efficient.

src/directives/badge/badge.directive.ts
donna-marijne: No we don't need duplicates, and Ashley's point about RTL languages is well made. Although I'm not aware of anyone using UX Aspects in an RTL locale, it's good forward planning to establish APIs which won't cause problems if/when we do have to address it.

It's good to be accommodating of things like synonyms when processing input that's coming from a user, but this is a developer API with documentation and TypeScript type definition. Extra checks for developer convenience unfortunately turn into performance problems when you have a page with 10,000 input bindings.

src/styles/notifications.less
```diff
@@ -263,4 +263,112 @@ ux-alert {
...
\ No newline at end of file
+}
+
+/*
+ * Badge
+ */
+
+.ux-badge-container {
+    position: relative;
+    overflow: visible;
```
donna-marijne, line 274: There might be components which use overflow: hidden to e.g. truncate content text when a max-width is specified, in which case this would get overridden. I don't think there's any way of automatically resolving that conflict, so probably we would have to add a note to the documentation recommending to use a wrapper element in this situation.

src/styles/notifications.less
```diff
@@ -263,4 +263,112 @@ ux-alert {
...
+
+    .ux-badge {
```
donna-marijne, line 276: Looking at the upcoming badge-inline component, you might consider un-nesting this style rule along with some others below so that they can be used separately.

donna-marijne:
Just one other thought regarding the aggressive line breaking on your code formatter, we have:
```
max_line_length = 120
```
in .editorconfig in the 2.0 version of this repository. Might help make the code a bit more readable.

https://github.com/UXAspects/UXAspects-2/blob/master/.editorconfig

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,276 @@
+import { AfterContentChecked, Directive, ElementRef, Input, OnChanges, OnDestroy, Renderer2, SimpleChanges } from '@angular/core';
+import { ColorService } from './../../services/color/color.service';
```
ashley-hunter, line 2: Just noticed these import paths are starting with a `./` this can be removed as is isn't required.

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -1,81 +1,79 @@
...
+    <button class="btn btn-primary" [uxBadge]="9" badgeSize="LARGE" [badgeHidden]="hideButton">
         Messages
     </button>
     <button class="btn" (click)="hideButton = !hideButton">
         Hide/reveal badge
     </button>
 </p>
 <p>
-    <button class="btn btn-primary" [uxBadge]="99" [badgeSize]="'LarGE'">
+    <button class="btn btn-primary" [uxBadge]="99" badgeSize="LarGE">
```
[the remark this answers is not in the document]
darrenmcc-microfocus: Hey was just testing this worked, these examples are not the final ones, just me testing that everything is working. I'll update this will final examples shortly.

src/directives/badge/badge.directive.ts
```diff
@@ -11,170 +11,158 @@ export type BadgeSize = 'small' | 'medium' | 'large';
...
      * Set the badge vertical position in relation to the parent element
      */
-    private _defaultVPos: BadgeVerticalPosition = 'above';
+    private _defaultVerticalPosition: BadgeVerticalPosition = 'above';
 
-    @Input('badgeVerticalPosition')
-    get vPos(): string {
-        return this._vPos;
+    @Input()
+    get badgeVerticalPosition(): string {
```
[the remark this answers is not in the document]
darrenmcc-microfocus: Thanks for explanation @ashley-hunter. I've changed the ones that can be removed now.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,276 @@
...
+    }
+
+    private updateBadge(): HTMLElement {
+        const badgeElement: HTMLSpanElement = this._renderer.createElement(
+            'span'
+        );
+        this.clearExisting();
+
+        badgeElement.classList.add(this._className);
+        badgeElement.classList.add('hidden');
```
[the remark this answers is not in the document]
darrenmcc-microfocus: 'hidden' is a bootstrap utility class for hidding items, I was hiding it while it is being manipulated so it doesn't 'shift' around the screen on a slow load.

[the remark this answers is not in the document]
donna-marijne: We're avoiding using Bootstrap classes inside components at the moment, with the aim of dropping it as a dependency. For this use case, assuming that it is a problem that you have observed, I would suggest applying `display: none` or `visibility: hidden` directly to the element, which avoids the risk that it will be redefined by the consuming application.

src/directives/badge/badge.directive.ts
darrenmcc-microfocus: Thanks again for feedback, so very good points made especially about direction. I've removed the duplicates.

src/directives/badge/badge.directive.ts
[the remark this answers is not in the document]
darrenmcc-microfocus: Ok sure, I should be keeping this library independent of others as you pointed out. I've changed it to manipulate the display style instead.

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,85 @@
+<div class="badge-separator">
+    <button class="btn btn-primary" [uxBadge]="'NEW'" badgeSize="small" badgeOverlap="true">
```
ashley-hunter, line 2: `[uxBadge]="'NEW'"` can be changed to `uxBadge="NEW"` to avoid any unnecessary bindings. There are a few others that can be converted to static strings too.

The `badgeOverlap` property should be a binding as `badgeOverlap="true"` is the string `true` whereas `[badgeOverlap]="true"` is the a boolean `true` which is correct in this instance.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,204 @@
...
+    @Input()
+    private badgeVerticalPosition: BadgeVerticalPosition = this._defaultVerticalPosition;
+
+    /**
+     * Set the badge horizontal position in relation to the parent element
+     */
+    private _defaultHorizontalPosition: BadgeHorizontalPosition = 'after';
+
+    @Input()
+    private badgeHorizontalPosition: BadgeHorizontalPosition = this._defaultHorizontalPosition;
```
ashley-hunter, line 92: Inputs must be public so we can drop the private here

e2e/pages/app/badge/badge.testpage.data.ts
```diff
@@ -0,0 +1,37 @@
+import { BadgeHorizontalPosition, BadgeSize, BadgeVerticalPosition } from '../../../../src/directives/badge';
```
ashley-hunter, line 1: If I recall I think the e2e tests are setup so this should be importing from `@ux-aspects/ux-aspects`. If that doesn't work then we should use this path but include `/index` on the end and Angular can sometimes be a bit awkward when it comes to import paths excluding the `index` part of the path

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,204 @@
...
+     */
+    @Input() badgeMaxValue: number;
+
+    /**
+     * Badge size (based on CSS styles)
+     */
+    private _defaultBadgeSize: BadgeSize = 'medium';
+
+    @Input()
+    private badgeSize: BadgeSize = this._defaultBadgeSize;
```
ashley-hunter, line 113: `_defaultBadgeSize` is only ever used here so we can just change this to:


```
@Input() badgeSize: BadgeSize = 'medium';
```

I believe `_defaultVerticalPosition` and `_defaultHorizontalPosition` and perhaps a few others may also be the same

docs/app/data/components-page.json
```diff
@@ -1016,6 +1016,11 @@
                         }
                     ]
                 },
+                {
+                    "title": "Badge",
+                    "component": "ComponentsBadgeComponent",
+                    "version": "Angular"
```
ashley-hunter, line 1022: We should also add a usage section here which describes the module a consumer needs to import and the selector they need to use eg, `[uxBadge]`

docs/app/pages/components/components-sections/notifications/badge/badge.component.ts
```diff
@@ -0,0 +1,37 @@
...
+        files: {
+            'app.component.html': this.snippets.raw.appHtml,
+            'app.component.ts': this.snippets.raw.appTs,
+        },
+        modules: [
+            {
+                imports: ['BadgeModule'],
+                library: '@ux-aspects/ux-aspects',
+            },
+            {
```
ashley-hunter, line 26: I don't think router should be required here?

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,204 @@
...
+
+        badgeElement.innerHTML = this._badgeDisplayContent;
+        badgeElement.style.setProperty('color', this.determineContentTextColor().toHex());
+
+        if (this.badgeAriaDescription) {
+            badgeElement.setAttribute('aria-describedby', this.badgeAriaDescription);
+        }
+
+        if (this._truncatedText) {
+            badgeElement.setAttribute('title', this._badgeContent);
```
ashley-hunter, line 177: Hmm, so this is adding a browser tooltip if there is truncation occurring. We have avoided the browser tooltips in all our other components/directives as there are MF styled tooltips, however I believe they are currently only a directive so we can't programmatically instantiate one here.

@donna-marijne What are your thoughts here?

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,204 @@
...
+    }
+    set badgeColor(color: string) {
+        this._badgeColor = ThemeColor.parse(this._colorService.resolve(color));
+    }
+    private _badgeColor: ThemeColor = this._defaultBackgroundColor;
+
+    /**
+     * Aria description value for accessibility devices
+     */
+    @Input() badgeAriaDescription: string;
```
ashley-hunter, line 76: If we look a the aria-described by docs: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute we can see that the value added to the element should refer to the id of the element used to describe the object.

If we look at what material does, it also adds an id to the badge element aria-describedby property:
image

And then has the description in another element:
image

I'm guessing it is using https://material.angular.io/cdk/a11y/api#AriaDescriber however this is only supported by CDK v8+ and we are required to to support v7 for now, so perhaps we could create a simple service to do this ourselves (private and not documented) that has the same API surface as the CDK one (although only implement the functions we need) and then we can swap out for the CDK one when we drop v7. @donna-marijne thoughts?

e2e/pages/app/app.module.ts
```diff
@@ -5,219 +5,232 @@ import { RouterModule, Routes } from '@angular/router';
 import { AppComponent } from './app.component';
 
 const routes: Routes = [
+    {
+        path: 'badge',
```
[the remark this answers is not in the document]
donna-marijne: I noticed that a custom prettier config had been added. We need the whole team to be using the same formatting settings, otherwise code is going to get rearranged with every commit, so preferably remove the custom config and update .editorconfig as suggested.

src/directives/badge/badge.directive.ts
donna-marijne: I did a small amount of digging on this since I know that a lot of sites do use `title` as a cheap tooltip.

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/title#Accessibility_concerns

These points lead me to believe that we should not add something like this unless it becomes a specific requirement. We should bear in mind that the truncation is fully under the implementor's control in this case, so it could be argued that it's up to the application itself to make provisions if the text length routinely exceeds the configured maximum (e.g. by adding a tooltip to the host element).

src/styles/notifications.less
donna-marijne: Saw that this was resolved, did you forget to make the change?

src/directives/badge/badge.directive.ts
donna-marijne: Hmm, I misunderstood this originally. Why would adding a hidden element for aria-describedby be better than aria-description? It'll be a string with no markup either way right?

ashley-hunter: Yeah this confused me a bit when I noticed what they were doing in material. I'm guessing that they do it this way because a badge could be applied to an icon which can be aria-hidden? I'm not sure whether this is something we need to consider, but that's my guess as to why they do it this way? It's the only thing I can think of based on their docs.

If we don't think this is necessary then we might be able to get rid of this property entirely? As an aria label could just be added to the element that the button is attached to?

donna-marijne: That could be it, yes. But I think if the icon has meaning to a reader in combination with its badge then it probably have a tooltip attached with a proper description like "134 unread messages." And I agree that if it's a button or something, it can be given an aria-label or description which incorporates the information presented by the badge, e.g. "Show your messages. You have 134 unread messages."

src/directives/badge/badge.directive.ts
darrenmcc-microfocus: This is now removed

src/directives/badge/badge.directive.ts
darrenmcc-microfocus: This is now removed

docs/app/pages/components/components-sections/notifications/badge/badge.component.ts
darrenmcc-microfocus: Hey, yeah it seems it is. The whole page breaks without it.

donna-marijne: What's the error? Why is it required?

e2e/pages/app/app.module.ts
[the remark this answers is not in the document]
darrenmcc-microfocus: Removed prettier which is causing the issues. This should hopefully not be a problem anymore.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,174 @@
...
+import { ThemeColor } from '../../services/color/theme-color';
+import { ContrastService } from '../accessibility/contrast-ratio/contrast.service';
+
+export type BadgeVerticalPosition = 'above' | 'below';
+
+export type BadgeHorizontalPosition = 'before' | 'after';
+
+export type BadgeSize = 'small' | 'medium' | 'large';
+
+@Directive({
```
ashley-hunter, line 12: I notice that hovering over the badge shows the same cursor as the button, it also propagates clicks. I notice material doesn't: https://material.angular.io/components/badge/overview

@donna-marijne  should we change this to match materials behavior?

docs/app/pages/components/components-sections/notifications/badge/snippets/app.html
```diff
@@ -0,0 +1,5 @@
+<button class="btn btn-info" [uxBadge]="content" [badgeMaxValue]="maxValue" [badgeOverlap]="overlap"
```
ashley-hunter, line 1: I notice a few places are using `btn-info` & `btn-primary` to set the color the buttons. We should use `button-primary` instead. This is a divergence from the bootstrap class name (I don't quite know the reason, this was implement before I start working on the project).

Also I don't believe we actually support `btn-info` any more, I believe this is a legacy style. We should probably stick to one of the button styles here which will help ensure it displays nicely in both documentation sites: https://pages.github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/#/ui-components/buttons

src/directives/badge/badge.directive.ts
donna-marijne: Yes, maybe `pointer-events: none` or something like that? Since the badge might overlap other clickable elements on the page, it should not intercept or block click events.

ashley-hunter: That sounds reasonable! Would we ever want something like a badge click event? Not required for this ticket, but just to consider as `pointer-events: none` would have to be reverted if this was ever the case, can you foresee any problems if this was the case? Probably not, just a thought.

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,157 @@
...
+                    <div class="col-md-6 col-sm-12">
+                        <label for="uxBadge">uxBadge</label>
+                        <div class="input-group m-b-sm">
+                            <input type="text" placeholder="Enter description" class="form-control"
+                                [(ngModel)]="badgeContent">
+                        </div>
+                    </div>
+                    <div class="col-md-6 col-sm-12">
+                        <label for="badgeMaxValue">badgeMaxValue</label>
+                        <ux-number-picker [(value)]="badgeMaxValue" min="0" max="1000000"></ux-number-picker>
```
donna-marijne, line 31: Could you use `ux-number-picker-inline` in the customize section? It will look much better in the Micro Focus documentation site.

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,157 @@
...
+    </button>
+    <ux-icon class="icon-example" name="filter" [uxBadge]="badgeContent" [badgeSize]="badgeSize"
+        [badgeMaxValue]="badgeMaxValue" [badgeOverlap]="badgeOverlap" [badgeColor]="selectedColor.name"
+        [badgeHidden]="badgeHidden" [badgeHorizontalPosition]="badgeHorizontalPosition"
+        [badgeVerticalPosition]="badgeVerticalPosition">
+        Messages
+    </ux-icon>
+</div>
+
+<hr />
```
donna-marijne, line 15: The `<hr>` element should be underneath the customize section.

e2e/pages/app/app.module.ts
```diff
@@ -5,219 +5,232 @@ import { RouterModule, Routes } from '@angular/router';
...
+        loadChildren: () => import('./badge/badge.module').then((m) => m.BadgeTestPageModule),
```
donna-marijne, line 10: As noted above, I'd appreciate it if you would revert the unnecessary changes in this file.

(We do actually prefer omitting the brackets from single-argument arrow functions in our code.)

e2e/pages/app/badge/badge.testpage.component.html
```diff
@@ -0,0 +1,52 @@
+<div class="badge-container">
```
donna-marijne, line 1: Although it wasn't noted in the ticket, I think the plan was to add a large number of static examples representing all the mockup examples that Roland provided, and just take a screenshot of that.

Looking at the mockups, you could just set up arrays of each testable property and loop through them in the template, e.g.
```ts
sizes = ['small', 'medium', 'large'];
```
```html
<button *ngFor="let size of sizes" ... [uxBadge]="99" [size]="size">...</button>
```

You could save horizontal space by using icon buttons instead of text buttons.

It would not be necessary to test all combinations, but I would like to see coverage of most of what Roland has mocked up.

e2e/tests/components/badge/badge.e2e-spec.ts
```diff
@@ -0,0 +1,169 @@
...
+        expect(await page.buttonBadge.getCssValue('color')).toBe('rgba(255, 255, 255, 1)');
+        expect(await page.buttonBadge.getCssValue('background-color')).toBe('rgba(2, 86, 98, 1)');
+
+        // check not hidden
+        expect(await page.buttonBadge.getCssValue('display')).toBe('block');
+
+        expect(await imageCompare('badge-initial')).toEqual(0);
+    });
+
+    it('should be able to create an empty badge', async () => {
```
donna-marijne, line 63: I think most of these tests should have been Karma tests, and as noted above we should leverage the e2e tests mainly to check the visuals of the directive. Protractor tests are notably slower than the Karma tests. Karma tests are also easier to manipulate, so you don't need extra buttons to change states.

e2e/tests/components/card-tabs/card-tabs.po.spec.ts
```diff
@@ -1,7 +1,6 @@
 import { $, $$, browser, ElementFinder } from 'protractor';
 
 export class CardTabsPage {
-
```
donna-marijne, line 4: Spurious changes here.

src/styles/notifications.less
```diff
@@ -263,4 +263,154 @@ ux-alert {
...
+        display: block;
+        position: absolute;
+        text-align: center;
+        background-color: #000;
+        color: #fff;
+        box-sizing: border-box;
+        padding: 0;
+        margin: 0;
+        text-transform: none !important;
+        font-family: @font-family !important;
```
donna-marijne, line 285: Please don't use `!important` - especially here, since we use a different font in the Micro Focus theme.

src/styles/notifications.less
```diff
@@ -263,4 +263,154 @@ ux-alert {
...
+    }
+
+    &.ux-badge-above,
+    &.ux-badge-top {
+        .ux-badge {
+            top: 0;
+        }
+
+        &.ux-badge-small .ux-badge {
+            top: -7px;
```
donna-marijne, line 323: I think you should be able to use `translateY` here as well, which would save having to set out a bunch of different hard-coded pixel values for each size variant.

src/directives/badge/badge.directive.ts
donna-marijne: Possibly, but I guess in that case we would still use something like `pointer-events: none` for non-clickable badges. There might also be a better way than using `pointer-events`, since I think that would also prevent a user selecting the badge text in order to copy it.

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,157 @@
+<div class="badge-separator">
+    <button class="btn btn-primary" [uxBadge]="badgeContent" [badgeSize]="badgeSize" [badgeMaxValue]="badgeMaxValue"
```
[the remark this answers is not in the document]
darrenmcc-microfocus: I have a feeling this is an old issue as most of the examples are removed now. It was probably the anchors where using a # as a dummy href but that redirects to main root in angular.

docs/app/data/components-page.json
```diff
@@ -9,8 +9,7 @@
                     "title": "Toggle Buttons",
                     "component": "ComponentsToggleButtonsComponent",
                     "version": "Angular",
-                    "externalUrl":
-                        "http://valor-software.com/ngx-bootstrap/#/buttons",
+                    "externalUrl": "http://valor-software.com/ngx-bootstrap/#/buttons",
```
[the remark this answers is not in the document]
darrenmcc-microfocus: Hey, I understand, this could make merges much more complicated. I think i'll need to compare my settings with someone else to stop this happening as it auto happens on save for all files. Maybe there is a formatter I should be using for JSON files or a difference in config?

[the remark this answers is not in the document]
donna-marijne: For myself, typically I avoid doing a whole file reformat unless it's a real mess. I don't think your setting are necessarily wrong, it's just that this file has been changed a lot from 1.8 to 2.0 so I'd like to leave the big reformat until the changes are merged back into this repo.

.gitignore
```diff
@@ -23,4 +23,4 @@ target/
 .vscode
 .idea
 *.iml
-.tmp
+.tmp
```
darrenmcc-microfocus, line 26: I can't seem to remove this.

donna-marijne: `.tmp` will be removed when we bring in the 2.0 changes.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,197 @@
...
+                    finalText = `${maxValue}+`;
+                } else if (finalText.length > maxValue) {
+                    finalText = `${finalText.substr(0, maxValue)}&hellip;`;
+                }
+            }
+
+            this._badgeDisplayContent = finalText;
+        }
+
+        this.updateBadge();
```
ashley-hunter, line 140: I think we should improve this slightly. Every time ngOnChanges is called this updateBadge function is called Which re-creates the DOM element every time which could be a bit expensive. We should probably only create the badge element once and then only perform update that are required. So perhaps have individual functions that get called only when required to avoid any unnecessary dom manipulation.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,197 @@
...
+        if (this._renderer.destroyNode) {
+            this._renderer.destroyNode(this._badgeElement);
+        }
+    }
+
+    private updateBadge(): HTMLElement {
+        const badgeElement: HTMLSpanElement = this._renderer.createElement('span');
+        this.clearExisting();
+
+        badgeElement.classList.add(this._className);
```
ashley-hunter, line 153: For all this DOM manipulation we should use the renderer functions, eg. renderer.addClass, renderer.setStyle etc..

e2e/pages/app/badge/badge.testpage.component.ts
```diff
@@ -0,0 +1,13 @@
...
+
+@Component({
+    selector: 'app-badge',
+    templateUrl: './badge.testpage.component.html',
+    styleUrls: ['./badge.testpage.component.less'],
+})
+export class BadgeTestPageComponent {
+    sizes: BadgeSize[] = [ 'small', 'medium', 'large' ];
+    verticalPositions: BadgeVerticalPosition[]  = [ 'above', 'below' ];
+    horizontalPositions: BadgeHorizontalPosition[] = ['after', 'before'];
```
donna-marijne, line 12: Some color and border color combinations might be useful here.

e2e/tests/components/badge/badge.po.spec.ts
```diff
@@ -0,0 +1,60 @@
+import { $, browser, by, element } from 'protractor';
+
+export class BadgePage {
+    iconNoContent = $('#iconExampleNoContent');
```
donna-marijne, line 4: I think the majority of this file can be removed now.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,207 @@
...
+export type BadgeHorizontalPosition = 'before' | 'after';
+
+export type BadgeSize = 'small' | 'medium' | 'large';
+
+@Directive({
+    selector: '[uxBadge]',
+    exportAs: 'ux-badge',
+    host: {
+        class: 'ux-badge-container',
+        '[class.ux-badge-small]': 'badgeSize === "small"',
```
donna-marijne, line 17: Since we have the inline badge coming up, I think we'll have to make an adjustment here, The modifier classes should be on the `span.ux-badge` element rather than the parent element. This will allow us to use the badge styles on a standalone inline element.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,207 @@
...
+            ? ThemeColor.parse(
+                this._contrastService.getContrastColor(this._badgeColor, this._lightColor, this._darkColor).toRgba()
+            )
+            : this._lightColor;
+    }
+
+    private parseThemeColor(color: string): ThemeColor {
+        let themeColor: ThemeColor = null;
+
+        if (color) {
```
donna-marijne, line 201: We would usually write this as
```ts
return color ? ThemeColor.parse(this._colorService.resolve(color)) : null;
```

e2e/pages/app/app.module.ts
```diff
@@ -6,206 +6,227 @@ import { AppComponent } from './app.component';
...
 
 const routes: Routes = [
+    {
+        path: 'badge',
+        loadChildren: () => import('./badge/badge.module').then(m => m.BadgeTestPageModule),
+    },
     {
         path: 'card-tabs',
-        loadChildren: () => import('./card-tabs/card-tabs.module').then(m => m.CardTabsTestPageModule)
+        loadChildren: () => import('./card-tabs/card-tabs.module').then(m => m.CardTabsTestPageModule),
```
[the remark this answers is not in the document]
darrenmcc-microfocus: I tried removing these but vs code keeps adding them on save, I don't even think they are language related. I'll open it in textedit to fix.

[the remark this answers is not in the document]
donna-marijne: You'll need to disable format on save:
image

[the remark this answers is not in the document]
darrenmcc-microfocus: Sure. I'll make sure they don't get in at all so it's not an issues, these are spilling over from my incorrect config at the start.

src/directives/badge/badge.directive.ts
darrenmcc-microfocus: @donna-marijne I had to add a check that the colour is valid or it errors out - but this is directly related to typing in a new colour in the docs.
```suggestion
       return color && this._colorService.colorExists(color) ? ThemeColor.parse(this._colorService.resolve(color)) : null;
```

e2e/pages/app/badge/badge.testpage.component.html
```diff
@@ -0,0 +1,92 @@
...
+    <!-- icon example no content -->
+    <ul class="element-list">
+        <ng-container *ngFor="let vertPos of verticalPositions">
+            <ng-container *ngFor="let horzPos of horizontalPositions">
+                <li>
+                    <ux-icon name="filter" uxBadge badgeColor="warning" badgeVerticalPosition="{{ vertPos }}"
```
ashley-hunter, line 7: Could we update any attributes using interpolation, eg:

```html
badgeVerticalPosition="{{ vertPos }}"
```

Should be:

```html
[badgeVerticalPosition]="vertPos"
```

While technically this will work for any string values it will not for any other type and we always recommend using property binding instead.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,214 @@
...
+        private readonly _colorService: ColorService,
+        private readonly _contrastService: ContrastService
+    ) { }
+
+    ngAfterViewInit(): void {
+        const badgeElement: HTMLSpanElement = this._renderer.createElement('span');
+        this._renderer.addClass(badgeElement, this._className);
+        this._renderer.addClass(badgeElement, `ux-badge-${this.badgeSize}`);
+        this._renderer.setStyle(badgeElement, 'display', 'none');
+        this._renderer.setProperty(badgeElement, 'innerHTML', this._badgeDisplayContent);
```
ashley-hunter, line 125: Could we consider updating `textContent` instead of `innerHTML` as innerHTML can be exploited by XSS attacks. If we want to support HTML content at some time in the future we would add support for a TemplateRef to be passed as the content instead of an HTML string

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    constructor(
+        private readonly _element: ElementRef<HTMLElement>,
+        private readonly _renderer: Renderer2,
+        private readonly _colorService: ColorService,
+        private readonly _contrastService: ContrastService
+    ) { }
+
+    ngAfterViewInit(): void {
+        const badgeElement: HTMLSpanElement = this._renderer.createElement('span');
```
ashley-hunter, line 121: The lifecycle hook functions are getting a bit long and have a bit of duplicated code. Perhaps we could split them up a bit, to make them easier to maintain eg:

```typescript
private createBadge(): void {}
private setBadgeColor(): void {}
private setBadgeBorderColor(): void {}
...
```

That way we can move all the logic for each into its one function, each with a single responsibility.

src/directives/badge/badge.spec.ts
```diff
@@ -0,0 +1,222 @@
...
+    borderColor: string = null;
+    maxValue: number = null;
+    hidden: boolean = false;
+    size: string = 'medium';
+    verticalPosition: string = 'above';
+    horizontalPosition: string = 'after';
+    overlap: boolean = false;
+}
+
+fdescribe('Badge', () => {
```
ashley-hunter, line 39: Just need to removed the `f` as the build will fail otherwise

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,225 @@
...
+        return this._badgeColor
+            ? ThemeColor.parse(
+                this._contrastService.getContrastColor(this._badgeColor, this._lightColor, this._darkColor).toRgba()
+            )
+            : this._lightColor;
+    }
+
+    private parseThemeColor(color: string): ThemeColor {
+        // check for hash color
+        if (this._colorService.isHex(color)) {
```
ashley-hunter, line 214: Can I just get some clarification about why this is needed? 

And is this a limitation as we also support rgb and rgba colors when parsing ThemeColors?

Could we do something like this just instead?

```typescript
return this._colorService.colorExists(color) ? 
    ThemeColor.parse(this._colorService.resolve(color)) : 
    ThemeColor.parse(color);
```

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,225 @@
...
+        this._renderer.appendChild(this._element.nativeElement, this._badgeElement);
+        this._renderer.removeStyle(this._badgeElement, 'display');
+    }
+
+    ngOnChanges(changes: SimpleChanges): void {
+        // if the badge is visible set changed values
+        if (this._badgeElement) {
+            // set badge content and get display friendly version of text based on max length and type of val
+            if (changes.badgeContent || changes.badgeMaxValue) {
+                let finalText: string = (changes.badgeContent && changes.badgeContent.currentValue) || this.badgeContent || null;
```
ashley-hunter, line 137: Can this be a const now too?

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,225 @@
...
+
+    private setBadgeBorderColor(): void {
+        if (this._badgeBorderColor) {
+            this._renderer.setStyle(this._badgeElement, 'border-color', this._badgeBorderColor.toHex());
+        } else {
+            this._renderer.removeStyle(this._badgeElement, 'border-color');
+        }
+    }
+
+    private setBadgeSize(previousValue: string): void {
```
ashley-hunter, line 196: Can this parameter be made optional instead of having to pass `null`. Perhaps `previousSize` might also be clearer.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+    ngOnDestroy(): void {
+        if (this._renderer.destroyNode) {
+            this._renderer.destroyNode(this._badgeElement);
+        }
+    }
+
+    private setContent(finalText: string, maxValue: number): void {
```
ashley-hunter, line 165: Perhaps `finalText` could now just be `content`?

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,225 @@
...
+        this.setBadgeBorderColor();
+        this.setBadgeSize(null);
+        this.setContent(this._badgeContent, this.badgeMaxValue);
+        this._renderer.appendChild(this._element.nativeElement, this._badgeElement);
+        this._renderer.removeStyle(this._badgeElement, 'display');
+    }
+
+    ngOnChanges(changes: SimpleChanges): void {
+        // if the badge is visible set changed values
+        if (this._badgeElement) {
```
ashley-hunter, line 134: To avoid having to nest everything, could we do this?

```typescript
if (!this._badgeElement) {
    return;
}
```

src/directives/badge/badge.directive.ts
darrenmcc-microfocus: The only issue with this is it doesn't allow for null.

darrenmcc-microfocus: If I add this before your logic it will work.

```suggestion
        if (!color) {
        return null;
        }
```

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,158 @@
+<div class="row badge-row">
+    <div class="col-sm-3 col-xs-6">
```
donna-marijne, line 2: I do think it would be better if these occupied the full width (`col-md-6`) and controls centered within the column.

e2e/pages/app/badge/badge.testpage.component.ts
```diff
@@ -0,0 +1,19 @@
...
+    colorCombinations: Map<string, string> = new Map<string, string>([
+        [ 'warning', 'critical' ],
+        [ 'grey5', 'grey1' ],
+        [ 'chart1', 'chart2' ],
+        [ 'chart3', 'chart4' ]
```
donna-marijne, line 17: As discussed earlier, maybe hex and rgb values in here as well?

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,158 @@
...
+
+<hr />
+
+<p>
+    The <code>uxBadge</code> directive can be added to any element to provide additional information as a badge, such as
+    status text or unread count.
+</p>
+<p><strong>Note:</strong> the badge directive relies on the CSS overflow property being set to <code>visible</code> for
+    the subject element due to how the badge element attaches to it. If you want to use <code>hidden</code>, you will
+    need to add a wrapper element around the subject so it only after the contents.</p>
```
donna-marijne, line 108: Typo here, "only after the contents."

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,158 @@
...
+        The color you want to use for the background color. This should be a string value that matches one of the colors
+        available in the <code>ColorService</code>.
+    </tr>
+    <tr uxd-api-property name="badgeBorderColor" type="string">
+        The color you want to use for the border of the badge. This should be a string value that matches one of
+        the colors available in the <code>ColorService</code>. If this value is not set the badge will have a transparent border.
+    </tr>
+    <tr uxd-api-property name="badgeValueLimit" type="number">
+        Sets a character or number limit on the content within the badge. If the content is a number and larger than
+        this value it will display this value with an appended<strong>+</strong>, otherwise the content will be treated
```
donna-marijne, line 124: Need to add whitespace before `<strong>+</strong>`

src/styles/notifications.less
```diff
@@ -108,4 +108,144 @@ ux-alert {
...
+        }
+    }
+
+    &.ux-badge-hidden .ux-badge {
+        display: none;
+    }
+}
+
+button.ux-badge-container:focus-within,
+button.ux-badge-container:focus {
```
ashley-hunter, line 234: Just a question. Does this only apply to buttons or could this also apply to links or any element with a tab index?

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
darrenmcc-microfocus: That was at the request of Roland. He mentioned it should be consistent with the other components.

e2e/pages/app/badge/badge.testpage.component.ts
darrenmcc-microfocus: I added a check to the karma tests. I will add another couple of examples of both in here too.

src/styles/notifications.less
darrenmcc-microfocus: Good point, this should be on any .badge-container.

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
donna-marijne: Did he give any example of what this style is consistent with? Looking through the widgets category we have examples that are [centered](https://pages.github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/#/ui-components/widgets#circular-progress) and examples that are [left-aligned](https://pages.github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/#/ui-components/widgets#spinner) so I don't really think we have anything to be consistent with.

darrenmcc-microfocus: I see what you mean. The majority of examples in widgets -> Microfocus and all in the notifications -> OpenSource are left aligned . Does this mean that the others should be updated for consistency sake, so all are left aligned? Or I can change this example back.

donna-marijne: No, I think it should be two columns, center aligned within each.

src/styles/notifications.less
```diff
@@ -108,4 +108,144 @@ ux-alert {
...
+            transform: translateX(50%);
+            right: 0;
+        }
+    }
+
+    &.ux-badge-hidden .ux-badge {
+        display: none;
+    }
+
+    &:focus-within,
```
ashley-hunter, line 230: I seem to be seeing the badge getting a focus outline?

image

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,158 @@
...
+        (<strong>&hellip;</strong>).
+    </tr>
+    <tr uxd-api-property name="badgeVerticalPosition" type="'above' | 'below'" defaultValue="'above'">
+        The vertical position of the badge in relation to its subject element.
+    </tr>
+    <tr uxd-api-property name="badgeHorizontalPosition" type="'before' | 'after'" defaultValue="'after'">
+        The horizontal position of the badge in relation to its subject
+        element.
+    </tr>
+    <tr uxd-api-property name="badgeOverlap" type="string" defaultValue="true">
```
ashley-hunter, line 135: Type here should be boolean

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,158 @@
...
+        This will determine whether the badge content will overlap its subject element. If overlap is
+        <code>false</code> the badge will start after the end of the subject element, otherwise
+        if it is set to <code>true</code> the badge horizontal center will be at the end of its subject element.
+    </tr>
+    <tr uxd-api-property name="badgeSize" type="string" defaultValue="'small' | 'medium' | 'large'">
```
ashley-hunter, line 140: The type here should be `'small' | 'medium' | 'large'` and the default value `medium`

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
donna-marijne: Let's think of it this way, if you left-align then you would have to add a significant left margin to allow testing of the "before" alignments with potentially large strings, so it's not actually going to be aligned with anything. (For that reason you should probably add `col-xs-12` in there as well, just in case.)

src/styles/notifications.less
donna-marijne: I think we've used something like this in the past to identify focusable elements: https://stackoverflow.com/a/30753870

Obviously stuff like iframe is irrelevant here.

docs/app/pages/components/components-sections/notifications/badge/snippets/app.html
```diff
@@ -0,0 +1,30 @@
+<div class="row">
+    <div class="col-sm-3 col-xs-6">
```
donna-marijne, line 2: This file needs to be kept in sync with the documentation example. I usually just copy and paste before committing.

src/styles/notifications.less
donna-marijne: Doesn't look like there's a way around it because the badge element is part of the button's region. I'd suggest removing this rule altogether, because our focus indicator does not have the same issue.

e2e/pages/app/badge/badge.testpage.component.html
```diff
@@ -0,0 +1,92 @@
...
+                </li>
+            </ng-container>
+        </ng-container>
+    </ul>
+    <hr />
+    <!-- size examples - using icons -->
+    <ul class="element-list">
+        <li *ngFor="let size of sizes">
+            <ux-icon name="mail" uxBadge="NEW" badgeVerticalPosition="above" badgeHorizontalPosition="after"
+                badgeSize="{{ size }}" badgeColor="critical"></ux-icon>
```
ashley-hunter, line 30: Just noticed there are a few places in this file where we are using `{{ }}` for attributes when we should use property binding instead eg: `[badgeSize]`

darrenmcc-microfocus: Sorry good spot. That's fixed now.

src/directives/badge/badge.directive.ts
```diff
@@ -0,0 +1,223 @@
+import { AfterViewInit, Directive, ElementRef, HostBinding, Input, OnChanges, OnDestroy, Renderer2, SimpleChanges } from '@angular/core';
+import { ColorService } from '../../services/color/color.service';
+import { ThemeColor } from '../../services/color/theme-color';
+import { ContrastService } from '../accessibility/contrast-ratio/contrast.service';
+import { timeHours } from 'd3';
```
ashley-hunter, line 5: Just noticed this import, I think this should be removed

src/directives/badge/badge.directive.ts
```diff
@@ -192,6 +192,8 @@ export class BadgeDirective implements AfterViewInit, OnChanges, OnDestroy {
         } else {
             this._renderer.removeStyle(this._badgeElement, 'border-color');
         }
+
+        this._renderer.setStyle(this._badgeElement, 'background-clip', this._badgeBorderColor ? 'padding-box' : 'border-box');
```
ashley-hunter, line 196: Can we use `padding-box`, doesn't look well supported?

image

ashley-hunter: Nevermind... I see this is background-clip, not box-sizing

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -50,13 +50,13 @@
...
                         <div class="input-group m-b-sm">
-                            <input type="text" class="form-control" [(ngModel)]="selectedColor">
+                            <input type="text" class="form-control" [(ngModel)]="selectedColor" [ngModelOptions]="{ updateOn: 'blur' }" />
                         </div>
                     </div>
                     <div class="col-md-6 col-sm-6">
                         <label for="badgeBorderColor">badgeBorderColor</label>
                         <div class="input-group m-b-sm">
-                            <input type="text" class="form-control" [(ngModel)]="selectedBorderColor">
+                            <input type="text" #userBorderColor class="form-control" [(ngModel)]="selectedBorderColor" [ngModelOptions]="{ updateOn: 'blur' }" />
```
ashley-hunter, line 59: I think `userBorderColor` can be removed here

src/directives/badge/badge.directive.ts
darrenmcc-microfocus: https://www.w3schools.com/CSSref/css3_pr_background-clip.asp

I think this is ok for back to IE9 and early versions of other browsers.

ashley-hunter: Yeah, I initially I thought it was box sizing being set, so just ignore this comment

docs/app/pages/components/components-sections/notifications/badge/badge.component.html
```diff
@@ -0,0 +1,158 @@
...
+<p><strong>Note:</strong> the badge directive relies on the CSS overflow property being set to <code>visible</code> for
+    the subject element due to how the badge element attaches to it. If you want to use <code>hidden</code>, you will
+    need to add a wrapper element around the subject so it only affects the contents.</p>
+<p>
+    The appearance and behavior of the directive can be configured using the following properties:
+</p>
+
+<uxd-api-properties tableTitle="Inputs">
+    <tr uxd-api-property name="badgeColor" type="string">
+        The color you want to use for the background color. This should be a string value that matches one of the colors
```
ashley-hunter, line 115: Do these colour inputs also support CSS colour values too? If so we should probably mention that.

e2e/pages/app/badge/badge.testpage.component.html
```diff
@@ -0,0 +1,92 @@
...
+    <hr />
+    <!-- hidden example -->
+    <a href="#/badge" uxBadge="7483" badgeMaxValue="99" badgeVerticalPosition="above" badgeHorizontalPosition="after"
+        badgeSize="medium" badgeColor="chart4" [badgeHidden]="true">
+        Hidden badge example
+    </a>
+    <hr />
+    <!-- color and border examples -->
+    <ul class="element-list">
+         <li *ngFor="let colors of colorCombinations.entries()">
```
ashley-hunter, line 87: We should use the key value pipe here instead of calling the entries function:

https://angular.io/api/common/KeyValuePipe

darrenmcc-microfocus: I didn't know about the K -> V pipe, nice!

docs/app/pages/components/components-sections/notifications/badge/badge.component.ts
```diff
@@ -12,6 +13,8 @@ import { IPlaygroundProvider } from '../../../../../interfaces/IPlaygroundProvid
 })
 @DocumentationSectionComponent('ComponentsBadgeComponent')
 export class ComponentsBadgeComponent extends BaseDocumentationSection implements IPlaygroundProvider {
+    colorPalleteDocumentationRoute: string;
```
ashley-hunter, line 16: Typo: `pallete` -> `palette`

docs/app/pages/components/components-sections/notifications/badge/badge.component.ts
```diff
@@ -0,0 +1,55 @@
...
+        ],
+    };
+
+    constructor(
+        @Inject(DOCUMENTATION_TOKEN) private readonly documentationType: DocumentationType,
+        private readonly colorService: ColorService
+    ) {
+        super(require.context('./snippets/', false, /\.(html|css|js|ts)$/));
+
+        this.colorPaletteDocumentationRoute = documentationType === DocumentationType.MicroFocus ? '/ui-components/styling' : '/css/color-palette';
```
donna-marijne, line 53: You also need to provide the fragment for the Micro Focus documentation, since this link displays the color service documentation.

PR #1326, opened 2020-04-28
EL-3760 - Fix tab `active` input

donna-marijne:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3760

* Add a `BehaviorSubject` to the TabsetService store the selected tab.
* Add `activeChange` to enable 2-way binding.
* Add karma tests for `active` and `activeChange`.
* Remove several duplicate tests from TabsetTestComponent.
* The original parent pom is apparently no longer accessible to this build, so I moved it on to the next release.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3760-tabset-active

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3760-tabset-active~CI/)


donna-marijne:
Thank you, I think this should resolve those problems.

src/components/tabset/tabset.component.html
```diff
@@ -8,10 +8,10 @@
     [attr.aria-label]="ariaLabel"
     [attr.aria-orientation]="stacked === 'none' ? 'horizontal' : 'vertical'">
 
-	<li role="presentation"
+    <li role="presentation"
         class="nav-item"
         *ngFor="let tab of _tabset.tabs; let index = index"
-        [class.active]="tab.active"
+        [class.active]="(_tabset.activeTab$ | async) === tab"
```
ashley-hunter, line 14: I just noticed one thing regarding focus. If i programmatically set the active tab, the focusable item remains on the previously selected tab.

Eg. Select the 4th tab, then click the button to activate the second tab, then use the keyboard to return focus to the tabset. The 4th tab will regain focus:

image

src/components/tabset/tab/tab.component.ts
```diff
@@ -10,13 +11,18 @@ let uniqueTabId = 0;
...
 
     /** Define the tab unique id */
     @Input() id: string = `ux-tab-${++uniqueTabId}`;
 
     /** Define the active state of this tab */
-    @Input() active: boolean = false;
+    @Input()
+    set active(active: boolean) {
+        if (active) {
+            this._tabset.setTabActive(this);
```
ashley-hunter, line 23: Just noticed one more case where I still get an expression changed error. See the plunker, on initial load press the button to select the second tab:

https://plnkr.co/edit/Wh4yzMnwocPKNT8q?preview

[Image from Gyazo](https://gyazo.com/d634c30ea83d75e7c4cf8dff9a1edaee)

src/components/tabset/tabset.component.html
donna-marijne: I'll take a look but I think this is a pre-existing bug.

ashley-hunter: Ah right, no problem. In that case feel free to raise a ticket and we can deal with it another time

src/components/tabset/tab/tab.component.ts
donna-marijne: Do you think using BehaviorSubject is the wrong approach here? We seem to have this problem every time we use rxjs in components. I've added in `tick` and it solves the problem as usual, but it feels like a hack to trigger an unnecessary extra round of change detection like that.

ashley-hunter: Perhaps moving it to `ngOnChanges` may help in this case? I'm not 100% sure this will fix the issue, but it could be worth a try. I agree that we seem to be having this problem quite a lot with setters, and perhaps it may be better to avoid them where possible.

donna-marijne: Unfortunately it's the same problem. I've also experimented with removing the observable altogether but there are still expression changed error.

Everything seems to work with the `tick` solution except for two karma tests:
```
HeadlessChrome 81.0.4044 (Windows 10.0.0) Tabset Component should change to new tab when tab is programmatically selected by tab index FAILED
        Error: Expected 'Schedule' to equal 'Solution'.
```
and
```
HeadlessChrome 81.0.4044 (Windows 10.0.0) Tabset Component should change to new tab when tab is programmatically selected by tab instance FAILED
        Error: Expected 'Schedule' to equal 'Solution'.
```

The usual `detectChanges`/`whenStable` routine does not seem to work in this case, which I don't really understand. The only way I got that assertion to pass is to put it behind a 10ms `setTimeout`. Do you have any ideas?

ashley-hunter: I'm guessing it is because the tick operator makes the change asynchronous, so it will be evaluated on the next Javascript cycle, whereas the test is expecting it to be synchronous. You should just be able to wrap the test in `fakeAsync` and call the `tick` function, eg:

```typescript
it('should emit activeChange output when tab is selected', fakeAsync(async () => {
    await clickTab(2);

    tick(1); // we may not even need the 1

   // expectation here
}));
```

donna-marijne: For some reason it needs an actual time delay, 100ms worked where 1ms did not.
```ts
    fit('should change to new tab when tab is programmatically selected by tab index', fakeAsync(async () => {

        const tab1 = getTabItem(0);
        const tab3 = getTabItem(2);

        // check first tab is active
        expect(tab1.classList.contains('active')).toBeTruthy();

        spyOn(component, 'onTabActivated');

        // programatically change tab by index
        component.tabset.selectTab(2);
        tick(100);

        // check second tab is now active/first tab is not active
        expect(tab3.classList.contains('active')).toBeTruthy();
        expect(tab1.classList.contains('active')).toBeFalsy();

        // check tab contains correct content
        expect(nativeElement.querySelector('.tab-pane[aria-hidden="false"] h4').innerHTML).toEqual('Solution');

        expect(component.onTabActivated).toHaveBeenCalled();
    }));
```

If it's just the next cycle surely it should not need to wait a specific amount of time?

ashley-hunter: Hmm, actually I thought the tick operator was using `Promise.resolve` which would fire on the next cycle, however it turns out it is using `requestAnimationFrame` so that could be a variable amount of time, but at 60fps it would likely be ~17ms, so that might be why there needs to be a delay?

donna-marijne: Thank you, it does seem reliable if not ideal for a unit-type test.

PR #1329, opened 2020-05-07
Input drop-down: make filter bi-directional

tobias-schreiner:
#### Checklist

#### Ticket / Issue
Input drop-down filter cannot be cleared programmatically

Make filter attribute bi-directional

#### Documentation CI URL
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/microfocusbbn_CI_UXAspects_input-dropdown-set-filter

#### Downstream Build(s)
https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf%7Eux-aspects-micro-focus%7Einput-dropdown-set-filter%7ECI/


tobias-schreiner:
We tested the new feature by automatically clearing the filter every ten seconds. (Button was not possible, since clicking outside closes the drop-down. The user needs to enable the automatic clear with the new check-box,) 

We were not sure if this should go into the documentation page or is too confusing for the user. The mechanism is also contained in the snippets, but most likely even less necessary there.

Please advise if and where we should keep it.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -68,6 +68,11 @@
                         <input type="text" class="form-control" [(ngModel)]="placeholder">
                     </div>
                 </div>
+                <div class="row uxd-customize-row">
+                    <div class="col-md-4 col-sm-12">
+                        <ux-checkbox [(value)]="resetFilter">Reset filter every 10 seconds</ux-checkbox>
```
donna-marijne, line 73: The customize section is mainly meant for binding actual input properties. This seems more like a test case than something informative to consumers.

I would suggest either leaving the example as it was, or a text input binding directly to the new `filter` input.

tobias-schreiner: Unfortunately, the drop-down closes as soon as the user clicks outside. Therefore, a filter text input does not help to see the feature. I have reverted the example.

donna-marijne:
We will aim to include this in 2.0.3, however I do not anticipate any further changes required to this PR. Thanks for your help on this.

PR #1332, opened 2020-05-11
EL-3902 - Column Picker - grouping unselected columns

darrenmcc-microfocus:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3902

Add in grouping to the table column picker

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3902-column-picker-grouping/#/components/tables

#### Downstream Build(s)

https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf%7Eux-aspects-micro-focus%7EEL-3902-column-picker-grouping%7ECI/


src/components/table/column-picker/column-picker.component.ts
```diff
@@ -41,39 +70,163 @@ export class ColumnPickerComponent {
...
     /** Emits when the selected items change */
-    @Output() selectedChange = new EventEmitter<ReadonlyArray<string>>();
+    @Output() selectedChange = new EventEmitter<ReadonlyArray<ColumnPickerValue>>();
 
     /** Emits when the deselected items change */
-    @Output() deselectedChange = new EventEmitter<ReadonlyArray<string>>();
+    @Output() deselectedChange = new EventEmitter<ReadonlyArray<ColumnPickerValue>>();
+
+    /** The Nested tree control used for the deselect tree */
+    treeControl: NestedTreeControl<ColumnPickerTreeNode> = new NestedTreeControl<ColumnPickerTreeNode>(item => item.children);
```
darrenmcc-microfocus, line 83: I tried the FlatTreeControl instead of nested first but there were a number of reservations I had to using it. The structure has to be laid out exactly as it's displayed - each node has a level and each node must also have a parent node. There is too much room for human error inputing the values. With the NestedTreeControl I can build the structure but still retain the original flat structure of elements and have control over how it is built.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -41,39 +70,163 @@ export class ColumnPickerComponent {
...
+    activeIndex: number = 0;
+
+    isFocused: boolean = false;
 
     /** Cache selection during reordering */
-    private _selection: ReadonlyArray<string> = [];
+    private _selection: ReadonlyArray<ColumnPickerValue> = [];
+
+    /** Focus manager for cycling through the tree nodes */
+    private _focusKeyManager: FocusKeyManager<CdkNestedTreeNode<ColumnPickerTreeNode>>;
```
darrenmcc-microfocus, line 106: This needs work. I'm having some issues with getting the keyboard events to trigger on all items and cycle through the heirarchy - it tends to only work on the currently focused level, so subnodes of a group or top level nodes only.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.ts
```diff
@@ -3,6 +3,7 @@ import { BaseDocumentationSection } from '../../../../../components/base-documen
 import { DocumentationSectionComponent } from '../../../../../decorators/documentation-section-component';
 import { IPlayground } from '../../../../../interfaces/IPlayground';
 import { IPlaygroundProvider } from '../../../../../interfaces/IPlaygroundProvider';
+import { ColumnPickerValue, ColumnPickerGroupSetting } from 'src/components/table';
```
ashley-hunter, line 6: This should be a relative import instead of absolute

src/common/utilities/key-events.ts
```diff
@@ -0,0 +1,7 @@
+export enum KeyCodes {
```
ashley-hunter, line 1: We don't need to create our own list of keycodes here, use `@angular/cdk/keycodes` instead.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -15,7 +44,7 @@ export class ColumnPickerComponent {
     @Input() locked: ReadonlyArray<string> = [];
 
     /** Define a list of columns that are not selected or locked */
-    @Input() deselected: ReadonlyArray<string> = [];
+    @Input() deselected: ReadonlyArray<ColumnPickerValue> = [];
```
ashley-hunter, line 47: I don't think we can change the type of this or the type of `selectedChange`. 
We should probably change this to `string | ColumnPickerValue` which should continue to use strings if people were previously using strings and ColumnPickerValue if it is being used. Changing this to always be `ColumnPickerValue` would be a breaking change.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -41,39 +70,180 @@ export class ColumnPickerComponent {
...
+                break;
+
+            // Select and deselect current item
+            case KeyCodes.Enter:
+                if (!this._deselectedSelection.find((deselectedSelectionItem: ColumnPickerTreeNode) => deselectedSelectionItem === node)) {
+                    this._deselectedSelection = [...this._deselectedSelection, node];
+                } else {
+                    this._deselectedSelection = this._deselectedSelection.filter((deselectedSelectionItem: ColumnPickerTreeNode) => node !== deselectedSelectionItem);
+                }
+                // console.log('updated _deselectedSelection: ', this._deselectedSelection);
```
ashley-hunter, line 234: This can be removed

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -187,6 +355,50 @@ export class ColumnPickerComponent {
...
+        return (column as ColumnPickerGroupItem).name !== undefined;
+    }
+
+    /** Get the column name based on type */
+    getColumnName(item: ColumnPickerValue) {
+        return this.isColumnPickerItem(item) ? item.name : item;
+    }
+
+    /** Check if tree group has visible children */
+    nodeHasChildren(index: number, node: ColumnPickerTreeNode): boolean {
```
ashley-hunter, line 369: Any functions that would be private but are being used in the view should be prefixed with an underscore to ensure that no consumer of the component tries to relying on these "private" methods

src/components/table/column-picker/column-picker.component.html
```diff
@@ -40,7 +71,7 @@
             <ux-icon name="chevron-left"></ux-icon>
         </button>
 
-        <button class="btn button-secondary btn-block" [disabled]="deselected.length === 0" (click)="addAllColumns()">
+        <button class="btn button-secondary btn-block" [disabled]="hasDeselectedItems()" (click)="addAllColumns()">
```
ashley-hunter, line 74: Is is possible to calculate this whenever selection changes and store the result in a variable? Ideally we want to avoid calling functions from a binding where possible as this can lead to performance problems

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -41,39 +70,180 @@ export class ColumnPickerComponent {
...
+                expandable: false
+            });
+        });
+
+        this.treeData = treeData;
+        this.treeDataSource = new ArrayDataSource(treeData);
+    }
+
+    ngAfterViewInit(): void {
+        this._focusKeyManager = new FocusKeyManager(this.treeNodes).withVerticalOrientation();
```
ashley-hunter, line 195: I believe you mentioned you were having some issues with keyboard control on standup? You might be able to simplify this by using our `uxTabbableList` directive?

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
```diff
@@ -79,6 +86,24 @@
...
+    <tr uxd-api-property name="name" type="string">
+        The name of the column.
+    </tr>
+</uxd-api-properties>
+
+<uxd-api-properties tableTitle="ColumnPickerGroupSetting Interface">
+    <tr uxd-api-property name="group" type="string">
+        The name of the group this setting object is related to.
+    </tr>
+    <tr uxd-api-property name="initiallyExpanded?" type="boolean">
```
donna-marijne, line 102: The `?` isn't part of the property name. I think the proper way of denoting this would be to add `defaultValue="false"`.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
```diff
@@ -45,13 +49,16 @@
     <tr uxd-api-property name="columnMovedAnnouncement" type="(column: string, delta: number) => string">
         Define a function that returns a column move announcement.
     </tr>
+    <tr uxd-api-property name="groupSettings" type="ReadonlyArray<ColumnPickerGroupSetting>">
+        Define a list of settings relating to the groups created for the deselected column.
```
donna-marijne, line 53: I would just add "See below for a listing of the available settings."

e2e/tests/components/column-picker/column-picker.e2e-spec.ts
```diff
@@ -31,6 +35,55 @@ describe('Column Picker Tests', () => {
...
     });
 
+    it('should allow a group to navigate, close and expand using mouse and keyboard', async () => {
+        // close the group
+        const firstGroup = await page.deselectedListGroupBtns.get(0);
+        await firstGroup.click();
+        expect(await imageCompare('column-picker-close-deselected-group-1')).toEqual(0);
+        // expand the group
+        await firstGroup.click();
+        expect(await imageCompare('column-picker-initial')).toEqual(0);
```
donna-marijne, line 45: I would suggest not reusing screenshots. If one fails and not the other, the output could be obscured by the test that runs later. This could make diagnosing failures difficult.

You could probably cut some of the screenshots out of this test without losing coverage. For example, you could assert that the number of elements in the deselected list is consistent with the group being expanded instead of this `imageCompare`.

e2e/tests/components/column-picker/column-picker.e2e-spec.ts
```diff
@@ -31,6 +35,55 @@ describe('Column Picker Tests', () => {
...
+
+        // press up enter to expand the group
+        await browser.actions().sendKeys(Key.ENTER).perform();
+        expect(await imageCompare('column-picker-close-expanded-group-1')).toEqual(0);
+    });
+
+    it('should allow a group to navigate close, reopen using the keyboard', async () => {
+        // close the group
+        const firstGroup = await page.deselectedListGroupBtns.get(0);
+        await firstGroup.click();
```
donna-marijne, line 80: This seems to do the same as the first part of the previous test case. Can it be removed.

e2e/tests/components/column-picker/column-picker.e2e-spec.ts
```diff
@@ -67,44 +120,32 @@ describe('Column Picker Tests', () => {
     });
 
     it('should allow moving a deselected column', async () => {
-        expect(await page.getColumnCount(ColumnPickerList.Deselected)).toBe(18);
-        expect(await page.getColumnCount(ColumnPickerList.Selected)).toBe(6);
```
donna-marijne, line 71: Perhaps these counts should be updated rather than removed?

e2e/tests/components/column-picker/column-picker.e2e-spec.ts
```diff
@@ -67,44 +120,32 @@ describe('Column Picker Tests', () => {
...
         await page.selectColumn(ColumnPickerList.Deselected, 0);
         await page.selectBtn.click();
         expect(await page.getColumnCount(ColumnPickerList.Deselected)).toBe(17);
         expect(await page.getColumnCount(ColumnPickerList.Selected)).toBe(7);
-        expect(await page.getDeselection()).toBe('[ "Category", "Date Created", "Date Modified", "Department", "Document ID", "Flag", "From", "Icon", "Importance", "Location", "Location ID", "Message", "Organization", "Time", "Time Created", "Time Modified", "Work Completed" ]');
```
donna-marijne, line 76: This should probably not be removed.

src/components/table/column-picker/column-picker.component.html
```diff
@@ -12,39 +12,84 @@
...
-        </div>
-    </div>
+            <div *ngIf="_nodeHasAvailableChildren(node)"
+                 class="column-picker-tree-group-node">
+
+                <button uxTabbableListItem
+                        (click)="_setNodeExpanded(node, !node.isExpanded)"
+                        (keydown.arrowright)="_setNodeExpanded(node, true)"
+                        (keydown.arrowleft)="_setNodeExpanded(node, false)"
+                        [attr.aria-label]="'Toggle ' + node.name"
```
donna-marijne, line 49: If you are applying an aria-label within a component then we need to expose a localization hook. However, I believe if you apply the correct roles and aria attributes, and use the appropriate keyboard conventions, the label may not be necessary. Also this button isn't part of the tab ring, so this label will not be encountered.

I suggest reviewing this ARIA practices document and making sure this matches as closely as possible:
https://www.w3.org/TR/wai-aria-practices/examples/treeview/treeview-2/treeview-2a.html

If you haven't already, install ChromeVox for testing.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -41,60 +43,174 @@ export class ColumnPickerComponent {
     /** Define a function that return a column move announcement */
     @Input() columnMovedAnnouncement: (column: string, delta: number) => string = this.getColumnMovedAnnouncement;
 
+    /** group settings for customising display */
```
donna-marijne, line 46: We should aim to reuse the text from the documentation for doc comments, it will help if we ever get round to generating docs from source.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -41,60 +43,174 @@ export class ColumnPickerComponent {
...
 
+    ngOnChanges(changes: SimpleChanges): void {
+        // recreate tree when deselected changes
+        if (changes.deselected && changes.deselected.currentValue !== changes.deselected.previousValue) {
+            this._rebuildDeselectTree();
+        }
+    }
+
+    /** Parse data into suitable format for the FlatTreeComponent to understand and initialize deselect tree */
+    _rebuildDeselectTree(): void {
```
donna-marijne, line 97: I believe this can be made `private`. Check for the other functions as well, they only need to be public if they are either part of the API of the component, or used directly in the view.

e2e/tests/components/column-picker/column-picker.e2e-spec.ts
darrenmcc-microfocus: The issue with this is that the deselected always stores all items and the component just hides and reveals the relevant one. So this list never changes. The assertion of the counts combined with the selected column will be a clear enough indication it is working correctly. Let me know if you disagree - alternatively I can write something that checks a list of column names exist in the deselected list.

src/components/table/column-picker/column-picker.component.html
darrenmcc-microfocus: Sorry I just copied and pasted their example, didn't give this enough thought so I need to be mindful in future. I'll get it fixed to match the guildlines.

darrenmcc-microfocus: I've went through the guidelines and it will not be possible to follow it fully - we would need to use the CdkNestedTree to follow it fully and that will also lead to more issues with keyboard navigation. The only part missing is the ability to set `role="group"` on the child items. It will mean that the audio won't report that a group is anymore than a button. I implemented a overridable function in the same way that the selected reorder has been implemented so that it is clear the group items are collapsable / expandable.

donna-marijne: If you check the DOM you can see the `role` has already been set for you by the CDK tree, so that's not a problem.

src/components/table/column-picker/column-picker.component.html
```diff
@@ -12,39 +12,86 @@
...
-    </div>
+            <div *ngIf="_nodeHasAvailableChildren(node)"
+                class="column-picker-tree-group-node">
+
+                <button uxTabbableListItem
+                        (click)="_setNodeExpanded(node, !node.isExpanded)"
+                        (keydown.arrowright)="_setNodeExpanded(node, true)"
+                        (keydown.arrowleft)="_setNodeExpanded(node, false)"
+                        [style.visibility]="node.expandable ? 'visible' : 'hidden'"
+                        [attr.aria-label]="getDeselectedAriaLabel(node)"
```
donna-marijne, line 52: Just had another look at this and it's not quite right. It should be:
```
[attr.aria-label]="deselectedAriaLabel(node)"
```
Otherwise any custom function provided by the application will be ignored. The `getDeselectedAriaLabel` function should be renamed to `getDefaultDeselectedAriaLabel` and made private.

e2e/tests/components/column-picker/column-picker.e2e-spec.ts
donna-marijne: No I think that's fine, but if that JSON is not being used in the tests then it should be removed from the test page.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
```diff
@@ -46,11 +46,14 @@
     <tr uxd-api-property name="selectedAriaLabel" type="(column: string, index: number) => string">
         Define a function to get the aria label of reorderable items in the selected column.
     </tr>
+    <tr uxd-api-property name="deselectedAriaLabel" type="(node: ColumnPickerTreeNode) => string">
```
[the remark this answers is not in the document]
darrenmcc-microfocus: I've realised this function is only used for grouped items so expandable isn't even needed. Just `column` and `isExpandable`.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
```diff
@@ -79,6 +89,42 @@
...
+    <tr uxd-api-property name="name" type="string">
+        The name of the column.
+    </tr>
+</uxd-api-properties>
+
+<uxd-api-properties tableTitle="ColumnPickerGroupSetting Interface">
+    <tr uxd-api-property name="group" type="string">
+        The name of the group this setting object is related to.
+    </tr>
+    <tr uxd-api-property name="initiallyExpanded" type="boolean">
```
ashley-hunter, line 105: What are the thoughts on this `groupSettings` object changing over time. Eg:

1. The user may want to programmatically expand a group after initial load, in which case we need to add some logic to ngOnChanges to perform this change when the input changes.
2. Could the user ever want to know when the expanded state changes? In which case an event that is emitted on change with support for two way binding?

Regardless of whether we implement this now or not, we probably should rename `initiallyExpanded` to just `expanded` as this would allow us to make these changes at some point in the future without having to rename this property.

Thoughts? @donna-marijne

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -388,21 +396,30 @@ export interface ColumnPickerActionsContext {
...
 /** An interface representing settings of groups defined in ColumnPickerGroupItem objects */
 export interface ColumnPickerGroupSetting {
+    // The name of the group this setting object is related to.
     group: string;
+    // Defines if this group will be expanded on load. This is an optional property.
     initiallyExpanded?: boolean;
 }
 
 /** Interface representing a tree node item. This normalises data into one format */
 export class ColumnPickerTreeNode {
```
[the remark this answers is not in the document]
darrenmcc-microfocus: Sorry this should be an interface, I'm not sure why I changed it.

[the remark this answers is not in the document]
darrenmcc-microfocus: @donna-marijne  As for the comment. It was following the pattern of matching what was in the docs but this has been removed from the docs since.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -388,21 +396,30 @@ export interface ColumnPickerActionsContext {
...
+    // The name of the column or group.
```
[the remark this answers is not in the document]
darrenmcc-microfocus: @donna-marijne That style of comment is for functions only - well at least according to the linter in VS code. My understanding was that we wanted to document the code as much as possible for the end user's benefit. That would mean that I'd only be able to comment above the function and have remove all the rest. Please let me know you thoughts.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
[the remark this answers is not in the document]
donna-marijne: I think it should either apply to all nodes, or be renamed to clarify that it only applies to group headers.

src/components/table/column-picker/column-picker.component.html
darrenmcc-microfocus: Good point that's a mistake on my behalf. I've followed the same pattern for the selectedAriaLabel to keep them consistent.

src/components/table/column-picker/column-picker.component.ts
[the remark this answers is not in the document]
donna-marijne: I think it's correct as a class since it's only used internally. Clean Code is on OReilly.

src/components/table/column-picker/column-picker.component.ts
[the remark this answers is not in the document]
donna-marijne: Which linter are you using? There is absolutely no rule that restricts doc-comments to functions. In this file there are dozens of doc comments on properties.

```
/** A tree-friendly representation of the deselected data */
_treeData: ColumnPickerTreeNode[];
```

> That would mean that I'd only be able to comment above the function and have remove all the rest.

Could you rephrase this because I'm not sure what it means?

src/components/table/column-picker/column-picker.component.ts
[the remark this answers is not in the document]
darrenmcc-microfocus: @donna-marijne Ok sure changed back and the comment updated.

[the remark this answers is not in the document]
ashley-hunter: The only downside with a class is that it is a runtime construct so will be included in the JavaScript output, whereas an interface is a compile time construct only and not included in any build output, so its slightly better for us to use interfaces unless we need any methods etc.. as we don't increase our bundle size unnecessarily.

src/components/table/column-picker/column-picker.component.ts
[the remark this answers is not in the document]
darrenmcc-microfocus: @donna-marijne  The code snippet shown was confusing I thought you meant inner comments but I see I used `//` on one of the function comments. Fixed now, good spot!

src/components/table/column-picker/column-picker.component.html
donna-marijne: Yes, looks like that one is wrong as well, but we can fix that another time.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
donna-marijne: It's certainly a possible use case. I think if we were planning to ever make this collection more interactive then it would be better to drop "settings" from the name and associated type. In the future we could add a `groupsChange` output when the state changes.

src/components/table/column-picker/column-picker.component.ts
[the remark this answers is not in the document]
donna-marijne: Ah, I did not think of that. I think there are a few functions which ought to be refactored from `ColumnPickerComponent` to `ColumnPickerTreeNode`. However, realistically speaking we will never come back to refactor this, so it might as well be an interface.

src/components/table/column-picker/column-picker.component.ts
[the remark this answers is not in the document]
donna-marijne: To be clear, I am talking about this and other class/interface properties, not functions.

e2e/tests/components/column-picker/column-picker.e2e-spec.ts
darrenmcc-microfocus: It is still used for the initial check so I think it can stay.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
darrenmcc-microfocus: @ashley-hunter @donna-marijne I agree. I'll rename the property mentioned above.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
```diff
@@ -42,16 +46,22 @@
     <tr uxd-api-property name="selectedAriaLabel" type="(column: string, index: number) => string">
         Define a function to get the aria label of reorderable items in the selected column.
     </tr>
+    <tr uxd-api-property name="deselectedAriaLabel" type="(column: string, isExpanded: boolean) => string">
```
donna-marijne, line 49: I think you were going to rename this one?

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
```diff
@@ -42,16 +46,22 @@
...
+        Define a function to get the aria label of a group in the deselected list.
+    </tr>
     <tr uxd-api-property name="columnMovedAnnouncement" type="(column: string, delta: number) => string">
         Define a function that returns a column move announcement.
     </tr>
+    <tr uxd-api-property name="groups" type="ReadonlyArray<ColumnPickerGroup>">
+        See below for a listing of the available settings for groups.
```
donna-marijne, line 56: Needs a description, something like "An optional array of objects to configure column groups."

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.ts
```diff
@@ -1,4 +1,5 @@
 import { Component } from '@angular/core';
+import { ColumnPickerGroupItem, ColumnPickerGroup } from '../../../../../../../src/components/table';
```
donna-marijne, line 2: I suspect this should be `from '@ux-aspects/ux-aspects'`.

e2e/pages/app/column-picker/column-picker.testpage.component.html
```diff
@@ -3,12 +3,14 @@
     [deselectedTitleTemplate]="showCustomTitles ? title : null"
     [(selected)]="selected"
     [(deselected)]="deselected"
+    [(groups)]="groups"
```
donna-marijne, line 6: There's no `groupChange` so this should be `[groups]="..."`.

e2e/pages/app/column-picker/column-picker.testpage.component.html
```diff
@@ -3,12 +3,14 @@
...
     [locked]="locked">
 </ux-column-picker>
 
 <ng-template #title>Custom Title</ng-template>
 
 <p id="selected-list">{{ selected | json }}</p>
 <p id="deselected-list">{{ deselected | json }}</p>
+<p id="group-settings-list">{{ groups | json }}</p>
```
donna-marijne, line 14: This will always be the value provided by the test case, so it doesn't provide any coverage.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -203,14 +372,54 @@ export class ColumnPickerComponent {
...
+/** An interface representing settings of groups defined in ColumnPickerGroupItem objects */
+export interface ColumnPickerGroup {
+    // The name of the group this setting object is related to.
+    name: string;
+    // Defines if this group will be expanded on load. This is an optional property.
+    expanded?: boolean;
+}
+
+/** Class representing a tree node item. Normalises data for both groups and columns into one format */
+export class ColumnPickerTreeNode {
```
donna-marijne, line 414: This should not be exported.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -203,14 +372,54 @@ export class ColumnPickerComponent {
...
+    // The name of the column or group.
```
donna-marijne, line 415: Change to doc comments, or remove altogether - as previously requested.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
darrenmcc-microfocus: Sorry I didn't update it in the docs. Fixed now.

src/components/table/column-picker/column-picker.component.ts
ashley-hunter, line 414: I think we also decided to change this to an interface to avoid any unnecessary increase to our bundle size

src/components/table/column-picker/column-picker.component.ts
darrenmcc-microfocus: As this was changed to an interface changes are not required.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -203,14 +372,54 @@ export class ColumnPickerComponent {
...
+
+/** An interface representing settings of groups defined in ColumnPickerGroupItem objects */
+export interface ColumnPickerGroup {
+    // The name of the group this setting object is related to.
+    name: string;
+    // Defines if this group will be expanded on load. This is an optional property.
+    expanded?: boolean;
+}
+
+/** Class representing a tree node item. Normalises data for both groups and columns into one format */
```
ashley-hunter, line 413: Just need to update the comment here as it talks about a `class`. Perhaps `Represents a tree node item....` would be better and that way it is less likely to get out of date

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
```diff
@@ -46,14 +46,14 @@
...
-    <tr uxd-api-property name="deselectedAriaLabel" type="(column: string, isExpanded: boolean) => string">
+    <tr uxd-api-property name="deselectedGroupAriaLabel" type="(column: string, isExpanded: boolean) => string">
         Define a function to get the aria label of a group in the deselected list.
     </tr>
     <tr uxd-api-property name="columnMovedAnnouncement" type="(column: string, delta: number) => string">
         Define a function that returns a column move announcement.
     </tr>
     <tr uxd-api-property name="groups" type="ReadonlyArray<ColumnPickerGroup>">
-        See below for a listing of the available settings for groups.
+        An optional array of objects to configure column groups.
```
donna-marijne, line 56: Add "See below for a listing of the available settings for groups" back in again.

docs/app/pages/components/components-sections/tables/column-picker/column-picker.component.html
```diff
@@ -53,7 +53,7 @@
         Define a function that returns a column move announcement.
     </tr>
     <tr uxd-api-property name="groups" type="ReadonlyArray<ColumnPickerGroup>">
-        An optional array of objects to configure column groups.
+        See below for a listing of the available settings for groups.
```
donna-marijne, line 56: I'll just fix it.

darrenmcc-microfocus: Sorry I thought I just did.

PR #1353, opened 2020-06-15
EL-3834 - Add in number picker precision fix and tests

darrenmcc-microfocus:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3834

Fix the precision issue in number picker - numbers above 1000000 do not work at present.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3834-number-picker-precision-bug

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3834-number-picker-precision-bug~CI/)


src/components/number-picker/number-picker.component.ts
```diff
@@ -207,5 +211,10 @@ export class NumberPickerComponent implements ControlValueAccessor, OnDestroy, O
         this._propagateChange(value);
     }
 
+    /** Determine precision based on the length of passed number (number of digits) */
+    determinePrecision(value: number): number {
```
ashley-hunter, line 215: By doing this this completely ignores the value the user may have provided as the precision input, it will always either be the default precision or large precision.

Is there an issue with just setting the default precision value to the large precision? That way the consumer is free to override it if they need?

docs/app/pages/components/components-sections/input-controls/number-picker/number-picker.component.html
```diff
@@ -48,7 +48,7 @@
         Can be used to show a red outline around the input to indicate an invalid value. By default the
         error state will appear if the user enters a number below the minimum value or above the maximum value.
     </tr>
-    <tr uxd-api-property name="precision" type="number" defaultValue="6">
+    <tr uxd-api-property name="precision" type="number" defaultValue="6 (maximum 6 digits) or JavaScript MAX_SAFE_INTEGER length">
```
donna-marijne, line 51: This text doesn't below in the defaultValue field.

src/components/number-picker/number-picker.component.ts
```diff
@@ -10,6 +10,9 @@ export const NUMBER_PICKER_VALUE_ACCESSOR: any = {
     multi: true
 };
 
+export const LARGE_INT_PRECISION: number =  Number.MAX_SAFE_INTEGER.toString().length - 1;
+export const DEFAULT_PRECISION: number = 6;
```
donna-marijne, line 14: I don't think either of these should be exported.

src/components/number-picker/number-picker.component.ts
darrenmcc-microfocus: Yes for some reason that didn't work for me when I tested it the first time around. I've re-evaluated the solution.

src/components/number-picker/number-picker.component.ts
darrenmcc-microfocus: @donna-marijne. I've moved these tests to Karma, I just thought testing the interaction may be difficult in karma tests but I was wrong. I'm thinking it could be a worthwhile exercise to move all the increment/decrement tests into the karma tests at the same time to speed up e2e running. What do you think?

src/components/number-picker/number-picker.component.ts
```diff
@@ -44,7 +43,18 @@ export class NumberPickerComponent implements ControlValueAccessor, OnDestroy, O
     @Input('aria-labelledby') labelledBy: string;
 
     /** Define the precision of floating point values */
-    @Input() precision: number = DEFAULT_PRECISION;
+    private _precision: number = null;
+
+    @Input()
+    get precision(): number {
```
ashley-hunter, line 49: Is this setter/getter needed? We want to avoid them where possible as they are much slower than a simple variable value. Normally when it comes to inputs we just set a default initial value, which will always be used unless they override it which would simplify this to just:

```typescript
@Input() precision: number = Number.MAX_SAFE_INTEGER.toString().length - 1;
```

docs/app/pages/components/components-sections/input-controls/number-picker/number-picker.component.html
```diff
@@ -48,8 +48,8 @@
         Can be used to show a red outline around the input to indicate an invalid value. By default the
         error state will appear if the user enters a number below the minimum value or above the maximum value.
     </tr>
-    <tr uxd-api-property name="precision" type="number" defaultValue="6 (maximum 6 digits) or JavaScript MAX_SAFE_INTEGER length">
-        Define the precision of floating point values.
+    <tr uxd-api-property name="precision" type="number" defaultValue="null">
```
ashley-hunter, line 51: Normally we just omit the default value if it is null. In this case it won't be `null` if its set the the length of a maximum number by default, however - in this case its not a simple data type, so perhaps just leave the defaultValue off?

e2e/protractor.config.js
```diff
@@ -43,7 +43,7 @@ exports.config = {
   afterLaunch: onAfterLaunch,
 
   // Spec patterns are relative to this config file
-  specs: ['./tests/**/**/*.e2e-spec.ts'],
+  specs: ['./tests/**/**/number-picker.e2e-spec.ts'],
```
ashley-hunter: I don't think this should be commited

src/components/number-picker/number-picker.component.ts
darrenmcc-microfocus: I used the getter setter here to default to the MAX_SAFE_INTEGER when the precision is unset. I'll remove it now if that is not a behaviour we want to follow.

src/components/number-picker/number-picker.component.spec.ts
```diff
@@ -3,15 +3,18 @@ import { async, ComponentFixture, TestBed } from '@angular/core/testing';
 import { FormBuilder, FormGroup, FormsModule, ReactiveFormsModule, Validators } from '@angular/forms';
 import { NumberPickerModule } from './number-picker.module';
 
+/** Basic number picker example */
```
donna-marijne, line 6: Don't like this description. Each test fixture has a specific purpose; this one tests the number picker in a reactive forms context.

Try not to add comments unless they add real value.

e2e/tests/components/number-picker/number-picker.po.spec.ts
```diff
@@ -1,10 +1,12 @@
-import { browser, by, element, ElementFinder } from 'protractor';
+import { browser, by, element, ElementFinder, ExpectedConditions } from 'protractor';
```
donna-marijne, line 1: Presumably this can be reverted too.

docs/app/pages/components/components-sections/input-controls/number-picker/number-picker.component.html
```diff
@@ -48,8 +48,8 @@
         Can be used to show a red outline around the input to indicate an invalid value. By default the
         error state will appear if the user enters a number below the minimum value or above the maximum value.
     </tr>
-    <tr uxd-api-property name="precision" type="number" defaultValue="6">
-        Define the precision of floating point values.
+    <tr uxd-api-property name="precision" type="number">
```
donna-marijne, line 51: Maybe adding `defaultValue="15"` here would be less confusing than the current description.

src/components/number-picker/number-picker.component.ts
donna-marijne: The current implementation has that behavior, FYI.

PR #1359, opened 2020-06-18
EL-3956 - Marquee Wizard: visitedChange not working

darrenmcc-microfocus:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3956

visitedChange is not being emitted when the visited state is changed by changing the valid state.

visitedChange should not be emitted when the setter for visited is invoked.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3956-marquee-wizard-visitedchange

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3956-marquee-wizard-visitedchange~CI/)


src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -1,11 +1,13 @@
...
+import { CommonModule } from '@angular/common';
+import { ChangeDetectorRef, Component, OnDestroy } from '@angular/core';
 import {
     async,
     ComponentFixture,
     fakeAsync,
     TestBed,
     tick
 } from '@angular/core/testing';
+import { StepChangingEvent } from '../wizard';
```
ashley-hunter, line 10: I think this should be `../wizard/index`

src/components/wizard/wizard-step.component.ts
```diff
@@ -78,4 +81,22 @@ export class WizardStepComponent {
...
+
+    setValid(value: boolean): void {
+        if (this._valid === value) {
+            return;
+        }
+
+        const changeVisited = (value && this._valid) || !this._valid;
+        this._valid = value;
+
+        if (changeVisited) {
```
ashley-hunter, line 98: I don't think there is a need for this variable here, consider just in-lining the expression

src/components/marquee-wizard/marquee-wizard-step.component.ts
```diff
@@ -51,4 +36,10 @@ export class MarqueeWizardStepComponent extends WizardStepComponent {
         this.completed = completed;
         this.completedChange.emit(completed);
     }
+
+    ngOnChanges(changes: SimpleChanges): void {
```
ashley-hunter, line 40: Could this cause any issues if someone was to programmatically change the valid value rather than using the Input? As this will not get called?

darrenmcc-microfocus: Good spot, this won't work. I'll find a more suitable solution

src/components/wizard/wizard.component.ts
```diff
@@ -250,15 +250,15 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
+        this.onFinishing.emit();
 
         // Disable the button while waiting on validation
         this.finishDisabled = true;
 
         try {
             // Fetch validation status
             const validationResult = this.isStepValid();
-            this.getCurrentStep().valid = validationResult instanceof Promise ? await validationResult : validationResult;
+            this.getCurrentStep()._valid = validationResult instanceof Promise ? await validationResult : validationResult;
```
ashley-hunter, line 261: Should this be calling the `setValid` function? I know there is logic involved in setting the valid state, just wondering if that logic needs to be run when this change is made? 

Also, I don't know for sure, but I imagine this could change the valid state, in which case we would need to emit the latest value to the user as well if it has changed?

src/components/wizard/wizard.component.ts
```diff
@@ -357,7 +357,7 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
 
         // if there is no validator then return the valid state
         if (!currentStep.validator) {
-            return currentStep.valid;
+            return currentStep._valid;
```
ashley-hunter, line 360: I think here is probably good to use the public `valid` property. We largely only want to use the private one inside the component it is a part of. We can use setters/getters in code fine, its mainly in templates we want to avoid them as change detection can end up calling them many many times.

src/components/wizard/wizard.component.ts
```diff
@@ -195,9 +195,9 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
         }
 
         // check if current step is invalid
-        if (!step.valid) {
+        if (!step._valid) {
```
ashley-hunter, line 198: I think `valid` would be correct here, see comment with explanation below.

src/components/wizard/wizard.component.spec.ts
```diff
@@ -249,3 +250,107 @@ describe('Wizard with validator', () => {
...
+        nativeElement = fixture.nativeElement;
+        visitedChanged = spyOn(component, 'visitedChanged');
+
+        fixture.detectChanges();
+        await fixture.whenStable();
+        fixture.detectChanges();
+        await fixture.whenStable();
+    });
+
+    it('should trigger a visitedChange event when valid modified on the current step, when other steps ahead are visited', async () => {
```
ashley-hunter, line 327: Could we also add a test perhaps testing programmatically setting the valid state? I know its fairly common for teams to programmatically modify step properties on the wizard so it would probably be good to test it

src/components/wizard/wizard-step.component.ts
```diff
@@ -20,10 +20,19 @@ export class WizardStepComponent {
     @Input() disableNextWhenInvalid: boolean | undefined;
 
     /** Allows you to define whether or not a step is valid. The user will not be able to proceed to the next step if this property has a value of false. */
-    @Input() valid: boolean = true;
+    _valid: boolean = true;
+
+    @Input()
```
ashley-hunter, line 25: Could we add a comment above this input explaining its purpose, it helps editors show a bit of information about the input without users having to consult the documentation

src/components/wizard/wizard.component.spec.ts
ashley-hunter: We can ignore this now that we are no longer relying on `ngOnChanges`

src/components/wizard/wizard.component.spec.ts
```diff
@@ -249,3 +257,111 @@ describe('Wizard with validator', () => {
...
+        // valid true and no call of visitedChange
+        await clickButton(WizardSelectors.ToggleValidity);
+
+        // valid now false and should trigger visitedChange
+        await clickButton(WizardSelectors.ToggleValidity);
+
+        expect(visitedChanged).toHaveBeenCalledTimes(1);
+    });
+
+    it('should update a valid state when the value is modified programatically', async () => {
```
ashley-hunter, line 357: I think this can be removed now

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -273,3 +281,108 @@ describe('Marquee wizard with delayed step creation', () => {
...
+
+        visitedChanged.calls.reset();
+
+        // valid true and no call of visitedChange
+        await clickButton(MarqueeWizardSelectors.ToggleValidity);
+
+        // valid now false and should trigger visitedChange
+        await clickButton(MarqueeWizardSelectors.ToggleValidity);
+
+        expect(visitedChanged).toHaveBeenCalledTimes(1);
```
ashley-hunter, line 379: We should also check the value that it was called with, so that it is emitting the correct value. Same for the wizard test

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -273,3 +281,108 @@ describe('Marquee wizard with delayed step creation', () => {
...
+        nativeElement = fixture.nativeElement;
+        visitedChanged = spyOn(component, 'visitedChanged');
+
+        fixture.detectChanges();
+        await fixture.whenStable();
+        fixture.detectChanges();
+        await fixture.whenStable();
+    });
+
+    it('should trigger a visitedChange event when valid modified on the current step, when other steps ahead are visited', async () => {
```
ashley-hunter, line 359: https://plnkr.co/edit/dnuBVz3Z9BIyfey9

When I try this here I don't see visitedChange getting emitted. So for example, I open the marquee wizard, hit next until I get to a step near the end. Now I navigate back to the second step, mark it as invalid. The other steps remain visited and don't seem to get reset?

Am I misunderstanding this? Or doing something wrong?

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -273,3 +282,137 @@ describe('Marquee wizard with delayed step creation', () => {
...
+        await clickButton(MarqueeWizardSelectors.ToggleValidity);
+
+        await clickButton(MarqueeWizardSelectors.NextButton);
+
+        expect(visitedChanged).toHaveBeenCalledWith(false);
+        expect(visitedChanged).toHaveBeenCalledTimes(1);
+
+        const stepsList = component.stepsList.toArray();
+        // step 1 should be valid and visited
+        expect(stepsList[0]._valid).toBeTruthy();
```
ashley-hunter, line 391: Can we use the `valid` property instead as we want to test the public API, not the private stuff. We do this for multiple reasons:

1) We want to use the same properties as the consuming application would be using to make sure they work
2) It can help prevent breaking changes, if our tests are relying on the public properties, if we were to remove them or change their data type (ie. breaking change) then the test will throw compilation errors
3) Private APIs are allowed to be changed at any time, apps shouldn't be using them so we can making breaking changes on them any time without affecting users. By relying on private APIs in the tests this means anytime we change the private implementation we will have to update the tests, which in turn means are tests are likely not going to catch any bugs introduced by the changes.

src/components/wizard/wizard.component.spec.ts
```diff
@@ -249,3 +258,131 @@ describe('Wizard with validator', () => {
...
+        await clickButton(WizardSelectors.ToggleValidity);
+
+        await clickButton(WizardSelectors.NextButton);
+
+        expect(visitedChanged).toHaveBeenCalledWith(false);
+        expect(visitedChanged).toHaveBeenCalledTimes(1);
+
+        const stepsList = component.stepsList.toArray();
+        // step 1 should be valid and visited
+        expect(stepsList[0]._valid).toBeTruthy();
```
ashley-hunter, line 366: Use public API as mentioned above

src/components/wizard/wizard.component.ts
```diff
@@ -347,6 +349,28 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
         return this.steps.toArray()[index];
     }
 
+    /**
+     * If a step in the wizard becomes invalid, all steps sequentially after
+     * it, should become unvisited and incomplete
+     */
+    setNextStepsUnvisited(currentStep: WizardStepComponent, cb?: (step: WizardStepComponent | MarqueeWizardStepComponent) => void): void {
```
ashley-hunter, line 356: We don't want this to be a public method, mark it as `private` if it is only used in this class and `protected` if it is also used in marquee wizard

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -273,3 +282,137 @@ describe('Marquee wizard with delayed step creation', () => {
...
+        expect(stepsList[0].visited).toBeTruthy();
+        expect(stepsList[0].completed).toBeTruthy();
+
+        // step 2 should be invalid and not visited
+        expect(stepsList[1]._valid).toBeFalsy();
```
ashley-hunter, line 396: Could we perhaps add a checks to see if the `visitedChange` has emitted on these steps correctly?

src/components/wizard/wizard.component.ts
```diff
@@ -347,6 +349,28 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
+        const steps = this.steps.toArray();
+        const affected = steps.slice(this.step + 1);
+
+        affected.forEach(step => {
+
+            // if the step is not the current step then also mark it as unvisited
+            if (step !== currentStep) {
+                step.visited = false;
```
ashley-hunter, line 364: Is this ever going to emit visitedChange? Would we need to use the `setVisited` function here instead? As this is the component changing the visited state and not the user we would need to ensure that we emit the new value.

src/components/wizard/wizard.component.ts
```diff
@@ -353,15 +353,17 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
      * If a step in the wizard becomes invalid, all steps sequentially after
      * it, should become unvisited and incomplete
      */
-    setNextStepsUnvisited(currentStep: WizardStepComponent, cb?: (step: WizardStepComponent | MarqueeWizardStepComponent) => void): void {
+    protected _setNextStepsUnvisited(currentStep: WizardStepComponent, cb?: (step: WizardStepComponent | MarqueeWizardStepComponent) => void): void {
```
ashley-hunter, line 356: We can remove the underscore prefix here. There are only two times we prefix things with an underscore:

1) Private class variables. Often class variables will have public setters/getters to to retain the same name the setter would be called `valid` and the private variable would be `_valid` keeping the association very obvious. To be consistent and support any future setter/getter we prefix all private variables with an underscore.

2) Methods that should be private or protected but can't be because they are used within a template.

In this case the function is marked as protected so typescript will prevent anyone from calling it, so we can omit the underscore prefix.

src/components/wizard/wizard.component.ts
ashley-hunter: Did we determine if we needed to emit validChange here?

PR #1364, opened 2020-06-26
EL-3987 - wizard visitedchange

darrenmcc-microfocus:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3987

visitedChange is not being emitted when the visited state is changed by changing the valid state.
visitedChange should not be emitted when the setter for visited is invoked.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3987-wizard-visitedchange

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3987-wizard-visitedchange~CI/)


src/components/marquee-wizard/marquee-wizard-step.component.ts
```diff
@@ -19,30 +18,6 @@ export class MarqueeWizardStepComponent extends WizardStepComponent {
     /** Detect if an icon has been defined using the directive */
     @ContentChild(MarqueeWizardStepIconDirective, { read: TemplateRef, static: false }) _iconTemplate: TemplateRef<void>;
 
-    get valid(): boolean {
```
darrenmcc-microfocus, line 22: WizardStepComponent now handles this as the service is now used by both.

src/components/marquee-wizard/marquee-wizard.component.html
```diff
@@ -37,8 +37,6 @@
 
             <li *ngFor="let step of steps; let index = index"
                 role="tab"
-                uxTabbableListItem
```
darrenmcc-microfocus, line 40: Reorganised these properties to match the `WizardComponent` for consistency and readability.

src/components/marquee-wizard/marquee-wizard.component.html
```diff
@@ -47,7 +45,13 @@
                 [attr.aria-setsize]="steps.length"
                 [attr.aria-selected]="step.active"
                 [attr.aria-controls]="id + '-step-' + index"
+                [attr.aria-labelledby]="id + '-step-' + index + '-label'"
```
darrenmcc-microfocus, line 48: Some of the properties were being created in the component and most in the template. As there is no real requirement to create the likes of IDs in the component I removed the injected dependencies and added them here.

src/components/wizard/wizard.component.ts
```diff
@@ -347,6 +349,27 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
         return this.steps.toArray()[index];
     }
 
+    /**
+     * If a step in the wizard becomes invalid, all steps sequentially after
+     * it, should become unvisited and incomplete
+     */
+    protected setNextStepsUnvisited(currentStep: WizardStep): void {
```
darrenmcc-microfocus, line 356: `WizardStep` interface enabled both Wizard and MarqueeWizard steps to be handled in the `WizardStepComponent` class.

src/components/marquee-wizard/marquee-wizard.component.ts
```diff
@@ -45,32 +43,24 @@ export class MarqueeWizardComponent extends WizardComponent implements OnDestroy
...
 
-    /** Unsubscribe from all subscriptions when component is destroyed */
-    private _onDestroyed = new Subject<void>();
-
     get isTemplate(): boolean {
         return this.description && this.description instanceof TemplateRef;
     }
 
-    constructor(marqueeWizardService: MarqueeWizardService,
+    constructor(protected wizardService: WizardService,
```
ashley-hunter, line 50: I don't think this needs to be `protected` here, as we are passing it to the base class which has it as a protected member. 
Just this should do:

```typescript
constructor(wizardService: WizardService,
```

src/components/wizard/wizard.component.ts
ashley-hunter: Hi Darren, I may have come up with a better way of doing this than requiring the WizardStep interface. It should be a fairly easy change. We can make the service use a generic instead eg:

```typescript
@Injectable()
export class WizardService<TWizardStep> {
    valid$ = new Subject<WizardValidEvent<TWizardStep>>();
}

export interface WizardValidEvent<TWizardStep> {
    step: TWizardStep;
    valid: boolean;
}
```

Then we can just define the type in the constructor, eg:

```typescript
constructor(wizardService: WizardService<MarqueeWizardStepComponent>,
        private _resizeService: ResizeService,
        private _elementRef: ElementRef<HTMLElement>)
```

This should keep the circular dependency out but mean we don't have to keep an additional interface maintained anytime either step component changes

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -276,105 +279,148 @@ describe('Marquee wizard with delayed step creation', () => {
...
+interface WizardStep {
+    header: string;
+    content: string;
+    valid: boolean;
+    visited?: boolean;
+    completed?: boolean;
+}
+
 @Component({
-    selector: 'marquee-wizard-validation-app',
```
donna-marijne, line 280: Was it intentional to remove this test fixture?

src/components/marquee-wizard/marquee-wizard.component.ts
```diff
@@ -102,37 +92,14 @@ export class MarqueeWizardComponent extends WizardComponent implements OnDestroy
...
-        const current = steps.findIndex(step => step === state.step);
-        const affected = steps.slice(current);
-
-        affected.forEach(step => {
-
-            // the step should no longer be completed
-            step.completed = false;
-
-            // if the step is not the current step then also mark it as unvisited
-            if (this.resetVisitedOnValidationError && step !== state.step) {
```
donna-marijne, line 132: Looks like you've forced your changes on top of the more recent changes instead of doing a merge. Please revert and re-do this.

darrenmcc-microfocus: Apologies I should have seen that in the changes.

src/components/wizard/wizard.component.ts
```diff
@@ -150,13 +150,13 @@ export class WizardComponent implements OnInit, OnDestroy {
     private _step: number = 0;
     protected _onDestroy = new Subject<void>();
 
-    constructor(protected _wizardService: WizardService) {
+    constructor(protected _wizardService: WizardService<WizardStepComponent | MarqueeWizardStepComponent>) {
```
ashley-hunter, line 153: Remove `MarqueeWizardStepComponent` from this class. This class is the base class and should not know about any other uses of it. Otherwise this list could have to grow anytime we added a different wizard implementation. There should be no references to anything marquee wizard related in this class.

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -276,105 +279,148 @@ describe('Marquee wizard with delayed step creation', () => {
...
+    content: string;
+    valid: boolean;
+    visited?: boolean;
+    completed?: boolean;
+}
+
 @Component({
     selector: 'marquee-wizard-validation-app',
     template: `
-        <ux-marquee-wizard
```
donna-marijne, line 282: Can you restore this deleted test please?

src/components/wizard/wizard.component.ts
```diff
@@ -146,19 +147,24 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
 
     private _step: number = 0;
-    private _onDestroy = new Subject<void>();
-
-    ngAfterViewInit(): void {
+    protected _onDestroy = new Subject<void>();
+
+    constructor(protected _wizardService: WizardService<WizardStepComponent>) {
+        // watch for changes to valid subject
+        this._wizardService.valid$.pipe(
```
ashley-hunter, line 154: We need to unsubscribe from this on component destroy. We can also remove the `{ }` from the subscribe function as there is only one statement.

src/components/wizard/wizard.component.ts
```diff
@@ -152,6 +152,7 @@ export class WizardComponent implements OnInit, OnDestroy {
     constructor(protected _wizardService: WizardService<WizardStepComponent>) {
         // watch for changes to valid subject
         this._wizardService.valid$.pipe(
+            takeUntil(this._onDestroy),
```
ashley-hunter, line 155: `takeUntil` should always be the last operator (unless there is a `shareReplay` operator which should come after)

src/components/wizard/wizard.component.ts
```diff
@@ -347,12 +351,37 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
+        this.getFutureSteps().forEach(step => {
+            // if the step is not the current step then also mark it as unvisited
+            if (step.visited && step !== this.getCurrentStep()) {
+                step.setVisited(false);
+            }
+        });
+    }
+
+    /**
+     * Get the steps ahead of the currently active steps
```
donna-marijne, line 373: This is an incorrect description and function name, since it includes the currently active step.

src/components/wizard/wizard.component.spec.ts
```diff
@@ -35,11 +33,20 @@ export class WizardAsyncValidationTestComponent {
     firstStepValidator: () => boolean | Promise<boolean>;
     lastStepValidator: () => boolean | Promise<boolean>;
 
-    stepChanging(_: StepChangingEvent) {}
-    stepChange(_: number) {}
-    onNext(_: number) {}
-    onFinishing() {}
-    onFinish() {}
+    stepChanging(_: StepChangingEvent) {
+    }
```
donna-marijne, line 37: I really dislike this auto formatting. Could you reconfigure your formatter to ignore lines like this, and undo these unnecessary changes?

docs/app/pages/components/components-sections/wizard/wizard/wizard.component.html
```diff
@@ -113,6 +113,9 @@
     <tr uxd-api-property name="disableNextWhenInvalid" type="boolean">
         If set to <code>true</code> the 'Next' or 'Finish' button will become disabled when the current step is invalid.
     </tr>
+    <tr uxd-api-property name="resetVisitedOnValidationError" type="boolean" defaultValue="true">
```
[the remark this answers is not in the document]
ashley-hunter: Yes, perhaps this should be false by default to retain existing functionality?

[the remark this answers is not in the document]
darrenmcc-microfocus: @donna-marijne I'll make that change if you are happy that is right solution.

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -328,6 +340,7 @@ describe('Marquee wizard with validation', () => {
         component.step1Visited = true;
         component.step2Visited = true;
         fixture.detectChanges();
+        await fixture.whenStable();
```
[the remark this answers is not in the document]
darrenmcc-microfocus: Sorry I added this, removed now.

src/components/wizard/wizard.component.spec.ts
ashley-hunter: I think this was probably bad configuration in my editor 🤦‍♂️ sorry Darren

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -376,5 +390,146 @@ describe('Marquee wizard with validation', () => {
...
         return !stepElements[index].classList.contains('invalid');
     }
+});
+
+
+/**
+ * Test visitedChange events firing correctly
+ **/
+@Component({
+    selector: 'marquee-wizard-visited-change-app',
```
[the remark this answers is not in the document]
darrenmcc-microfocus: I probably should have refactored this as I already wrote these tests when you were working on that urgent fix at the same time. I'll condense them into one.

src/components/wizard/wizard.component.ts
darrenmcc-microfocus: Apologies this should have been updated to match the functionality.

src/components/wizard/wizard.component.spec.ts
darrenmcc-microfocus: No problem @ashley-hunter, I'll back that out in my next commit.

docs/app/pages/components/components-sections/wizard/wizard/wizard.component.html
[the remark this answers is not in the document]
donna-marijne: Yes I think so, as long as the default value for the marquee wizard remains true.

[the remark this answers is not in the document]
darrenmcc-microfocus: @donna-marijne In terms of the docs do I need to split them out to show both values, or just remove the defaultValue?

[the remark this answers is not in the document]
donna-marijne: In the Wizard section defaultValue should be false, and in the Marquee Wizard section defaultValue should be true.

src/components/wizard/wizard-step.component.ts
```diff
@@ -64,18 +69,33 @@ export class WizardStepComponent {
...
 
     constructor(
-        private readonly _changeDetector: ChangeDetectorRef,
-        private readonly _elementRef: ElementRef,
-        private readonly _renderer: Renderer2) { }
+        private readonly _wizardService: WizardService<WizardStepComponent>,
+        private readonly _changeDetector: ChangeDetectorRef) {
+    }
+
+    setVisited(value: boolean): void {
```
donna-marijne, line 84: To make this super clear and prevent future misunderstandings which mess up our outputs, I'd like this and any future methods which do the same thing to be named like:

setVisitedAndEmitChangeEvent

That way someone won't accidentally use this in a setter for `visited`, for example.

src/components/wizard/wizard.service.ts
```diff
@@ -1,19 +1,18 @@
...
  * between the marquee wizard steps and the containing marquee wizard.
- * We cannot inject the Host due to the steps being content children 
+ * We cannot inject the Host due to the steps being content children
  * rather than view children.
  */
 @Injectable()
-export class MarqueeWizardService {
-    valid$ = new Subject<MarqueeWizardValidEvent>();
+export class WizardService<TWizardStep> {
+    valid$ = new Subject<WizardValidEvent<TWizardStep>>();
```
donna-marijne, line 12: Since you are making changes here, this should be an EventEmitter and be named something like validChange because it is an event and not a representation of state.

src/components/wizard/wizard.component.spec.ts
```diff
@@ -249,3 +246,126 @@ describe('Wizard with validator', () => {
...
+
+interface WizardStep {
+    header: string;
+    content: string;
+    visited?: boolean;
+    valid?: boolean;
+}
+
+@Component({
+    selector: 'wizard-visited-change-test-app',
```
donna-marijne, line 258: Synchronise this with the marquee wizard tests above. Since you have introduced `resetVisitedOnValidationError` into the standard wizard, those tests need to be done here as well.

src/components/wizard/wizard.component.ts
```diff
@@ -146,33 +152,31 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
-            this.setWizardStepIds();
-            this.update();
-        });
+        this.steps.changes.pipe(tick(), takeUntil(this._onDestroy)).subscribe(() => this.update());
+
+        // watch for changes to valid subject
+        this._wizardService.valid$.pipe(
+            filter((event: WizardValidEvent<WizardStepComponent>) => !event.valid),
+            takeUntil(this._onDestroy)
+        ).subscribe(() => this.setNextStepsUnvisited());
```
donna-marijne, line 171: This should be setting steps unvisited from the invalid step onwards, not the active step. The `valid` property can be changed on any step via the binding.

src/components/wizard/wizard.component.ts
```diff
@@ -347,12 +351,37 @@ export class WizardComponent implements AfterViewInit, OnDestroy {
...
+                step.setVisited(false);
+            }
+        });
+    }
+
+    /**
+     * Get the currently active step and all steps beyond it
+     */
+    protected getFutureSteps(): WizardStepComponent[] {
+        return this.steps.toArray().slice(this.step);
```
donna-marijne, line 376: It would probably be better to change this to actually return the future steps, then the comment is not needed.

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -276,46 +282,89 @@ describe('Marquee wizard with delayed step creation', () => {
...
+                Step Three Content
+            </ux-marquee-wizard-step>
+            <ux-marquee-wizard-step
+                header="Step Four"
+                [valid]="step4Valid"
+                [(visited)]="step4Visited"
+                [(completed)]="step4Completed"
+                (visitedChange)="visitedChanged(3, $event)">
+                Step Four Content
+            </ux-marquee-wizard-step>
```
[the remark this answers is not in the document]
donna-marijne: Having seen the issue with `setNextStepsUnvisited`, there is in fact value in having three steps and modifying the validity on step 2. So let's go with that instead.

src/components/wizard/wizard.service.ts
```diff
@@ -1,19 +1,18 @@
...
  * between the marquee wizard steps and the containing marquee wizard.
- * We cannot inject the Host due to the steps being content children 
+ * We cannot inject the Host due to the steps being content children
  * rather than view children.
  */
 @Injectable()
-export class MarqueeWizardService {
-    valid$ = new Subject<MarqueeWizardValidEvent>();
+export class WizardService<TWizardStep> {
+    validChange$ = new Subject<WizardValidEvent<TWizardStep>>();
```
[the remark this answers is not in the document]
ashley-hunter: I don't think this should be an `EventEmitter`? An event emitter would only be for an `@Output` whereas this is in a service?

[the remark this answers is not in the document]
donna-marijne: We've used EventEmitters in services in the past.

https://github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/blob/master/src/components/side-menu/side-menu/side-menu.service.ts#L21

I don't think Subject is an appropriate type for an event.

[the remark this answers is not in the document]
donna-marijne: If you don't want it to be an event then it should probably be a complete description of the validation state of the wizard. Like a collection of observables representing the state of each step.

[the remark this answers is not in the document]
ashley-hunter: It probably shouldn't have been used in that service. An `EventEmitter` is just a `Subject`:

```typescript
export interface EventEmitter<T> extends Subject<T> {
```

Comment from Angular Team:

> EventEmitter is really an Angular abstraction, and should be used pretty much only for emitting custom Events in components. Otherwise, just use Rx as if it was any other library.

See this on the proper use of EventEmitter: https://stackoverflow.com/questions/36076700/what-is-the-proper-use-of-an-eventemitter#answer-36076701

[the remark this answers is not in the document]
donna-marijne: Fair enough, so we should rework this to be stateful then?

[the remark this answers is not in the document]
ashley-hunter: If we need to know the valid state of every step at any time then yes, we can update this to use a BehaviorSubject with the state of every step. If we just need to be notified when the valid state of any step changes then a Subject will suffice.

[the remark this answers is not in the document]
donna-marijne: Honestly yes to to former because essentially we are converting state in a component to an event in a service and back to state in another component. But that's beyond the scope of this issue. I thought using EventEmitter would be a quick fix to clarify a confusing system but since it's not recommended to do so, please disregard this feedback. Thanks.

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -1,13 +1,12 @@
...
-    ComponentFixture,
-    fakeAsync,
-    TestBed,
-    tick
-} from '@angular/core/testing';
+import { ChangeDetectorRef, Component, OnDestroy } from '@angular/core';
+import { async, ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { MarqueeWizardModule } from './marquee-wizard.module';
 
+const NEXT_BUTTON_SELECTOR: string = '.button-primary';
```
donna-marijne, line 5: I believe this is unused now so can be removed.

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -276,95 +273,146 @@ describe('Marquee wizard with delayed step creation', () => {
...
+        expect(isStepValid(1)).toBe(false);
+        expect(isStepValid(2)).toBe(true);
 
         expect(isStepVisited(0)).toBe(true);
-        expect(isStepVisited(1)).toBe(false);
+        expect(isStepVisited(1)).toBe(true);
+        expect(isStepVisited(2)).toBe(false);
+
+        // check visitedChange was emitted 1 time only
+        expect(visitedChanged.calls.all().length).toBe(1);
```
donna-marijne, line 383: Can you assert the values of `setXVisited` in this test please?

src/components/marquee-wizard/marquee-wizard.component.spec.ts
```diff
@@ -276,95 +273,146 @@ describe('Marquee wizard with delayed step creation', () => {
...
 
         expect(isStepValid(0)).toBe(true);
         expect(isStepValid(1)).toBe(true);
+        expect(isStepValid(2)).toBe(true);
 
         expect(isStepVisited(0)).toBe(true);
         expect(isStepVisited(1)).toBe(true);
+        expect(isStepVisited(2)).toBe(true);
+
+        visitedChanged.calls.reset();
```
donna-marijne, line 401: Can you remove this please?

src/components/wizard/wizard.component.spec.ts
```diff
@@ -1,14 +1,10 @@
...
-    ComponentFixture,
-    fakeAsync,
-    TestBed,
-    tick
-} from '@angular/core/testing';
+import { async, ComponentFixture, fakeAsync, TestBed, tick } from '@angular/core/testing';
 import { StepChangingEvent } from './wizard.component';
 import { WizardModule } from './wizard.module';
 
+const NEXT_BUTTON_SELECTOR = '.wizard-footer .button-primary';
```
donna-marijne, line 6: This can be removed.

src/components/marquee-wizard/marquee-wizard.component.spec.ts
darrenmcc-microfocus: @donna-marijne Is this in every test or just this one?

donna-marijne: You can certainly add it to the other test as well if you think it will improve coverage.

PR #1369, opened 2020-07-02
EL-3953 - Compute notification spacing from margins

donna-marijne:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3953

* Remove default value for `spacing`.
* Calculate the outer height of a notification based on height + margins when `spacing` is not defined.

Making this change because increased spacing between notifications was requested due to the larger box shadow in the 2020 theme. This allows us to control the spacing via margins if the consumer has not provided an explicit `spacing` value.

* Add e2e tests for the notification list.
* Fix an issue where the notifications list fills the height of the screen after changing position from bottom to top.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3953-notifications-2020

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3953-notifications-2020~CI/)


src/components/notification/notification-list.component.ts
```diff
@@ -99,8 +99,38 @@ export class NotificationListComponent implements AfterViewInit, OnDestroy {
...
-            top = top + element.scrollHeight + notification.spacing;
+            top = top + this.getNotificationHeightPixels(notification, elements[i]);
+        }
+    }
+
+    private updateListPosition(elements: ElementRef[], notifications: NotificationRef[]): void {
+        if (this.position === 'bottom-left' || this.position === 'bottom-right') {
+
+            // calculate the total height of all notifications including spacing
+            this._bottom = notifications.reduce((total, notification, index) =>
```
DeanHaire, line 110: I think `totalHeight` would be a better name here. Then you could also remove the comment as it would then be clearer what the value represents.

src/components/notification/notification-list.component.ts
```diff
@@ -99,8 +99,38 @@ export class NotificationListComponent implements AfterViewInit, OnDestroy {
...
+                total + this.getNotificationHeightPixels(notification, elements[index]), 0);
+        } else {
+
+            // In a top position, bottom should be unset.
+            this._bottom = undefined;
+        }
+    }
+
+    /** Get the height of the notification, including spacing if configured. */
+    private getNotificationHeightPixels(notification: NotificationRef, elementRef: ElementRef): number {
```
DeanHaire, line 120:
```suggestion
    private getNotificationHeightInPixels(notification: NotificationRef, elementRef: ElementRef): number {
```
I think including 'in' within the name makes this more readable. Same with `getElementOuterHeightPixels` below.

src/components/notification/notification-list.component.ts
donna-marijne: If you mean the `_bottom` property, that's used to derive the CSS `bottom` value (see above) so renaming it would be contradictory. The comment probably causes the confusion here so I have removed it now.

DeanHaire:
```suggestion
            this._bottom = notifications.reduce((totalHeight, notification, index) =>
```
No, I meant the `total` within `reduce` could possibly be changed to `totalHeight` to give a future reader of this line a quicker understanding of its purpose

donna-marijne: Oh, yes, that's reasonable. Thanks.

PR #1373, opened 2020-07-10
Logical Expression Builder

eriklittwin:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-4091

Create the Logical Expression Builder component, including Input components for the following types:
* text
* number
* date
* date range
* enum

Additional changes have been made in ux-aspects-micro-focus and quantum-ux-aspects.

#### Related Pull Requests

* Pull request for Micro Focus Style: https://github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/pull/653
* Pull request for Micro Focus Style Variables: https://github.houston.softwaregrp.net/caf/quantum-ux-aspects/pull/153


src/components/logical-expression-builder/leb-condition/leb-condition.component.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    public handleFieldSelected(selectedField: FieldDefinition): void {
+        // get operators for new field type
+        if (selectedField) {
+            this._field = selectedField;
+            this.operators = this._lebService.getOperatorsByFieldType(this._field.fieldType);
+            this._operator = null;
+            this._value = null;
+
+            this.buildCondition();
```
tobias-schreiner, line 118: Why does the condition need to be set?

src/components/logical-expression-builder/leb-condition/leb-condition.component.ts
```diff
@@ -0,0 +1,173 @@
...
+            value: this._value ?? null
+        };
+    }
+
+    public confirmCondition(): void {
+        this._lebService.setEditBlocked(false);
+
+        this.editable = false;
+        this.buildCondition();
+        this._condition = { ...this._condition, editable: false };
```
tobias-schreiner, line 147: You could pass this as an optional parameter to buildCondition()

src/components/logical-expression-builder/leb-group/leb-group.component.html
```diff
@@ -0,0 +1,64 @@
...
+
+<ng-container *ngFor="let child of subExpression?.children; index as index">
+    <ux-leb-condition
+        *ngIf="child.type === 'condition'"
+        (conditionChange)="handleGroupChange($event, index)"
+        [condition]="child"
+        [id]="index"
+        (conditionDeleted)="removeConditionAtIndex($event)"
+        (conditionEmbedded)="embedConditionAtIndex($event)"
+    [indent]="indent + 40">
```
tobias-schreiner, line 32: Make 40 a constant

src/components/logical-expression-builder/leb-group/leb-group.component.html
```diff
@@ -0,0 +1,64 @@
...
+            [uxMenuTriggerFor]="addMenu"
+            type="button"
+            class="btn btn-icon button-primary"
+            aria-label="Add">
+            <span class="ux-icon ux-icon-add" aria-hidden="true"></span>
+        </button>
+    </td>
+</tr>
+
+<ux-menu #addMenu>
```
tobias-schreiner, line 56: The AngularJS component shows both field names and logical operators immediately (can be part of next pull request)

src/components/logical-expression-builder/leb-group/leb-group.component.spec.ts
```diff
@@ -0,0 +1,25 @@
+/*
```
tobias-schreiner, line 1: Use xdescribe instead of comments - like this, the test is still visible in the test list and shows as disabled

src/components/logical-expression-builder/leb-group/leb-group.component.ts
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public addCondition() {
+        this.subExpression.children = [...this.subExpression.children, {
+            type: 'condition',
+            field: null,
+            operator: null,
+            value: null,
+            editable: true,
+        }];
```
tobias-schreiner, line 51: is there no groupChange event needed?

src/components/logical-expression-builder/leb-group/leb-group.component.ts
```diff
@@ -0,0 +1,91 @@
...
+    public handleSelectedOperatorChange(selectedOperator: LogicalOperatorDefinition) {
+        this.selectedLogicalOperator = selectedOperator;
+        this.subExpression = { ...this.subExpression, logicalOperator: this.selectedLogicalOperator.name };
+        this.groupChange.emit(this.subExpression);
+    }
+
+    public handleGroupChange(event: ExpressionGroup | ExpressionCondition, index: number) {
+        this.subExpression.children[index] = event;
+
+        // remove children that have been deleted
```
tobias-schreiner, line 38: Only one child can be deleted at a time - remove it from the array instead

src/components/logical-expression-builder/logical-expression-builder.component.ts
```diff
@@ -0,0 +1,149 @@
...
+
+    constructor(private _lebService: LogicalExpressionBuilderService) {
+    }
+
+    public isExpressionEmpty(): boolean {
+        return !this.expression?.type;
+    }
+
+    public getLogicalOperatorName(): string {
+        return ('logicalOperator' in this.expression) ? this.expression.logicalOperator : null;
```
tobias-schreiner, line 63:
```suggestion
        return this.expression.logicalOperator ?? null;
```

src/components/logical-expression-builder/logical-expression-builder.component.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    public handleGroupChange(expression: LogicalExpressionBuilderExpression) {
+        let temp = { ...expression };
+
+        // make expression just a condition if it contains exactly one group with exactly one condition in it
+        if (temp?.type === 'group'
+            && ('children' in temp)
```
tobias-schreiner, line 71:
```suggestion
            && (temp.hasOwnProperty('children'))
```
However, this is already covered in the next line.

src/components/logical-expression-builder/logical-expression-builder.component.ts
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        if (expression.type === 'group' && ('children' in expression)) {
+            const children = expression.children.map((child) => this.addEditableFieldToConditionsInExpression(child));
+
+            return { ...expression, children };
+        }
+    }
+
+    private cleanExpression(expression: LogicalExpressionBuilderExpression): LogicalExpressionBuilderExpression {
```
tobias-schreiner, line 128: Prepend with _

src/components/logical-expression-builder/logical-expression-builder.component.ts
```diff
@@ -0,0 +1,149 @@
...
+    public addGroup() {
+        // adds a group to the condition if there is only one condition to the expression and a second one is added
+        const firstCondition = { ...this.expression };
+
+        this.expression = {
+            type: 'group',
+            logicalOperator: this._lebService.getLogicalOperators()[0].name,
+            children: [
+                firstCondition,
+                { type: 'condition', field: null, operator: null, value: null, editable: true },
```
tobias-schreiner, line 101: Make this a constant or function

src/components/logical-expression-builder/logical-expression-builder.component.ts
```diff
@@ -0,0 +1,149 @@
...
+            ]
+        };
+
+        this._lebService.setEditBlocked(true);
+        this.expressionChange.emit(this.cleanExpression(this.expression));
+    }
+
+    /* Helper methods for adding and removing the 'editable' property to and from conditions */
+
+    private addEditableFieldToConditionsInExpression(expression: LogicalExpressionBuilderExpression): LogicalExpressionBuilderExpression {
```
tobias-schreiner, line 111: Can this be solved differently, e.g. interpreting undefined as false?

src/components/logical-expression-builder/logical-expression-builder.module.ts
```diff
@@ -0,0 +1,61 @@
...
+    DateTimePickerModule,
+    PopoverModule,
+    AccessibilityModule,
+    DateRangePickerModule,
+    FormsModule,
+    HoverActionModule,
+    MenuModule,
+    IconModule
+  ],
+    exports: [...EXPORTS, L10nPipe],
```
tobias-schreiner, line 56: Why is the L10nPipe exported?

src/components/logical-expression-builder/pipes/display-value.pipe.ts
```diff
@@ -0,0 +1,62 @@
...
+export class DisplayValuePipe implements PipeTransform {
+    private readonly _transformFunction: DisplayValueFunction;
+
+    constructor(private _lebService: LogicalExpressionBuilderService, private _datePipe: DatePipe) {
+        this._transformFunction = this._lebService.getDisplayValueFunction();
+    }
+
+    transform(value: any, fieldType?: string, fieldData?: any): string {
+        if (this._transformFunction) {
+            return this._transformFunction(value, fieldType, fieldData);
```
tobias-schreiner, line 16: You could allow to return undefined from the transform function, so that the standard display value is used instead

src/components/logical-expression-builder/pipes/l10n.pipe.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+    it('should return localized string if it exists', () => {
+        expect(pipe.transform('testValue')).toEqual('Test Value');
+    });
+
+    it('should return undefined if localized string does not exists', () => {
+        expect(pipe.transform('xy')).toBeUndefined();
+    });
+
+    it('should return undefined if localized string is empty', () => {
+        expect(pipe.transform('testEmptyString')).toBeUndefined();
```
tobias-schreiner, line 35: This should result in '', not undefined

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.html
```diff
@@ -0,0 +1,16 @@
...
+    (expressionChange)="queryChanged($event)"
+    [fields]="fields"
+    [operators]="operators"
+    [logicalOperators]="logicalOperators"
+    [localizedStrings]="localizedStrings">
+</ux-logical-expression-builder>
+
+<h4>Search Query</h4>
+
+<pre>{{ preview | json }}</pre>
```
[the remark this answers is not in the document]
eriklittwin: added max-height with scrollbar

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.ts
```diff
@@ -0,0 +1,133 @@
...
+            { name: 'less than', label: 'less than', component: NumberInputComponent },
+            { name: 'greater than', label: 'greater than', component: NumberInputComponent },
+        ],
+    };
+
+    @ViewChild('iconTemplate') iconTemplate: TemplateRef<any>;
+
+    fields: FieldDefinition[] = [
+        { name: 'name', label: 'Name', fieldType: 'text' },
+        { name: 'date', label: 'Date', fieldType: 'date', data: {} },
```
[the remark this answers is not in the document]
eriklittwin: it doesn't, I removed those

src/components/logical-expression-builder/input-components/date-input/date-input.component.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    showTime: boolean;
+    showTimezones: boolean;
+    showMeridians: boolean;
+    showSpinners: boolean;
+
+    inputSubscription: Subscription;
+
+    ngAfterViewInit(): void {
+        this.inputSubscription = fromEvent(this.dateInput.nativeElement, 'input')
```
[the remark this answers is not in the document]
eriklittwin: Here I used the example from the UXA DatePicker docs. And if I use `(input)="parse($event)"` I don't get the debounce on the event. The date would change on every letter the user types.

src/components/logical-expression-builder/leb-condition/leb-condition.component.html
```diff
@@ -0,0 +1,81 @@
+<tr *ngIf="editable; else uneditable" class="leb-condition-row" [style.padding-left.px]="indent">
```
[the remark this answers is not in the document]
eriklittwin: I wanted to make sure the css-selectors are specific to this component, so there won't be any interference with other/future selectors. Would you drop the prefix entirely or use a different one?

src/components/logical-expression-builder/leb-condition/leb-condition.component.ts
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    public handleOperatorSelected(selectedOperator: OperatorDefinition): void {
+        this._operator = selectedOperator;
+        this.createInputComponent();
+
+        if (selectedOperator) {
+            this.buildCondition();
+        }
+    }
```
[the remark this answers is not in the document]
eriklittwin: It is cleared inside `_createInputComponent()`. I have to clear it manually in `handleFieldSelected()` if the condition is already complete and the user changes the field type.

src/components/logical-expression-builder/interfaces/DisplayValueFunction.ts
```diff
@@ -0,0 +1 @@
+export type DisplayValueFunction = (value: any, fieldType?: string, fieldData?: any) => string;
```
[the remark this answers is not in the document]
eriklittwin: Users can provide this function to override the format function in the DisplayValuePipe to customize how the entered value is displayed in static mode.

src/components/logical-expression-builder/interfaces/LogicalExpressionBuilderExpression.ts
```diff
@@ -0,0 +1,15 @@
+export interface ExpressionCondition {
+    type: string;
+    field: string;
+    operator: string;
+    value: any;
+    editable?: boolean;
```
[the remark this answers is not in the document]
eriklittwin: Basically for type checking in the other components

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.html
```diff
@@ -0,0 +1,16 @@
...
+
+<ng-template #iconTemplate>
```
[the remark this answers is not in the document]
eriklittwin: Removed for now, as I couldn't pass it in the fieldDefinition

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.html
```diff
@@ -0,0 +1,16 @@
...
+    <ux-icon name="add"></ux-icon>
+</ng-template>
```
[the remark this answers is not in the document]
eriklittwin: removed for now

src/components/logical-expression-builder/input-components/date-input/date-input.component.html
```diff
@@ -0,0 +1,27 @@
+<input
```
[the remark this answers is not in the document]
eriklittwin: will check

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.ts
```diff
@@ -0,0 +1,133 @@
...
+        }
+    }*/
+
+    preview: LogicalExpressionBuilderExpression;
+
+    constructor(public colorService: ColorService) {
+        super(require.context('./snippets/', false, /\.(html|css|js|ts)$/));
+    }
+
+    queryChanged(query: LogicalExpressionBuilderExpression): void {
```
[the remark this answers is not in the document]
eriklittwin: App does not load when I do that, need to investigate

tobias-schreiner:
Hallo Erik,

Angular sorgt dafür, dass Deine CSS-Klassen nur in Deiner Komponente sichtbar sind, Du kannst also den Präfix getrost weglassen.

Gruß

                Tobias

tobias-schreiner:
Hallo Erik,

Das ist mir klar. Aber warum ist das so wichtig, dass Du es implementiert hast?

Gruß

                Tobias

src/components/logical-expression-builder/interfaces/DisplayValueFunction.ts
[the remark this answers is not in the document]
eriklittwin: Because in our last meeting with UX Aspects and Timo, we talked about how to customize what is displayed. And if I remember correctly we said for the first iteration it is okay to just display as a text. Also, the implementation was pretty straight-forward and not very much effort.

src/components/logical-expression-builder/input-components/text-input/text-input.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="form-group m-b-nil" [ngClass]="{ 'has-error': !_valid }">
+    <input
+        type="text"
+        class="form-control"
+        [value]="value"
```
ashley-hunter, line 5: Can we not use `ngModel` here? Instead of manually updating the value ourselves?

src/components/logical-expression-builder/input-components/text-input/text-input.component.ts
```diff
@@ -0,0 +1,32 @@
+import { Component, EventEmitter, Input, Output } from '@angular/core';
+
+@Component({
+    selector: 'ux-text-input',
+    templateUrl: './text-input.component.html'
+})
+export class TextInputComponent {
+    @Output() valueChange: EventEmitter<string> = new EventEmitter<string>();
+    @Output() valid: EventEmitter<boolean> = new EventEmitter<boolean>();
```
ashley-hunter, line 9: Perhaps `validChange` would be a better name for the output. `valid` is a state, whereas an output is an event, so the name should indicate that. This would also allow us to add a `valid` input in future if that was ever a requirement

ashley-hunter (changes requested):
Thanks for the contribution. I have just taken a quick look so far and made a few suggestions. In general I notice a few common things across many files:

1. Remove the use of `public` on properties and methods as this does not align with the rest of our code base
2. Avoid the use of `any` as this completely turns off any type checking around that property. In almost all cases you can use a generic instead for any unknown type which can then be type checked by TypeScript once the type information is available.
3. Check some imports, in particular around NgModules and Components. `ng-packagr` has some issues if you try to assume the presence of an index.ts file, for example:

```typescript
import { SelectModule } from '../../../select';
import { IconModule } from '../../../icon';
```
Should explicitly list the `index` file as part of the import:
```typescript
import { SelectModule } from '../../../select/index';
import { IconModule } from '../../../icon/index';
```
4. Private methods should not be prefixed with an underscore, e.g:

```typescript
private _setTabIndex(): void {
```
The underscore prefix on methods is only required on `public` functions that should be private but can't be because they are used within an Angular template. The underscore is essentially a warning to any consuming application that this method should not be used. If the function is marked as private then typescript will enforce this for us.

5. `@Input` setters should not emit `@Outputs`. I notice in a few places there are Input setters that emit, e.g.:

```typescript
@Input()
    set value(value: string) {
        this._value = value ?? '';
        this.valueChange.emit(this.value);
        this._valid = this._validate(this.value);
        this.valid.emit(this._valid);
    }
```

In this case the user is telling us that the value has changed, so we do not need to emit to tell them that the value has changed. An output should only emit whenever the component is responsible for making the change internally and needs to inform the parent component.

The second part of this setter emits the `valid` state, now this is partially acceptable as it it is not the `valid` input that has changed. However, an output should only emit if the value has changed, so in this case we should be checking the previous valid state with the new valid state and only emit if there has been a change.

eriklittwin:
Thank you for your feedback! I will work on the requested changes and notify you once the changes are made.

src/components/logical-expression-builder/input-components/date-range-input/date-range-input.component.ts
```diff
@@ -0,0 +1,180 @@
...
+    Output,
+} from '@angular/core';
+import { DateTimePickerTimezone, timezones } from '../../../date-time-picker';
+import { formatDate } from '@angular/common';
+
+@Component({
+    selector: 'ux-date-range-input',
+    templateUrl: './date-range-input.component.html'
+})
+export class DateRangeInputComponent {
```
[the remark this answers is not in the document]
donna-marijne: Yes, I think that should be the default naming strategy unless there is a reasonable scenario that they would be used outside of the new component.

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.html
```diff
@@ -0,0 +1,331 @@
...
+            </li>
+        </ul>
+    </tr>
+</uxd-api-properties>
+
+<h4>Custom Input Components</h4>
+
+<p>
+    In addition to the predefined Input Components, it is also possible to built custom input components. These must
+    conform with the following interfaces:
```
donna-marijne, line 281: Please specify the name of the interface that the custom component must implement.

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.html
```diff
@@ -0,0 +1,331 @@
...
+</p>
+
+<uxd-api-properties tableTitle="Outputs">
```
donna-marijne, line 284: Should be Inputs.

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.html
```diff
@@ -0,0 +1,331 @@
...
+    <tr uxd-api-property name="value" type="any">
+        Contains the value of the respective condition.
+    </tr>
+
+    <tr uxd-api-property name="data" type="any">
+        Contains the data for the specific <code>field</code> defined in the fields definition. E.g., it can be used
```
donna-marijne, line 290: Perhaps something like:

"The configuration options for the custom component, e.g. the available options in a select dropdown. This is defined in the `fields` input of `ux-logical-expression-builder`."

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.html
```diff
@@ -0,0 +1,331 @@
...
+    <tr uxd-api-property name="valueChange" type="any">
+        This event will fire whenever the value has been modified. It will contain the latest value.
+    </tr>
+    <tr uxd-api-property name="valid" type="boolean">
+        This event will fire whenever the validity of the value has changed. It will contain <code>true</code> if the
+        value is valid, and <code>false</code> if it is not.
+    </tr>
+</uxd-api-properties>
+
+<h4>Example for a custom Input Component</h4>
```
donna-marijne, line 306: Capitalization within headers should be consistent, i.e.

"Example of a Custom Input Component"

or

"Example of a custom input component"

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.html
```diff
@@ -0,0 +1,331 @@
...
+    <ux-tab heading="HTML">
+        <uxd-snippet [content]="snippets.compiled.textareaInputComponentHtml"></uxd-snippet>
+    </ux-tab>
+    <ux-tab heading="TypeScript">
+        <uxd-snippet [content]="snippets.compiled.textareaInputComponentTs"></uxd-snippet>
+    </ux-tab>
+</ux-tabset>
+
+<h4>Example Code</h4>
+
```
donna-marijne, line 323: Since there are two examples adjacent to one another, this should have a description clarifying that it is the code for the example at the top of the section.

docs/app/pages/components/components-sections/search/logical-expression-builder/snippets/fields-definition.ts
```diff
@@ -0,0 +1,17 @@
...
+    { name: 'date_of_birth', label: 'Date of birth', fieldType: 'date', data: { dateFormat: 'short' } },
+    {
+        name: 'gender',
+        label: 'Gender',
+        fieldType: 'enum',
+        data: {
+            options: [
+                { name: 'male', label: 'Male' },
+                { name: 'female', label: 'Female' },
+                { name: 'diverse', label: 'Diverse' }
```
donna-marijne, line 12: In English I think the correct term would be "non-binary".

donna-marijne (changes requested):
Feedback on the documentation below.

e2e/tests/components/logical-expression-builder/logical-expression-builder.e2e-spec.ts
```diff
@@ -0,0 +1,183 @@
...
+
+describe('Logical Expression Builder', () => {
+    let page: LogicalExpressionBuilderPage.Page;
+
+    beforeEach(async () => {
+        page = new LogicalExpressionBuilderPage.Page();
+        await page.getPage();
+    });
+
+    it('should have correct initial state', async () => {
```
donna-marijne, line 12: We're not doing this "correct initial state" test anymore. e2e test cases should be use cases of the component. The relevant initial state is asserted as a prerequisite of each test case.

e2e/tests/components/logical-expression-builder/logical-expression-builder.e2e-spec.ts
```diff
@@ -0,0 +1,183 @@
...
+        expect(await page.getExpressionObject()).toEqual(null);
+        expect(await page.getConditionRowCount()).toEqual(0);
+        expect(await page.getGroupRowCount()).toEqual(0);
+        expect(await page.getValid()).toBeTruthy();
+        expect(btn).toBeDefined();
+        expect(await btn.getText()).toEqual('ADD CONDITION');
+    });
+
+    it('should display one row when expression is just a condition', async () => {
+        await page.setOneCondition();
```
donna-marijne, line 24: For an e2e (end-to-end) test case, you would build the expression as a user would. A user would not click a button to pre-fill the component, they would click the "add condition" button, click the dropdown, select a field type, and so on. What you have implemented here are more suitable as karma tests, because user interaction is not considered.

Since this component is replacing an existing AngularJS component, I assume that the use cases are well understood. Those use cases should convert quite readily into e2e test cases. If you would like some advice on designing e2e test cases, we can discuss it.

ashley-hunter (changes requested):
Had some time to play around with this in a bit more details. Just a few additional observations:

1. The error message looks huge, perhaps something more compact? It alters the size of the row when it appears which seems wrong.

image

2. If I add a date field and choose "before" the date value is pre-populated, but I cannot add the field until it click inside the field that has the date (this happens with other fields too):

image

3. Whenever I perform any action on row, e.g. editing, the page scrolls. I'm guessing there is some focus behaviour causing this, but it seems jarring to scroll when I click on a row, if there were lots of conditions I could easily lose my place because the row moved: 

[Image from Gyazo](https://gyazo.com/983575ad11ef335eefd9516e40fb7487)

4. If i tab to a row, then tab to a child button, then shift+tab to navigate back to the row, the buttons disappear:

[Image from Gyazo](https://gyazo.com/ba469c8626ef06c2e73f2dfb5b364873)

5. This may already be supported, but do we allow the app to add tooltips to these buttons? The middle button was not immediately obvious what its purpose what, allowing a tooltip would allow an explanation to optionally be provided.
image

6. If I change the root condition to NOT, I get an error saying "'not' needs exactly one child." If I then choose reset expression the error message persists.

7. In the example code there are quite a lot of arrays being used, e.g.:
image
When it comes to sample code and also in the components we should always use the `ReadonlyArray` type. If we use normal mutable arrays (especially when using OnPush change detection) and someone mutates an array, Angular is not able to identify this as a change and the UI will not update leaving people who don't know about this quirk confused. We should always use `ReadonlyArray`s in our code and in the sample code to avoid this kind of issue.

e2e/tests/components/logical-expression-builder/logical-expression-builder.po.spec.ts
```diff
@@ -0,0 +1,150 @@
+import { $, $$, browser, ElementArrayFinder, ElementFinder } from 'protractor';
+import { Expression } from '../../../../src/components/logical-expression-builder';
```
donna-marijne, line 2: This should be imported from @ux-aspects/ux-aspects. I'm surprised this was not a build error.

e2e/tests/components/logical-expression-builder/logical-expression-builder.po.spec.ts
```diff
@@ -0,0 +1,150 @@
...
+
+export namespace LogicalExpressionBuilderPage {
```
donna-marijne, line 4: No need for a namespace here.

donna-marijne (changes requested):
Some feedback on the e2e tests.

e2e/tests/components/logical-expression-builder/logical-expression-builder.e2e-spec.ts
eriklittwin: Okay, that makes sense. I think discussing this is a good idea since honestly, I haven't implemented E2E tests before this.

docs/app/pages/components/components-sections/search/logical-expression-builder/logical-expression-builder.component.html
```diff
@@ -0,0 +1,331 @@
...
+                <code>validateFunction: (value: number) => boolean</code> - If specified, this function is used to
+                validate the value entered by the user.
+            </li>
+        </ul>
+    </tr>
+
+    <tr uxd-api-property name="DateRangeInputComponent">
+        <p>
+            Contains a date range picker. It expects an object of the form
+            <code>{{ '{ start: number, end: number }' }}</code>
```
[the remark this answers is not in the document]
eriklittwin: I'll change the interface. But I do get an error if I remove the interpolation brackets, which is why I put them there in the first place.

donna-marijne (changes requested):
Here is some feedback from Roland, our designer. We can set up a DM/call if you have any questions.

---

# Behaviour
1.	When you add a new condition like this…
 
image


you then see this…
 
image

the combination of the new row (with blue focus) and the fact that the row below will have a hover on it (unless you then move your cursor) is a bit confusing visually, and you then need another click to focus on the condition dropdown.

Suggestion
After selecting “Add condition” could we then apply automatically focus to the condition dropdown to avoid the extra click, and make it clearer for the user >
 
image

2.	Similarly after selecting a condition you see this…
 
image

Suggestion
For similar reasons, after selecting Author on the dropdown (for example) could we then apply automatically focus to the next dropdown to help step the user through each field of the condition without them requiring extra clicks >
 
image

3.	If you add a logical operator…
 
image


you then see this…
 
image

So the operator is added but you are now in a mode for adding a condition under a default AND operator (with the shouty red message). If you want to change the operator type you appear to have to do that after adding one or more conditions.

Suggestion
Should we land with just the operator row added and give the user a chance to change that if necessary before proceeding (which affects how many conditions are required)?
 
image


# Styling
My comments here are:
*	I find the legibility of the bold text for the field criteria isn’t great and having the text on the right of the fields in MF blue makes them look clickable.
*	The buttons are quite prominent and “noisy” – I would suggest using flat buttons instead for those on the right of the rows.
*	The operator warning messages are quite shouty considering that the user can’t avoid them – I suggest making these look more like the warning messages we can add to text fields such as on my example above.

I made some test edits in Inspector to try some alternatives out – see below:
1.	This uses Dark Grey #323435 for the left-hand text in the fields and Cerulean # 1668c1 for the right-hand text.
2.	The font weight is reduced from Bold to Semi-bold (600) for the field criteria
3.	For the + buttons – I would suggest using a consistent style with the + buttons on the Search Builder component with the blue circular buttons that are 50% opacity until hovered.
 
image

src/components/logical-expression-builder/pipes/display-value.pipe.ts
```diff
@@ -0,0 +1,66 @@
...
+ * components. It can be overridden with the displayValueFunction input property. */
+@Pipe({ name: 'displayValue' })
+export class DisplayValuePipe implements PipeTransform {
+    private readonly _transformFunction: DisplayValueFunction;
+
+    constructor(private _lebService: LogicalExpressionBuilderService, private _dateFormatterPipe: DateFormatterPipe) {
+        this._transformFunction = this._lebService.getDisplayValueFunction();
+    }
+
+    transform(value: any, fieldType?: string, fieldData?: any): string {
```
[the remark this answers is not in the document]
tobias-schreiner: I would rather return an empty string instead of undefined.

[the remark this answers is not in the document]
ashley-hunter: Thats fine if it returns an empty string, it currently returns a `string` or `undefined`, so the return type should be accurate to reflect the behaviour. If it can possibly return `undefined` then the return type should be `string | undefined`, if it returns an empty string as a fallback then `string` is fine for the return type.

tobias-schreiner:
TODO in this pull request:

- Feedback for this and the connected pull requests
- UX Review Feedback (https://microfocusinternational.sharepoint.com/:x:/t/OpsBridge/ux/EaqdIG4O6upBpQ_W4l7m8W4BVudGql28Sne8APvnAUilUw?e=TzZNhM)
- Styling: Add logical operator: Wrong indentation / lines
- Re-enable end-to-end tests
- Drag and drop, copy and paste
- Integrate with OBM

PR #1377, opened 2020-07-16
EL-3986 - Angular 10

donna-marijne:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3986

* Update build and package dependencies for Angular 10.
* Update plunker dependencies in line with the documentation app.
* Remove Angular 7 from peer dependencies.
* Update Getting Started with compatible Angular versions.
* Remove the custom typescript-transform script and replace with downlevel-dts. 
* Remove `audioTracks` and `videoTracks` from MediaPlayerService since they return properties which are not supported in the browsers we support, and the typescript definitions have been removed. The removed functions are undocumented and unused.
* Restore the `@Optional` constructor parameter to ColorService, since Angular 7 is no longer supported.
* Updated some comments referring to the `providedIn: 'root'` issue. The issue also occurs in Angular 8 applications. Presumably this is a packaging issue because `providedIn: 'root'` has been supported since Angular 6.
* Correct some errors in tests, noticed when building the e2e pages in a CLI app.
* Add `@Injectable` to `BaseResizableTableService`, required because it uses Angular lifecycle hooks.
* Remove lifecycle hooks from `BaseResizableTableDirective`. It should have a `@Directive()` decoration, but this (without a selector) would not be supported in Angular 8. The lifecycle hooks will instead be done in the derived classes until Angular 8 support can be removed.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3986-angular-10

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3986-angular-10~CI/)


package.json
```diff
@@ -128,7 +128,7 @@
     "load-grunt-config": "0.19.2",
     "marked": "^0.8.0",
     "mini-css-extract-plugin": "^0.4.4",
-    "ng-packagr": "5.5.0",
+    "ng-packagr": "^9.1.5",
```
ashley-hunter, line 131: Just want to verify we weren't able to use v10 of ng-packagr? I know we avoided v9 when we updated to Angular 9 because we got build warnings:

image

Due to an issue where it didn't remove interface imports from the generated bundle:
https://github.com/ng-packagr/ng-packagr/issues/1559

Its harmless as tree-shaking will remove them as unused imports, but it was supposedly fixed in v10 of ng-packagr, just wanted to check if there were issues using v10

donna-marijne: Unfortunately I forgot where I saw the "Class constructor x cannot be invoked without 'new'" error. But according to this we lose our es5 bundle when we upgrade to ng-packagr@10:

https://github.com/angular/angular-cli/issues/18067#issuecomment-651647074

Do we still need that for IE11 support? At least for apps which don't use Angular CLI, like our documentation?

ashley-hunter: From my understanding (I could be wrong) but we don't need to ship an es5 bundle any more regardless of whether we need to support IE11. It was primarily for the bundling tools (e.g. webpack, rollup, systemjs) as they select the format that best suits them. In the case of CLI apps that rely on Webpack, as far as I know they will all be using the esm2015/fesm2015 bundles and have been for quite some time.

Essentially, the downleveling to ES5 happens as part of the application build rather than when we build our library so it removes the need for these other formats.

However if you are getting another issue anyway then perhaps we can't update it anyway.

ashley-hunter: Ah I just saw the issue link you posted in which case there could be an issue as application downleveling was a change in the Angular 10 CLI, so just ignore this point, we can just ignore the warnings

donna-marijne: Okay that's good to know. Let me try to reproduce the issue, there might be another workaround for it that allows us to upgrade properly.

ashley-hunter: Based on this comment in the issue you linked:

image

Looks like we will need to stick with ng-packagr v9, as application downleveling was only added in v10 so we will still need to provide es5 bundles 🙄 my misunderstanding, sorry.

donna-marijne: Thanks for confirming, then I guess we are stuck with this version until Angular 12 comes out.

ashley-hunter: Looks like it! That'll be fine, when v11 comes out we can perform Ivy builds using ng-packagr 9 so that will be an improvement we can make without having to upgrade before v12.

docs/app/services/playground/utilities/system-helper.ts
```diff
@@ -1,7 +1,7 @@
 import { DocumentationType } from '../tokens/documentation.token';
 
-const ANGULAR_VERSION = '9';
-const CDK_VERSION = '9';
+const ANGULAR_VERSION = '10';
+const CDK_VERSION = '10';
 const D3_VERSION = '5.9.2';
```
ashley-hunter, line 5: Should this be updated to match the downgraded d3 version?

src/package.json
```diff
@@ -67,13 +67,13 @@
     "@types/dragula": "^2.1.33"
   },
   "peerDependencies": {
```
ashley-hunter, line 69: Should this file also be updated to match the downgraded d3 version?

donna-marijne: I'm not fully sure that downgrading is the correct solution now. I built the library with d3@5.16 and installed it into a CLI project. In production mode I did not see the rgb error. So it seems like it's something to do with our documentation app.

Also noticed that this is broken in our 2.0.6 release, so one option is to make another ticket and deal with it separately.

https://uxaspects.github.io/UXAspects/#/charts/donut-charts#nested-donut-chart

If you have any ideas, let me know. I'll look into it a bit further.

ashley-hunter: Hmm, thats interesting! Perhaps we need to tweak our webpack optimization settings. I will take a look into it, thanks!

donna-marijne: Yes iot's probably something we are doing slightly differently in our webpack config. However I think I should revert the last commit here and deal with that at a later date, what do you think?

ashley-hunter: Yes, I don't think we should downgrade d3. I'm making our config the same as the CLI's to see if that makes any difference, if not then yeah we should raise another ticket.

donna-marijne: That's a good idea, thanks for your help!

ashley-hunter: I have just pushed some changes which should hopefully fix the issue 🤞

donna-marijne: Thanks! It looks good. I'll make the same changes in the other repo.

ashley-hunter: Thank you!

donna-marijne: Unfortunately those changes do not work in the downstream repo. The site will not load in production. Where did you find the changes to make?

I'm still inclined to defer this to another sprint. We are losing a lot of QA time to a pre-existing issue now.

ashley-hunter: I'm not sure of the exact cause, I believe it could be the version of the TerserPlugin, so you might be able to just update that. I just updated the config and dependencies to match those used in the CLI. I'm happy to defer it if its not trivial 👍!

donna-marijne: It has the same dependencies as far as I can see. However the webpack config there is significantly different from this one. I tried a couple of other changes but it's breaking in various ways so I'll leave it until we have some time to spend on it.

ashley-hunter: Sounds good! Are you happy to keep the changes in this repo? Or would you like me to revert them?

donna-marijne: I think it's fine here. It should not impact the library or the internal documentation site as far as I'm aware. The downstream build is running so we can verify when that's complete.

donna-marijne: Seems like it's now working okay with the older dependencies, so we can probably leave it until we switch over to Angular CLI:

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/caf_CI_ux-aspects-micro-focus_EL-3986-angular-10/#/ui-components/charts#nested-donut-chart

donna-marijne:
Looks like Dean is away for the day and hasn't given any feedback so far so I'll just merge this now.

PR #1390, opened 2020-07-29
EL-4014 Column Picker: deselected not updating; unexpected sorting

DeanHaire:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4014

- Fixed issue where deselected array is not modified when columns are moved
- Remove default sorting
- Allow sorting via `sort` input
- Documented `sort` input

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4014-deselected-updating-sorting

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4014-deselected-updating-sorting~CI/)

### DIFF
--- a/docs/app/pages/components/components-sections/tables/column-picker/snippets/sort.ts
+++ b/docs/app/pages/components/components-sections/tables/column-picker/snippets/sort.ts
@@ -1,2 +1,5 @@
-(a: ColumnPickerGroupItem, b: ColumnPickerGroupItem) =>
-    (a['name'] ?? a).localeCompare(b['name'] ?? b);
+(a: ColumnPickerGroupItem, b: ColumnPickerGroupItem) => {
+    const aCombined = a.group ? `${a.group}${a.name}` : a.name;
+    const bCombined = b.group ? `${b.group}${b.name}` : b.name;
+    return aCombined.localeCompare(bCombined);
+}
--- a/docs/app/pages/components/components-sections/tables/column-picker/snippets/sort.ts
+++ b/docs/app/pages/components/components-sections/tables/column-picker/snippets/sort.ts
@@ -1,5 +1,5 @@
-(a: ColumnPickerGroupItem, b: ColumnPickerGroupItem) => {
+sort = (a: ColumnPickerGroupItem, b: ColumnPickerGroupItem) => {
     const aCombined = a.group ? `${a.group}${a.name}` : a.name;
     const bCombined = b.group ? `${b.group}${b.name}` : b.name;
     return aCombined.localeCompare(bCombined);
-}
+};
--- a/docs/app/pages/components/components-sections/tables/column-picker/snippets/sort.ts
+++ b/docs/app/pages/components/components-sections/tables/column-picker/snippets/sort.ts
@@ -1,0 +1,2 @@
+sort = (a: string | ColumnPickerGroupItem, b: string | ColumnPickerGroupItem) =>
+    (a['name'] ?? a).localeCompare(b['name'] ?? b);
--- a/docs/app/pages/components/components-sections/tables/column-picker/snippets/sort.ts
+++ b/docs/app/pages/components/components-sections/tables/column-picker/snippets/sort.ts
@@ -1,2 +1,2 @@
-sort = (a: string | ColumnPickerGroupItem, b: string | ColumnPickerGroupItem) =>
+(a: ColumnPickerGroupItem, b: ColumnPickerGroupItem) =>
     (a['name'] ?? a).localeCompare(b['name'] ?? b);
--- a/src/components/table/column-picker/column-picker.component.spec.ts
+++ b/src/components/table/column-picker/column-picker.component.spec.ts
@@ -161,7 +161,7 @@ export class ColumnPickerSortTestComponent {
     ];
 
     sort = (a: string | ColumnPickerGroupItem, b: string | ColumnPickerGroupItem) =>
-        (a['name'] ?? a).localeCompare(b['name'] ?? b);
+        (a['name'] ?? a).localeCompare(b['name'] ?? b)
 }
 
 describe('Column Picker Component - Sort Input', () => {
--- a/src/components/table/column-picker/column-picker.component.spec.ts
+++ b/src/components/table/column-picker/column-picker.component.spec.ts
@@ -50,7 +50,7 @@ export class ColumnPickerTestComponent {
     ];
 }
 
-fdescribe('Column Picker Component', () => {
+describe('Column Picker Component', () => {
     let component: ColumnPickerTestComponent;
     let fixture: ComponentFixture<ColumnPickerTestComponent>;
     let nativeElement: HTMLElement;
@@ -164,7 +164,7 @@ export class ColumnPickerSortTestComponent {
         (a['name'] ?? a).localeCompare(b['name'] ?? b);
 }
 
-fdescribe('Column Picker Component - Sort Input', () => {
+describe('Column Picker Component - Sort Input', () => {
     let component: ColumnPickerSortTestComponent;
     let fixture: ComponentFixture<ColumnPickerSortTestComponent>;
     let nativeElement: HTMLElement;
--- a/src/components/table/column-picker/column-picker.component.spec.ts
+++ b/src/components/table/column-picker/column-picker.component.spec.ts
@@ -1,6 +1,7 @@
 import { Component } from '@angular/core';
 import { ComponentFixture, async, TestBed } from '@angular/core/testing';
 import { TableModule } from '../table.module';
+import { ColumnPickerGroupItem, ColumnPickerGroup } from './column-picker.component';
 
 @Component({
     selector: 'app-column-picker',
@@ -16,10 +17,7 @@ import { TableModule } from '../table.module';
 export class ColumnPickerTestComponent {
     selected: ReadonlyArray<string> = [
         'Type',
-        'Date',
-        'Requested by',
-        'Status',
-        'Completion'
+        'Date'
     ];
 
     locked: ReadonlyArray<string> = ['ID'];
@@ -27,22 +25,8 @@ export class ColumnPickerTestComponent {
     deselected: Array<string | ColumnPickerGroupItem> = [
         { group: 'Metadata', name: 'Department' },
         { group: 'Metadata', name: 'Author' },
-        { group: 'Metadata', name: 'Category' },
-        { group: 'Metadata', name: 'Date Created' },
-        { group: 'Metadata', name: 'Date Modified' },
         'Document ID',
-        'Flag',
-        'From',
-        'Icon',
-        'Importance',
-        'Location',
-        'Location ID',
-        'Message',
-        { group: 'Metadata', name: 'Organization' },
-        'Time',
-        'Time Created',
-        'Time Modified',
-        'Work Completed'
+        'Flag'
     ];
 
     groups: ColumnPickerGroup[] = [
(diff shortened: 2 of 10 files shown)


src/components/table/column-picker/column-picker.component.ts
```diff
@@ -176,11 +171,31 @@ export class ColumnPickerComponent implements OnChanges {
     }
 
     /** A function that can be called to add columns. If no columns are passed to the function, the items that are selected in the left column will be added. */
-    addColumns(columns: ReadonlyArray<string | ColumnPickerGroupItem> = this._deselectedSelection): void {
-        columns = columns.filter(column => this.selected.indexOf(this.getColumnName(column)) === -1);
+    addColumns(columns: ReadonlyArray<string | ColumnPickerGroupItem | ColumnPickerTreeNode> = this._deselectedSelection): void {
```
donna-marijne, line 174: What is the situation where `ColumnPickerTreeNode`s can be mixed in here? I've checked where it is called and the type indicates `ReadonlyArray<string | ColumnPickerGroupItem>` in every case.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -176,11 +171,31 @@ export class ColumnPickerComponent implements OnChanges {
...
+                return item;
+            }
+            if ('group' in column) { return column; }
+        });
+
+        columns = columns.filter(column => this.selected.indexOf(column) === -1);
 
         // add each item to the selected columns list
-        columns.forEach(column => this.selected = [...this.selected, this.getColumnName(column)]);
+        columns.forEach(column => this.selected = [...this.selected, column]);
```
donna-marijne, line 189: This should probably be 
```ts
this.selected = [...this.selected, ...columns];
```

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -176,11 +171,31 @@ export class ColumnPickerComponent implements OnChanges {
...
+
+        this.deselected = this.deselected.filter(deselectedColumn => {
```
donna-marijne, line 191: This should probably be
```ts
this.deselected = this.deselected.filter(column => columns.indexOf(column) === -1);
```

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -316,7 +328,7 @@ export class ColumnPickerComponent implements OnChanges {
     }
 
     /** Get the column name based on type */
-    private getColumnName(item: string | ColumnPickerGroupItem): string {
+    getColumnName(item: string | ColumnPickerGroupItem): string {
```
donna-marijne, line 331: We usually prefix private methods that need to be called in the view with an underscore.

src/components/table/column-picker/column-picker.component.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+        nativeElement = fixture.nativeElement;
+        fixture.detectChanges();
+    });
+
+    it('should display the correct number of deselected columns', async () => {
+        expect(component.deselected.length).toEqual(18);
+        let deselected = nativeElement.querySelector('.column-picker-list').querySelectorAll('.column-picker-list-item').length;
+        expect(deselected).toEqual(18);
+
+        component.deselected = ['Document ID', 'Flag', 'From', 'Icon'];
```
donna-marijne, line 77: I've found that this type of test is better as two separate tests, then it's clearer which scenario is failing. In this case you might introduce a second test named "should display the correct number of deselected columns after modification".

src/components/table/column-picker/column-picker.component.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+        expect(deselected).toEqual(4);
+    });
+
+    it('should not sort the deselected columns when the sort input is undefined', () => {
+        expect(component.deselected[0]).toEqual({ group: 'Metadata', name: 'Department' });
+    });
+
+    it('should update the number of deselected columns available to select', async () => {
+        let numberOfDeselectedColumns = nativeElement.querySelector('.column-picker-stats');
+        expect(numberOfDeselectedColumns.textContent.trim()).toEqual('0 of 18 selected');
```
donna-marijne, line 93: I think it would be reasonable to add this check into the first test since it's related.

src/components/table/column-picker/column-picker.component.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+
+        numberOfDeselectedColumns = nativeElement.querySelector('.column-picker-stats');
+        expect(numberOfDeselectedColumns.textContent.trim()).toEqual('0 of 4 selected');
+    });
+
+    it('should update the number of select columns added', async () => {
+        let numberOfSelectedColumns = nativeElement.querySelectorAll('.column-picker-stats')[1];
+        expect(numberOfSelectedColumns.textContent.trim()).toEqual('6 columns added');
+
+        component.selected = [];
```
donna-marijne, line 108: This should also be two tests.

src/components/table/column-picker/column-picker.component.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+        fixture = TestBed.createComponent(ColumnPickerSortTestComponent);
+        component = fixture.componentInstance;
+        nativeElement = fixture.nativeElement;
+        fixture.detectChanges();
+    });
+
+    it('should sort grouped columns when the sort input is provided', () => {
+        let names = [];
+        nativeElement.querySelectorAll('.column-picker-tree-node-level-1')
+            .forEach(element => names = [...names, element.textContent]);
```
donna-marijne, line 189: I recommend adding helper functions to make the test body easier to read. I did something pretty similar to this earlier this week, which might help to refer to.
https://github.com/UXAspects/UXAspects/blob/develop-2.1.1/src/directives/tree-grid/tree-grid.spec.ts#L165

src/components/table/column-picker/column-picker.component.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    it('should not sort the selected columns', () => {
+        const selected = nativeElement.querySelectorAll('.column-picker-list')[1].querySelectorAll('.column-picker-list-item');
+        selected.forEach((column, i) => {
+            expect(column.textContent.trim()).toEqual(component.selected[i]['name'] || component.selected[i]);
+        });
+    });
+});
+
+export interface ColumnPickerGroupItem {
```
donna-marijne, line 208: Import this instead of redefining it.

src/components/table/column-picker/column-picker.component.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+    group?: string;
+    name: string;
+}
+
+export interface ColumnPickerGroup {
```
donna-marijne, line 213: Import this as well.

src/components/table/column-picker/column-picker.component.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+    });
+
+    it('should sort grouped columns when the sort input is provided', () => {
+        let names = [];
+        nativeElement.querySelectorAll('.column-picker-tree-node-level-1')
+            .forEach(element => names = [...names, element.textContent]);
+        expect(names).toEqual(names.slice().sort());
+    });
+
+    it('should sort ungrouped columns when the sort input is provided', () => {
```
donna-marijne, line 193: I would suggest checking that the entire column is in the expected state rather than checking grouped and ungrouped separately.
For example:
```ts
expect(deselectedRows[0].innerText).toBe('Metadata', 'row 0');
expect(deselectedRows[1].innerText).toBe('Author', 'row 1');
```

This will be a lot more manageable if you trim the data set down to a minimal number of elements required to assert correct behavior.

src/components/table/column-picker/column-picker.component.spec.ts
```diff
@@ -0,0 +1,216 @@
...
+        'Importance',
+        'Location',
+        'Location ID',
+        'Message',
+        { group: 'Metadata', name: 'Organization' },
+        'Time',
+        'Time Created',
+        'Time Modified',
+        'Work Completed'
+    ];
```
donna-marijne, line 157: You can make this array a lot shorter without affecting coverage. It will make the tests simpler.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -98,74 +101,78 @@ export class ColumnPickerComponent implements OnChanges {
 
     /** Parse data into suitable format for the FlatTreeComponent to understand and initialize deselect tree */
     private rebuildDeselectTree(): void {
```
donna-marijne, line 103: This function is too large and needs to be refactored. Have a read of this, you should have access with your corporate credentials:

https://learning.oreilly.com/library/view/clean-code/9780136083238/chapter03.html

It should look something like
```ts
private rebuildDeselectTree(): void {
  const columns = this.getDeselectedColumnsInOrder();
  this._treeData = this.createTreeData(columns);
  // the rest is fine
}
```

A common practice in Angular is to do data manipulation in an associated service, so that the component is just handling presentation stuff and user input.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -98,74 +101,78 @@ export class ColumnPickerComponent implements OnChanges {
...
-        // create grouped columns and their parent nodes
-        groupedColumns.forEach((column: ColumnPickerGroupItem) => {
-            // if new group create parent node
-            if (!currentGroup || column.group !== currentGroup) {
-                // set children for current group
-                if (currentGroup != null) {
-                    treeData.find(node => node.name === currentGroup).children = children;
-                    children = [];
+        columns.forEach((column: ColumnPickerGroupItem) => {
+            if (groupedColumns.indexOf(column) != -1) {
```
donna-marijne, line 135: Always use `!==`.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -199,12 +208,13 @@ export class ColumnPickerComponent implements OnChanges {
         // remove each item from the selected columns list
         this.selected = this.selected.filter(column => columns.indexOf(column) === -1);
 
-        // add columns to deselected if not already there
+        const deselectedColumns = columns.filter(column => {
+            return this.deselected.indexOf(column) === -1 && this.deselected.indexOf({ name: column['name'], group: undefined}) === -1;
+        });
```
donna-marijne, line 225: This still doesn't look right, but you can probably fix this after resolving the above issues.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -288,8 +298,8 @@ export class ColumnPickerComponent implements OnChanges {
     }
 
     /** Provide a trackBy function for the reorderable options */
-    selectedTrackBy(index: number, column: string): string {
-        return index + column;
+    selectedTrackBy(index: number, column: string | ColumnPickerGroupItem): string {
+        return index + (column['name'] || column);
```
donna-marijne, line 314: `object['property']` syntax should be avoid wherever possible due to bypassing type checking. Consider implementing something like this: https://stackoverflow.com/a/44078574

```ts
return isColumnPickerGroupItem(column) ? column.name : column;
```

This would be more expressive. There are several places that this could be used, including the tests.

docs/app/pages/components/components-sections/tables/column-picker/snippets/sort.ts
```diff
@@ -0,0 +1,5 @@
+sort = (a: ColumnPickerGroupItem, b: ColumnPickerGroupItem) => {
+    const aCombined = a.group ? `${a.group}${a.name}` : a.name;
+    const bCombined = b.group ? `${b.group}${b.name}` : b.name;
```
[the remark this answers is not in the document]
DeanHaire: I think I might still need it. If `a.group` is undefined then the result would be `undefined${a.name}`, which would result in incorrect sorting.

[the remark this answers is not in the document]
donna-marijne: Sorry, I misread the updated code, you are right.

I've noticed that it has a failure case if you also have a property with the same name as a group. I think this is a problem with the code that converts the columns into tree nodes, however. If we allow arbitrary sorting, we will have to account for the fact that all items with the same group might not end up being consecutive, so the code that deals with the output may have to be more tolerant.

In pseudocode, it could be something like
```
for each column
  if column is group item
    groupNode = createOrFindGroupNode(column, treeData)
    groupNode.children.push(createTreeNode(column))
  else
    treeData.push(createTreNode(column))
```

src/components/table/column-picker/column-picker.component.ts
DeanHaire: Most of this code was already here, but with my additions I agree that it's definitely too long. I've moved it into a service as you suggested.

donna-marijne: Yes, it already needed refactoring before you started, thanks for taking care of it now.

src/components/table/column-picker/column-picker.service.ts
```diff
@@ -0,0 +1,77 @@
...
+    private createTreeNode(column: ColumnPickerGroupItem, treeData: ColumnPickerTreeNode[], level: number): void {
+        treeData.push({
+            name: column.name,
+            level,
+            expandable: false,
+            column
+        });
+    }
+
+    private isColumnPickerGroupItem(column: string | ColumnPickerGroupItem): column is ColumnPickerGroupItem {
```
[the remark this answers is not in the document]
DeanHaire: I removed the duplicate. However, `column is ColumnPickerGroupItem` seems to be the recommended approach - https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards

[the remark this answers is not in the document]
donna-marijne: Thank you, I was not aware of this technique.

PR #1392, opened 2020-08-06
EL-3959 - Date & Time Picker: incorrect date selected

DeanHaire:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3959

- `Date`'s set date, month and year methods accept invalid values and will automatically correct them. In this case, setting the month value before the day can cause a month to be incorrectly skipped. Using `setMonth` with both month and day values ensures that the date is correctly set.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3959-datepicker-incorrectdate

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3959-datepicker-incorrectdate~CI/)

### DIFF
--- a/e2e/tests/components/date-time-picker/date-time-picker.e2e-spec.ts
+++ b/e2e/tests/components/date-time-picker/date-time-picker.e2e-spec.ts
@@ -190,8 +190,12 @@ describe('Date Time Picker Tests', () => {
 
         // select 1st February
         date = await page.getDate('1', true);
-        await date.click();
 
-        expect(await page.getCurrentDate()).toBe('February 1, 2019, 12:00:00 PM');
+        if (await page.isClickable(date)) {
+            expect(await page.getCurrentDate()).toBe('February 1, 2019, 12:00:00 PM');
+        } else {
+            expect(await page.getCurrentDate()).toBe('January 31, 2019, 12:00:00 PM');
+        }
+
     });
 });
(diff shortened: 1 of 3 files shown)


e2e/tests/components/date-time-picker/date-time-picker.e2e-spec.ts
```diff
@@ -182,4 +182,20 @@ describe('Date Time Picker Tests', () => {
...
+
+    it('should correctly set the date when the current day value does not exist in the selected month', async () => {
+        // select 31st January
+        let date = await page.getDate('31');
+        await date.click();
+
+        // select 1st February
+        date = await page.getDate('1', true);
+
+        if (await page.isClickable(date)) {
```
ashley-hunter, line 194: Is the purpose of this because in the MF theme future and previous month buttons are invisible?

image

DeanHaire: Yep

ashley-hunter: Could we perhaps add a comment here describing the reason as it is not immediately obvious. Also perhaps the `isClickable` function might be better called `attemptClick` as `isClickable` suggests it is a getter with no side-effects when it is not so much checking if the element is clickable but performing a click and returning a success state

DeanHaire: Good point, I've made those changes to make it more clear 👍

PR #1396, opened 2020-08-21
EL-3926 - Select Disabled Options

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3926

Previously the disabled property was added to the item objects in the typeahead themselves, this meant having to update them any time selection changed etc... This was hard when using paging as the infinite scroll directive keeps these items in memory too, so they could further get out of sync. It also meant there was a function on the class to check if it something was disabled and a property on the item to store the disabled state and they could return different results if they are out of sync.

The typeahead also accepted an input called `disabledOptions` which was documented but was never used by the typeahead so was technically broken. I have removed this internal isDisabled property on each item and instead updated to use the `disabledOptions` input. As this is separate from the items this is no longer an issue with the infinite scroll directive and there now is only one way to check the disabled state ensuring they remain in sync.

- Updated documentation
- Added tests
- Fixed some typing issues
- Fixed a case where we assumed an object was defined when it might not be (occurred when switching from `object` types to `string`.
- Updated TSLint config to allow Angular type coercion static properties, e.g. `ngAcceptInputType_disabled`
- Removed `debugger` left in tests
- Removed a lot of unused code in `Typeahead` test

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3926-Select-Disabled-Options

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3926-Select-Disabled-Options~CI/)

### DIFF
--- a/docs/app/pages/components/components-sections/input-controls/typeahead/typeahead.component.html
+++ b/docs/app/pages/components/components-sections/input-controls/typeahead/typeahead.component.html
@@ -150,7 +150,7 @@
         <ul>
             <li><code>option: T</code> - the string or custom object representing the option.</li>
             <li><code>api: TypeaheadOptionApi</code> - provides the functions <code>getKey</code>,
-                <code>getDisplay</code> and <code>getDisplayHtml</code>. See the section below for more
+                <code>getDisplay</code>, <code>getDisplayHtml</code> and <code>getDisabled</code>. See the section below for more
                 information.</li>
         </ul>
     </tr>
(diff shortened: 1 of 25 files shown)


DeanHaire (dismissed):

docs/app/pages/components/components-sections/input-controls/typeahead/typeahead.component.html
```diff
@@ -202,14 +202,17 @@ <h4>TypeaheadOptionApi</h4>
...
+    <tr uxd-api-property name="getDisplay" args="option: T" returns="string">
         Returns the display value of the given <code>option</code>.
     </tr>
     <tr uxd-api-property name="getDisplayHtml" args="option: T" returns="string">
         Returns the display value of the given <code>option</code> with HTML markup added to highlight
         the part which matches the current <code>filter</code> value. Override the
         <code>ux-filter-match</code> class in CSS to modify the default appearance.
     </tr>
+    <tr uxd-api-property name="getDisabled" args="option: T" returns="boolean">
+        Returns the disabled state of the given <code>options</code>.
```
donna-marijne, line 214: Should be `option` (singular).

donna-marijne (changes requested):
Thank you for the fix and comprehensive tests! I just noticed a typo.

donna-marijne (approved):
Thank you, all looks good.

PR #1397, opened 2020-08-21
EL-4061 - Tag Input Focus Out

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4061

After some investigation the issue was slightly different than I think we initially suspected. Neither stopping propagation or creating a new output was the solution to the issue. 

The issue is caused by a Chrome bug, that will blur an input when a scrollbar is used if a parent element has a tabindex set. So the side panel is one potential cause as it has a tabindex="0", but really it could be any parent element.

The solution is to `preventDefault` if a mousedown event occurs within the scrollbar region of the typeahead.

Due to this, I couldn't find a way to add a test for this, as it requires the scroll to be performed using the mouse, its not standard behaviour that occurs whenever general scrolling happens, but if you have any suggestions I'm glad to try them out.

Additionally I improved some typing:
- The Tag template context was defined in select, but not in tag input, I have now changed this to provide correct context typing for both tag input and select.
- The `display` type was incorrect, it should have been a function that returns a string or a string, however due to slightly confusing typescript semantics it was checking only for a function the returned (string | string). So a string value was not correct. I have split the function into its own type to fix this.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4061-Tag-Input-Focus-Out

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4061-Tag-Input-Focus-Out~CI/)

### DIFF
--- a/src/components/tag-input/tag-input.component.ts
+++ b/src/components/tag-input/tag-input.component.ts
@@ -63,7 +63,7 @@ export class TagInputComponent<T = any> implements AfterContentInit, OnChanges, 
      * the option object as an argument, and should return the appropriate display value.
      * If the name of a property is provided as a string, that property is used as the display value.
      */
-    @Input() display: (option: T) => string | string;
+    @Input() display: TagInputDisplayFn<T> | string;
 
     /** Controls whether pasting text into the text input area automatically converts that text into one or more tags. */
     @Input() addOnPaste: boolean = true;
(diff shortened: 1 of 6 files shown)


Sdhilipkumar (approved):
Looks good.

src/components/tag-input/tag-input.component.ts
```diff
@@ -829,3 +829,11 @@ export interface TagApi<T = any> {
  * The function used to return custom class information, for use in `ngClass`.
  */
 export type TagClassFunction<T = any> = (tag: T, index: number, selected: boolean) => (string | string[] | Set<string>);
+
+export type TagInputDisplayFn<T> = (option: T) => string;
```
donna-marijne, line 833: I would suggest `TagInputDisplayFunction` to match the existing function type name.

donna-marijne (changes requested):
Thank you, looks like it works well, just had one suggestion.

src/components/tag-input/tag-input.component.ts
ashley-hunter: Thanks Donna, that should now be updated

donna-marijne (approved):
Looks good, thank you!

PR #1402, opened 2020-08-24
EL-3957 - Filter IDs

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3957

- Adding id Input to filter dropdown and filter dynamic component.
- Adding `id` property to the `Filter` interface
-  Updating documentation
- Adding karma tests to assert IDs
- Updated components to use `OnPush` change detection

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3957-Filter-Ids

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3957-Filter-Ids~CI/)

### DIFF
--- a/docs/app/pages/components/components-sections/tables/filters/filters.component.html
+++ b/docs/app/pages/components/components-sections/tables/filters/filters.component.html
@@ -6,34 +6,34 @@
 
 <table class="table table-hover">
     <thead>
-    <tr>
-        <th class="text-black">NAME</th>
-        <th class="text-black">AUTHOR</th>
-        <th class="text-black">DATE MODIFIED</th>
-        <th class="text-black">WORK COMPLETED</th>
-        <th class="text-black text-center">STATUS</th>
-    </tr>
+        <tr>
+            <th class="text-black">NAME</th>
+            <th class="text-black">AUTHOR</th>
+            <th class="text-black">DATE MODIFIED</th>
+            <th class="text-black">WORK COMPLETED</th>
+            <th class="text-black text-center">STATUS</th>
+        </tr>
     </thead>
     <tbody>
-    <tr class="clickable" *ngFor="let document of documents">
-        <td>{{ document.name }}</td>
-        <td>{{ document.author }}</td>
-        <td>{{ document.date }}</td>
-        <td>
-            <ux-spark theme="chart2"
-                      [value]="document.completed"
-                      barHeight="3"
-                      [inlineLabel]="document.completed">
-            </ux-spark>
-        </td>
-        <td class="text-center vertical-center-icon">
-            <ux-icon
-                *ngIf="document.active"
-                name="checkmark"
-                [attr.aria-label]="document.active ? 'Approved' : 'Not approved'">
-            </ux-icon>
-        </td>
-    </tr>
+        <tr class="clickable" *ngFor="let document of documents">
+            <td>{{ document.name }}</td>
+            <td>{{ document.author }}</td>
+            <td>{{ document.date }}</td>
+            <td>
+                <ux-spark theme="chart2"
+                          [value]="document.completed"
+                          barHeight="3"
+                          [inlineLabel]="document.completed">
+                </ux-spark>
+            </td>
+            <td class="text-center vertical-center-icon">
+                <ux-icon
+                    *ngIf="document.active"
+                    name="checkmark"
+                    [attr.aria-label]="document.active ? 'Approved' : 'Not approved'">
+                </ux-icon>
+            </td>
+        </tr>
     </tbody>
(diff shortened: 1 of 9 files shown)


src/components/filters/filter-dynamic/filter-dynamic.component.ts
```diff
@@ -85,13 +101,15 @@ export class FilterDynamicComponent implements OnInit, OnDestroy {
...
+
+            this._changeDetector.markForCheck();
         });
 
         // get the items to be displayed in the typeahead
         this.typeaheadItems = this.getItems();
 
         // determine if we should show the typeahead control
-        if (this.options && this.options.maxIndividualItems && this.options.maxIndividualItems + 1 >= this.filters.length) {
+        if (this.options?.maxIndividualItems && this.options.maxIndividualItems + 1 >= this.filters.length) {
```
DeanHaire, line 112: I personally think if statements should be easy to skim read and therefore be able to understand the purpose of the check without needing to delve into the actual logic. In cases like this, I think it would be beneficial to extract this into a well named variable that explains its purpose (which also removes the need for the comment above). Maybe something like `isTypeaheadRequired`? Then it would be `if (!isTypeaheadRequired)` here which I think makes it much easier to quickly understand the purpose and intent of the logic.

DeanHaire (commented):
Can you create a downstream build for this as well?

src/components/filters/filter-dynamic/filter-dynamic.component.ts
ashley-hunter: I have shortened the condition as part of it was unnecessary. I have moved it to a variable called `isTypeaheadHidden`. We are setting a `showTypeahead` variable so as this is the inverse the opposite would be `hide` or `hidden`. 

While I agree, if the code is clear then we should remove the comment, in this case I have updated the comment, as even with the variable name the logic can be hard to understand without context.

docs/app/pages/components/components-sections/tables/filters/filters.component.html
```diff
@@ -1,38 +1,39 @@
...
 <table class="table table-hover">
     <thead>
-        <tr>
-            <th class="text-black">NAME</th>
-            <th class="text-black">AUTHOR</th>
-            <th class="text-black">DATE MODIFIED</th>
-            <th class="text-black">WORK COMPLETED</th>
-            <th class="text-black text-center">STATUS</th>
-        </tr>
+    <tr>
```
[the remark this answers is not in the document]
donna-marijne: (This was a static firing test of the rocket emoji, and can safely be ignored.)

src/components/filters/filter-dynamic/filter-dynamic.component.ts
DeanHaire: I'm not sure if `isTypeaheadHidden` is accurate in this situation. `isTypeaheadHidden` implies to me that we're checking what the current state of the typeahead is but here we're actually checking if we need to hide the typeahead. So to me this currently reads as "if the typeahead is hidden then hide the typeahead" (which makes no sense) instead of "if the typeahead should be hidden, then hide the typeahead". Maybe something like `shouldHideTypeahead` would be better?

I agree that with the current variable name it's difficult to understand the context without the comment. But I think in general we should be aiming to remove the need for comments by using descriptive variable names. Maybe something like `numberOfFiltersExceedsMaxIndividualItems` (which while it's a long name, it does convey the purpose behind the if statement, which would then read as "if the number of filters exceeds the max individual items, then hide the typeahead"). I'm not sure, I'm fine if you want to stick with the updated comment for now though.

ashley-hunter: I'm happy to change it to `shouldHideTypeahead`. I'm not against long variable names where they bring value, however in this case the problem is more that `maxIndividualItems` probably wasn't a good name in the first place but we can't change that now. The term 'Individual items' does not make it obvious that these are the items that will always be visible and not in the typeahead list so I think its easier for someone who may not know the component very well to have the brief explanation here.

DeanHaire: Sounds good, I agree `maxIndividualItems` is probably the issue here. Think the current version is definitely clearer than what it initially was, so I'm happy with that 👍

DeanHaire (approved):

donna-marijne (approved):
Looks good, works well, thank you.

PR #1404, opened 2020-08-25
EL-3934 - menu-disabled-property

cmallon367:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3934

Made use of the host binding 'attr.disabled' to ensure that click events are not triggered when the menu-item is disabled
Added in a karma test to assert that no click event is emitted when disabled is true.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3934-menu-disabled-property

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3934-menu-disabled-property~CI/)

### DIFF
--- a/src/components/menu/menu-item/menu-item.component.ts
+++ b/src/components/menu/menu-item/menu-item.component.ts
@@ -1,5 +1,5 @@
 import { FocusableOption, FocusOrigin } from '@angular/cdk/a11y';
-import { ChangeDetectionStrategy, Component, ElementRef, HostListener, Input, OnDestroy, OnInit, Renderer2, Output, EventEmitter } from '@angular/core';
+import { ChangeDetectionStrategy, Component, ElementRef, EventEmitter, HostBinding, HostListener, Input, OnDestroy, OnInit, Output, Renderer2 } from '@angular/core';
 import { BehaviorSubject, Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import { isKeyboardTrigger } from '../../../common/index';
@@ -21,13 +21,22 @@ import { MenuItemType } from './menu-item-type.enum';
 export class MenuItemComponent implements OnInit, OnDestroy, FocusableOption {
 
     /** Define if this item is disabled or not */
-    @Input() disabled: boolean = false;
+    @Input() set disabled(value: boolean) {
+        this._disabled = value ? value : false;
+        this.disabledMenu = value ? value : null;
+    }
+
+    get disabled(): boolean {
+        return !this._disabled ? false : true;
+    }
 
     /** Define the role of the element */
     @Input() role: 'menuitem' | 'menuitemradio' | 'menuitemcheckbox' = 'menuitem';
 
     /** Emits when the menu item is clicked or the enter key is pressed. */
     @Output() activate = new EventEmitter<MouseEvent | KeyboardEvent>();
+
+    @HostBinding('attr.disabled') disabledMenu: boolean = null;
 
     /** Access the open state */
     get isOpen(): boolean {
@@ -54,6 +63,8 @@ export class MenuItemComponent implements OnInit, OnDestroy, FocusableOption {
 
     /** Automatically unsubscribe from observables on destroy */
     private readonly _onDestroy$ = new Subject<void>();
+
+    private _disabled: boolean = false;
 
     constructor(
         private readonly _menu: MenuComponent,
--- a/src/components/menu/menu-item/menu-item.component.ts
+++ b/src/components/menu/menu-item/menu-item.component.ts
@@ -1,5 +1,6 @@
 import { FocusableOption, FocusOrigin } from '@angular/cdk/a11y';
-import { ChangeDetectionStrategy, Component, ElementRef, EventEmitter, HostBinding, HostListener, Input, OnDestroy, OnInit, Output, Renderer2 } from '@angular/core';
+import { coerceBooleanProperty } from '@angular/cdk/coercion';
+import { ChangeDetectionStrategy, Component, ElementRef, EventEmitter, HostListener, Input, OnDestroy, OnInit, Output, Renderer2 } from '@angular/core';
 import { BehaviorSubject, Subject } from 'rxjs';
 import { takeUntil } from 'rxjs/operators';
 import { isKeyboardTrigger } from '../../../common/index';
@@ -12,6 +13,7 @@ import { MenuItemType } from './menu-item-type.enum';
     templateUrl: './menu-item.component.html',
     changeDetection: ChangeDetectionStrategy.OnPush,
     host: {
+        '[attr.disabled]': 'disabled ? "disabled" : null',
         '[attr.role]': 'role',
         '[class.disabled]': 'disabled',
         '[class.ux-menu-item]': 'true',
@@ -21,13 +23,12 @@ import { MenuItemType } from './menu-item-type.enum';
 export class MenuItemComponent implements OnInit, OnDestroy, FocusableOption {
 
     /** Define if this item is disabled or not */
-    @Input() set disabled(value: boolean) {
-        this._disabled = value ? value : false;
-        this.disabledMenu = value ? value : null;
+    @Input() set disabled(disabled: boolean) {
+        this._disabled = coerceBooleanProperty(disabled);
     }
 
     get disabled(): boolean {
-        return !this._disabled ? false : true;
+        return this._disabled;
     }
 
     /** Define the role of the element */
@@ -35,8 +36,6 @@ export class MenuItemComponent implements OnInit, OnDestroy, FocusableOption {
 
     /** Emits when the menu item is clicked or the enter key is pressed. */
     @Output() activate = new EventEmitter<MouseEvent | KeyboardEvent>();
-
-    @HostBinding('attr.disabled') disabledMenu: boolean = null;
 
     /** Access the open state */
     get isOpen(): boolean {
@@ -145,4 +144,5 @@ export class MenuItemComponent implements OnInit, OnDestroy, FocusableOption {
         this._renderer.setAttribute(this._elementRef.nativeElement, 'tabindex', isTabbable ? '0' : '-1');
     }
 
+    static ngAcceptInputType_disabled: boolean | string;
 }
--- a/src/components/menu/menu.spec.ts
+++ b/src/components/menu/menu.spec.ts
@@ -13,6 +13,7 @@ import { MenuModule } from './menu.module';
                 type="button"
                 id="menu-item-1"
                 uxMenuItem
+                [disabled]="disabled"
                 (activate)="item1Activated($event)">
                 Item One
             </button>
@@ -22,7 +23,7 @@ import { MenuModule } from './menu.module';
                 id="menu-item-disabled"
                 uxMenuItem
                 [disabled]="disabled"
-                (click)="onButtonClick()"
+                (click)="item2Click()"
                 #subMenuTrigger="ux-menu-trigger"
                 [uxMenuTriggerFor]="subMenu">
                 Item Two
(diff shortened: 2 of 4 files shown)


src/components/menu/menu-item/menu-item.component.ts
```diff
@@ -21,14 +21,23 @@ import { MenuItemType } from './menu-item-type.enum';
 export class MenuItemComponent implements OnInit, OnDestroy, FocusableOption {
 
     /** Define if this item is disabled or not */
-    @Input() disabled: boolean = false;
+    @Input() set disabled(value: boolean) {
```
ashley-hunter, line 24: Hey Christina, I realise this is just a draft, but just to give you a little bit of insight here, this is a common pattern called coercion, and the Angular CDK offers some utilities to help us.

```typescript
@Input() set disabled(disabled: boolean) {
    this._disabled = coerceBooleanProperty(disabled);
}

get disabled(): boolean {
    return this._disabled;
}
```

This way we don't have to do the ternary in the getter as all conversion is done in the setter. This also allows us to use string values in our html to avoid having to use bindings, e.g.

```html
<button disabled="true"
```

As the `coerceBooleanProperty` function will also convert strings. The problem is that in Angular you can turn on strict type checking, which ensures that you use the correct types in your templates. As `disabled` is expecting a boolean, if I use a string value this will throw an error.

So Angular provides us a way to allow the template to use string or a boolean, but if you are using the disabled property in code we only want to allow boolean.

To do this we add the following property to our component class:

```typescript
static ngAcceptInputType_disabled: boolean | string;
```

This tells Angular to allow a string or boolean type in a template. The CDK also gives us a `BooleanInput` type rather than having to use `boolean | string` each time, but its only in v9 and up so we need to wait until we drop v8 support.

As a side note, you should be able to remove:

```typescript
@HostBinding('attr.disabled') disabledMenu: boolean = null;
```

As you can either add the `HostBinding` to the disabled property as well, or use the host property in the component decorator, e.g.:

```typescript
@Component({
    ....
   host: {
      '[attr.disabled]': 'disabled'
   }
})
```

Which should help simply the code a bit. I believe there is a little bit written about coercion in the Tips and Tricks section, but it could probably be a bit more detailed, so perhaps this might be a good thing to add.

cmallon367: Thanks Ashley for the insight it's definitely very helpful and I will do a little more research into it this afternoon and update the tips & tricks page to what I find out

ashley-hunter: Thank you 😊

src/components/menu/menu.spec.ts
```diff
@@ -376,6 +382,32 @@ describe('MenuComponent', () => {
...
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        // get the child submenu item
+        const menuItem = document.querySelector<HTMLButtonElement>(
+            '#menu-item-disabled'
+        );
+
+        expect(menuItem).toBeTruthy();
+        spyOn(component, 'onButtonClick');
```
donna-marijne, line 402: We have an `activate` output as well, we should assert that that does not get called when the menu item is disabled. You could just use the first menu item in that case because it already has an activate handler.

src/components/menu/menu.spec.ts
```diff
@@ -376,6 +382,32 @@ describe('MenuComponent', () => {
...
+
+        // perform a click
+        menuItem.click();
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        expect(component.onButtonClick).not.toHaveBeenCalledWith();
```
donna-marijne, line 409: I believe this should be `toHaveBeenCalled`.

src/components/menu/menu.spec.ts
```diff
@@ -42,10 +46,12 @@ import { MenuModule } from './menu.module';
     `
 })
 export class MenuTestComponent {
+    disabled: boolean = false;
     @ViewChild('menuTrigger', { static: true }) trigger: MenuTriggerDirective;
     @ViewChild('subMenuTrigger', { static: true }) subMenuTrigger: MenuTriggerDirective;
 
-    item1Activated(_: MouseEvent | KeyboardEvent): void {}
+    item1Activated(_: MouseEvent | KeyboardEvent): void { }
+    onButtonClick(): void { }
```
donna-marijne, line 54: Should be `item2Click` to be consistent with the above.

donna-marijne (changes requested):

src/components/menu/menu.spec.ts
cmallon367: Should it not be .not.toHaveBeenCalledWith(); because we don't want the event to be called when disabled is true or am I misunderstanding

src/components/menu/menu.spec.ts
cmallon367: Would it be better to have them as two separate tests or just include them in the one test?

src/components/menu/menu.spec.ts
donna-marijne: `toHaveBeenCalledWith` verifies that a function has been called with the provided arguments, e.g. `.toHaveBeenCalledWith(2)` would check that a function has been called with an argument of 2. You're not checking any arguments here, you just want to check that it has not been called at all with any kind of argument.

src/components/menu/menu.spec.ts
donna-marijne: It's probably fine to check both in the same test, it's just one extra check.

src/components/menu/menu.spec.ts
```diff
@@ -376,6 +383,38 @@ describe('MenuComponent', () => {
...
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        // get the child submenu item
+        const menuItem = document.querySelector<HTMLButtonElement>(
+            '#menu-item-disabled'
+        );
+        expect(menuItem).toBeTruthy();
+
+        const item1Element = overlayContainerElement.querySelector('#menu-item-1') as HTMLButtonElement;
```
ashley-hunter, line 403: Perhaps we could name this something like `itemWithSubMenu` or something like this to indicate the difference between the two?

src/components/menu/menu.spec.ts
```diff
@@ -42,10 +47,12 @@ import { MenuModule } from './menu.module';
     `
 })
 export class MenuTestComponent {
+    disabled: boolean = false;
     @ViewChild('menuTrigger', { static: true }) trigger: MenuTriggerDirective;
     @ViewChild('subMenuTrigger', { static: true }) subMenuTrigger: MenuTriggerDirective;
 
-    item1Activated(_: MouseEvent | KeyboardEvent): void {}
+    item1Activated(_: MouseEvent | KeyboardEvent): void { }
```
ashley-hunter, line 54: Perhaps we could just name these functions `onActivate` and `onClick`, that way if we change the tests in future other items in the test can also use these functions, as they aren't really specific to any item.

Also the type can just be `MouseEvent`, a click event, regardless of if the mouse or keyboard is used, will fire a `MouseEvent`

src/components/menu/menu.spec.ts
```diff
@@ -376,6 +383,38 @@ describe('MenuComponent', () => {
 
         expect(component.item1Activated).toHaveBeenCalledTimes(1);
     });
+
+    it('should not emit a click event when disabled is true.', async () => {
```
ashley-hunter, line 387: Perhaps this should also mention that it checks for activate emit too? E.g. should not emit a click or activate event when disabled is true

ashley-hunter (changes requested):
Thanks! Changes look good, just a few suggestions

src/components/menu/menu.spec.ts
donna-marijne: The `activate` event should be `MouseEvent | KeyboardEvent`, this is a custom output which combines click and keydown.

src/components/menu/menu.spec.ts
```diff
@@ -376,6 +383,38 @@ describe('MenuComponent', () => {
...
+        spyOn(component, 'item2Click');
+        spyOn(component, 'item1Activated');
+
+        // perform a click
+        item1Element.click();
+        menuItem.click();
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        expect(component.item2Click).not.toHaveBeenCalled();
```
donna-marijne, line 415: I don't think this test is actually clicking item 2? I presume you mean item1Click?

If you want to test clicking a submenu trigger as well as a regular menu item, that should be a separate test. I think that would be worth doing, you can check that it doesn't open the submenu.

donna-marijne (changes requested):

src/components/menu/menu.spec.ts
```diff
@@ -354,28 +363,82 @@ describe('MenuComponent', () => {
...
+        await fixture.whenStable();
+
+        // open menu
+        component.trigger.openMenu();
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        // get the child submenu item
+        const itemWithSubMenu = overlayContainerElement.querySelector('#menu-item-1') as HTMLButtonElement;
+        expect(itemWithSubMenu).toBeFalsy();
```
ashley-hunter, line 433: Perhaps I'm misunderstanding this just, but why should the top level menu item not be found? Would it not just be the sub menu shouldn't exist? Correct me if I am misunderstanding this

donna-marijne: Also `#menu-item-1` is not the one with a submenu. That is currently `#menu-item-disabled` (which should obviously be renamed).

src/components/menu/menu.spec.ts
```diff
@@ -354,28 +363,82 @@ describe('MenuComponent', () => {
...
+            '#menu-item-disabled'
+        );
+        expect(subMenuItem).toBeTruthy();
+
+        spyOn(component, 'onActivate');
+        spyOn(component, 'onClick');
+
+        // perform a click
+        itemWithSubMenu.click();
+        subMenuItem.click();
```
donna-marijne, line 413: When I suggested that it was okay to combine `click` and `activate` into the same test, I did not mean clicking separate items. You can and should bind both events to the same menu item.

The test should be:
1. Set disabled = true
2. Click the disabled menu item.
3. Verify that neither `activate` nor `click` were emitted.

donna-marijne (changes requested):

src/components/menu/menu.spec.ts
```diff
@@ -13,17 +13,18 @@ import { MenuModule } from './menu.module';
...
+                (click)="onClick()"
                 [disabled]="disabled"
                 (activate)="onActivate($event)">
                 Item One
             </button>
             <ux-menu-divider></ux-menu-divider>
             <button
                 type="button"
-                id="menu-item-disabled"
+                id="sub-menu-item"
```
ashley-hunter, line 24: I think this `id` is misleading, this is an item in the top level menu, not the sub-menu as the `id` suggests. I don't think this id is ever used either?

ashley-hunter (changes requested):

src/components/menu/menu.spec.ts
cmallon367: That's an oversight, it needs removed it was for a previous test that had to be reconfigured

src/components/menu/menu.spec.ts
```diff
@@ -13,12 +13,17 @@ import { MenuModule } from './menu.module';
...
+                (click)="onClick()"
+                [disabled]="disabled"
+                (activate)="onActivate($event)">
                 Item One
             </button>
             <ux-menu-divider></ux-menu-divider>
             <button
+                type="button"
                 uxMenuItem
+                (click)="onClick()"
```
donna-marijne, line 25: This is now unused so it should be removed.

donna-marijne (changes requested):

ashley-hunter (approved):
Looks good to me, thanks

donna-marijne (approved):
Looks good, thanks.

PR #1405, opened 2020-08-26
EL-4083 - Selection Issue

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4083

The uxSelection directive and the uxSelectionItemDirective were both emitting far too often. The following changes have been made:

- Prevented the directives from emitting on init
- Prevented the directives from emitting when an input changes
- Prevented directives from emitting when an event has occurred but the value doesn't change
- Prevent directives from emitting until the output value has stabilized, for example, in row mode when a selection is made we first deselect all items, then select the newly selected item. The newly selected item should not emit the event when we initially deselect it.
- Added logic to ensure clicking on something like a checkbox emits as expected, e.g. when clicking a checkbox the selected input value may change, however the click on the row itself may also try and toggle the selected state, causing it to be toggled twice. I have added logic in there to prevent this double toggling.
- Added tests covering the scenarios

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4083-Selection-Issue

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4083-Selection-Issue~CI/)

### DIFF
--- a/src/directives/selection/selection.directive.ts
+++ b/src/directives/selection/selection.directive.ts
@@ -58,7 +58,7 @@ export class SelectionDirective<T> implements AfterContentInit, OnDestroy {
     private _onDestroy = new Subject<void>();
 
     /** Store the previous selection so we don't emit more than we have to */
-    private _lastSelection: ReadonlyArray<T>;
+    private _lastSelection: ReadonlyArray<T> = [];
 
     constructor(private _selectionService: SelectionService<T>, private _cdRef: ChangeDetectorRef) {
         _selectionService.selection$.pipe(debounceTime(0), takeUntil(this._onDestroy)).subscribe(items => {
--- a/src/directives/selection/selection.directive.ts
+++ b/src/directives/selection/selection.directive.ts
@@ -60,7 +60,7 @@ export class SelectionDirective<T> implements AfterContentInit, OnDestroy {
     private _lastSelection: ReadonlyArray<T>;
 
     constructor(private _selectionService: SelectionService<T>, private _cdRef: ChangeDetectorRef) {
-        _selectionService.selection$.pipe(takeUntil(this._onDestroy), debounceTime(0)).subscribe(items => {
+        _selectionService.selection$.pipe(debounceTime(0), takeUntil(this._onDestroy)).subscribe(items => {
             if (this.isSelectionChanged(items)) {
                 this.uxSelectionChange.emit(items);
             }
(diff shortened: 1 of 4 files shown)


src/directives/selection/selection.directive.ts
```diff
@@ -57,10 +58,10 @@ export class SelectionDirective<T> implements AfterContentInit, OnDestroy {
     private _onDestroy = new Subject<void>();
 
     /** Store the previous selection so we don't emit more than we have to */
-    private _lastSelection: ReadonlyArray<T>;
+    private _lastSelection: ReadonlyArray<T> = [];
 
     constructor(private _selectionService: SelectionService<T>, private _cdRef: ChangeDetectorRef) {
-        _selectionService.selection$.pipe(takeUntil(this._onDestroy), debounceTime(0)).subscribe(items => {
+        _selectionService.selection$.pipe(debounceTime(0), takeUntil(this._onDestroy)).subscribe(items => {
```
DeanHaire, line 64: What's the purpose of `debounceTime(0)`, waiting for 0ms? Is it needed for some reason?

ashley-hunter: Yes, so lets say we are in `row` mode, clicking on an item in row mode will select only that one row. What actually occurs is:
- deselect all rows
- select target row

Both of these actions will emit a new value on the observable. However the first is a transitional state, the other is the final state. There are many cases where this may fire multiple times before arriving on the final state.

The key point here that all these events fire synchronously the same tick of the event loop. 
`debounceTime(0)` will only proceed on a subsequent tick of the event loop, at which point all synchronous events that could have happened will have happened.

So, in short, the subscribe function will only get called after all the events have fired and the state is "stable".

DeanHaire: That makes sense, thanks!

DeanHaire:
Can you create a downstream build for this?

ashley-hunter:
https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-4083-Selection-Issue~CI/1/

DeanHaire (approved):
Looks good, seems to work properly in each scenario I've tested 👍

donna-marijne (approved):
Thank you, looks good.

PR #1409, opened 2020-09-03
EL-4093 - Select Repeated Calls

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4093

- The issue was actually with the underlying infinite scroll directive that performed a check to see if we are scrolled near the bottom, however when the element is not visible, the height is reported a 0 and the scroll position was also reported as 0, which with our previous calculation assumed we were scrolled to the bottom, in which case we should try and load the next page.
- I have corrected the logic to consider an element with no height (e.g. hidden)
- The ticket suggested creating a select test and performing the set of steps to reproduce the issue (e.g keyboard and mouse presses), but the issue is with the infinite scroll directive and is much easier to reproduce and test directly using it which I have chosen to do instead as this actually tests the underlying issue.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4093-Select-Repeated-Calls

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4093-Select-Repeated-Calls~CI/)


DeanHaire (approved):
Seems to fix the issue for me, code looks fine to me too 👍

donna-marijne (changes requested):
Thank you Ashley, this looks like a good fix that will benefit other components as well. Regarding the e2e test, I think we are missing coverage for keyboard operation of many of our components, so it could be helpful to take this opportunity to implement this issue as a test case since we had planned for it anyway.

ashley-hunter:
Added E2E test which performs selection using keyboard

DeanHaire (approved):

donna-marijne (approved):
Looks good, thank you for adding the test!

PR #1410, opened 2020-09-04
EL-4088 - Use currently active theme in plunkers

DeanHaire:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4088

- Load the appropriate stylesheets for the currently active theme within plunker examples
- Include Roboto when using the 2020 theme

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4088-roboto-plunker-themes

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4088-roboto-plunker-themes~CI/)


donna-marijne (changes requested):
This does get the job done, albeit with duplicated code. Perhaps extending the existing strategy pattern would be cleaner?
```ts
    private createPlaygroundStrategy(playground: IPlayground, theme: SiteTheme): PlaygroundStrategy {
        switch (playground.framework) {
            case 'css':
                return new CssPlaygroundStrategy(this._documentationType, this.createThemeStrategy(theme));
            case 'angular':
            default:
                return new AngularPlaygroundStrategy(this._documentationType, this.createThemeStrategy(theme));
        }
    }
```

```ts
return [...stylesheets, ...this._themeStrategy.getGlobalExternalStyles(assetsUrl)];
```

Also consider pulling the SiteThemeService up from the MF repo so that we don't duplicate that code either.

DeanHaire:
I've removed the duplicate code and simplified the implementation as much as I can. I've also lifted up the site theme service to make better use of that.

docs/app/services/playground/strategies/playground-strategy.ts
```diff
@@ -31,7 +34,28 @@ export abstract class PlaygroundStrategy {
 
     /** Get all the external styles that should be loaded in the `head` of the HTML */
     getGlobalExternalStyles(assetsUrl: string): string[] {
-        return [];
+        const stylesheets = [
```
donna-marijne, line 37: I don't think this should be added to the base abstract class. I think you can extend the strategy pattern as described previously to keep it clean. Let me know what was unclear and I can try to explain further.

donna-marijne (changes requested):

docs/app/app.component.ts
```diff
@@ -89,8 +90,10 @@ export class AppComponent implements OnInit {
     constructor(private router: Router,
         private activatedRoute: ActivatedRoute,
         private navigation: NavigationService,
-        ngZone: NgZone) {
+        ngZone: NgZone,
+        siteThemeService: SiteThemeService) {
         (<any>window).ngZone = ngZone;
+        siteThemeService.load();
```
donna-marijne, line 96: We don't actually need to load the service in the open source site as far as I'm aware. The service just needs to be here because all the playground code is here too. The MF site has it's own independent AppComponent.

docs/app/services/site-theme/site-theme.service.ts
```diff
@@ -0,0 +1,56 @@
...
+
+    constructor(
+        @Inject(DOCUMENT) private _document: Document,
+        @Inject(DOCUMENTATION_TOKEN) private _documentationType: DocumentationType,
+        private _persistentDataService: PersistentDataService
+    ) { }
+
+    load(): void {
+
+        if (this._documentationType !== DocumentationType.Keppel) {
```
donna-marijne, line 23: Can be removed if you avoid calling this function in the open source site.

docs/app/services/playground/strategies/themes/microfocus-2020-strategy.ts
```diff
@@ -0,0 +1,13 @@
+import { ThemeStrategy } from './theme-strategy';
+import { SystemJSHelper } from '../../utilities/system-helper';
+
+export class MicroFocus2020ThemeStrategy implements ThemeStrategy {
+    getStylesheets(assetsUrl: string): string[] {
+        return [
+            SystemJSHelper.getPackageUrl({ name: 'bootstrap', path: 'bootstrap@3.3.7/dist/css/bootstrap.min.css' }),
+            `${assetsUrl}/styles/ux-aspects.css`,
```
donna-marijne, line 8: Since you've added a KeppelThemeStrategy, you could extend that for the three Micro Focus themes and eliminate this bit of duplication as well.

donna-marijne (changes requested):
It looks good and works well, there's just a bit of cleanup.

docs/app/services/playground/strategies/themes/microfocus-2020-strategy.ts
DeanHaire: I don't think I can, the `ux-aspects.css` stylesheet is in different locations. i.e. `/css/ux-aspects.css` for keppel and `/styles/ux-aspects.css` for all micro focus themes

donna-marijne: Good point, don't worry about it then.

donna-marijne (approved):
Looks good, thank you.

PR #1413, opened 2020-09-09
EL-4085 - Row Selection

ashley-hunter:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4085

- Previously the focus state was shown based on whether or not the item that could be tabbed to, rather than if the item was actually focused.
- It is now added when the item is focused and removed when the item is blurred.
- Added unit tests

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4085-Row-Selection

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4085-Row-Selection~CI/)


donna-marijne (changes requested):
Thank you for this, change looks good. Looking at the issue though, I think there is a bit more to this since if you click a table row to select it, it still has a distinct color. Normally we only apply focus styling change to components when they gain focus via the keyboard. Maybe we should check with Roland, but I think that was the intent of raising this issue.

ashley-hunter:
Ah, ok, I thought the issue was that when the row loses focus the focus styling still remained. I will check with Roland.

ashley-hunter:
Ok.. for some reason I hit Update from base branch, Github errors out and merges this PR?! So I will raise another PR once I hear the response from Roland

PR #1415, opened 2020-09-10
EL-4064 - Theme selector

donna-marijne:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4064

* Add theme selector for use in the MF site.
* Rename and reorganize some types.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4064-theme-switch

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4064-theme-switch~CI/)


docs/app/components/navigation-bar/navigation-bar.component.html
```diff
@@ -14,6 +14,8 @@
 
         <div class="navigation-bar-spacer"></div>
 
+        <uxd-theme-selector *ngIf="themes" title="Theme" [themes]="themes"></uxd-theme-selector>
```
ashley-hunter, line 17: We might want to change the Input to something other than `title` as this will result in a browser tooltip when hovering on the theme switcher.

ashley-hunter (changes requested):
Looks good! Seems to work very well! Just noticed one small issue with the name of an input

docs/app/components/navigation-bar/navigation-bar.component.html
donna-marijne: Yes, good point, thank you!

docs/app/components/theme-selector/theme-selector.component.less
```diff
@@ -0,0 +1,17 @@
+@import "../../assets/styles/variables.less";
+
+.theme-selector-button {
+    height: 50px;
+    font-size: 0.9375rem;
+    border: none;
+    padding: 15px;
+    background-color: transparent;
+
+    ux-icon[name="chevron-down"] {
```
DeanHaire, line 10: Does it need to be this specific? If we were to to replace this with a different icon we would need to update this too. Would just `ux-icon { ... }` be ok instead here?

docs/app/interfaces/SiteTheme.ts
```diff
@@ -0,0 +1,11 @@
+export interface SiteTheme {
+    title: string;
+    id: SiteThemeId;
+}
+
+export enum SiteThemeId {
+    Keppel = 'Keppel',
+    MicroFocus = 'MicroFocus',
+    MicroFocus2020 = 'MicroFocus2020',
```
DeanHaire, line 9: Is it necessary to have both SiteTheme and SiteThemeId? I assume the only place you're using it is within the dropdown to display the theme name. If you were to change `MicroFocus2020` to `MicroFocus 2020` as the string value this could then be displayed in the dropdown and would potentially remove the need for both Theme and ThemeId?

donna-marijne: It's not a good idea to use ID or key values as display text since you would then have to make unnecessary changes to the application in order to modify the display text (and also invalidate any prior value for the setting).

DeanHaire (approved):

ashley-hunter (approved):
All good, thanks

PR #1427, opened 2020-10-05
EL-4086 - Masthead Breadcrumbs: remove unwanted href

DeanHaire:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4086

- Only add `href` when `routerLink` has been specified
- Add karma test for checking presence of `href`

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4086-masthead-breadcrumbs

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4086-masthead-breadcrumbs~CI/)

### DIFF
--- a/e2e/pages/app/page-header/page-header.testpage.component.html
+++ b/e2e/pages/app/page-header/page-header.testpage.component.html
@@ -12,7 +12,7 @@
         <router-outlet></router-outlet>
     </div>
 
-    <br />
+    <br>
 
     <button id="button1" class="btn button-primary" (click)="condensed = !condensed">Toggle Condensed</button>
 
@@ -30,7 +30,7 @@
         [secondaryNavigationAutoselect]="autoselect">
     </ux-page-header>
 
-    <br />
+    <br>
 
     <button id="align-left" class="btn button-primary" (click)="alignment = 'left'">Align Left</button>
     <button id="align-center" class="btn button-primary" (click)="alignment = 'center'">Align Center</button>
--- a/e2e/pages/app/page-header/page-header.testpage.component.html
+++ b/e2e/pages/app/page-header/page-header.testpage.component.html
@@ -1,25 +1,42 @@
 <div id="root">
+    <ux-page-header
+        id="pageHeader1"
+        header="My Page"
+        subheader="My Subheader"
+        [crumbs]="crumbs"
+        [items]="items"
+        [condensed]="condensed"
+        [iconMenus]="iconMenus">
+    </ux-page-header>
+    <div class="breadcrumb-content">
+        <router-outlet></router-outlet>
+    </div>
 
-<ux-page-header id="pageHeader1" header="My Page" subheader="My Subheader" [crumbs]="crumbs" [items]="items"
-                [condensed]="condensed" [iconMenus]="iconMenus"></ux-page-header>
+    <br />
 
-<br>
+    <button id="button1" class="btn button-primary" (click)="condensed = !condensed">Toggle Condensed</button>
 
-<button id="button1" class="btn button-primary" (click)="condensed = !condensed">Toggle Condensed</button>
+    <ux-page-header
+        class="m-t-xl m-b-xl"
+        id="pageHeader2"
+        header="My Page"
+        subheader="My Subheader"
+        [crumbs]="crumbs"
+        [items]="items2"
+        [condensed]="condensed"
+        [iconMenus]="iconMenus"
+        [secondaryNavigation]="true"
+        [secondaryNavigationAlignment]="alignment"
+        [secondaryNavigationAutoselect]="autoselect">
+    </ux-page-header>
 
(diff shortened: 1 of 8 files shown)


src/components/page-header/page-header.component.spec.ts
```diff
@@ -87,4 +100,16 @@ describe('Page Header Component', () => {
         expect(pageHeader.textContent).toContain('UX');
     });
 
+    it('should add href to breadcrumbs with a routerLink', () => {
+        const homeBreadcrumb = nativeElement.querySelector('.breadcrumb a');
+        expect(homeBreadcrumb.textContent.trim()).toBe('Home');
+        expect(homeBreadcrumb.hasAttribute('href')).toBe(true);
+    });
+
+    it('should not add href to breadcrumbs without a routerLink', () => {
```
DeanHaire, line 109: The acceptance mentions a new test page, but I think these karma tests achieve the same result and are probably faster than an e2e test anyway. So I'll stick with these unless there's a specific need for an e2e test that I'm not aware of yet

DeanHaire, line 109: The acceptance criteria mentions a new test page, but I think these karma tests achieve the same result and are probably faster, so I'll stick with this unless there's a specific need for an e2e test that I'm not aware of yet

cmallon367: I think that having both test coverage could be a good idea because while I do think the tests present are good I do see the use case for having a new test page currently to my knowledge we do not have any e2e tests that mention the routerLink so it would be good to have some coverage for this.

DeanHaire: I'm not sure if duplicating the karma tests into e2e tests provides much value though. The karma tests cover the fixed behaviour and duplicating tests would possibly be a bit overkill I think

cmallon367: That's true it might be overkill for the bug but if it is in the acceptance criteria I think it should be implemented

donna-marijne: It's not specifically useful for this fix, but it would be good coverage infill to verify that routerlinks work as expected when specified, which I don't think we can cover with Karma.

DeanHaire: @donna-marijne what would be the best way of verifying the routerLinks work? In the test above I've used `hasAttribute('href')` to check for the presence of `href`. Would possibly using `getAttribute('href')` and asserting on the value of the `href` be sufficient without needing an e2e test?

donna-marijne: I don't know what the href would look like within a karma test, or whether we could depend on that value through updates to Angular or Karma. Since the router is an external component we don't really want to be unit testing it. What we are interested in is the outcome i.e. that the expected content is rendered in the router outlet after clicking an item with a routerlink, or that nothing happens if a routerlink is not provided.

DeanHaire: I've added some e2e tests to cover the navigation 👍

e2e/pages/app/page-header/page-header.testpage.component.html
```diff
@@ -1,25 +1,42 @@
...
+        [crumbs]="crumbs"
+        [items]="items"
+        [condensed]="condensed"
+        [iconMenus]="iconMenus">
+    </ux-page-header>
+    <div class="breadcrumb-content">
+        <router-outlet></router-outlet>
+    </div>
+
+    <br />
```
cmallon367, line 15: Can you remove the extra space here and on line 33, I'm assuming its suppose to be <br>

e2e/tests/components/page-header/page-header.e2e-spec.ts
```diff
@@ -268,4 +269,24 @@ describe('Page Header Tests', () => {
...
 
     });
+
+    it('should add a href to breadcrumbs that specify a routerLink', async () => {
+        expect(await page.getBreadcrumbHref(0)).toContain('#/page-header/home');
+    });
+
+    it('should not add a href to breadcrumbs that do not specify a routerLink', async () => {
+        expect(await page.getBreadcrumbHref(1)).toBeNull();
+    });
```
donna-marijne, line 279: These two tests are duplicates of the karma tests, so we could leave them out. The goal of e2e tests is to assert the end user's experience of a feature, e.g. "I click on x and y happens."

donna-marijne (changes requested):
Thanks for the fix and the tests. You may be unaware but there are masthead tests in the ux-aspects-micro-focus repository which already have routing components and tests set up, so it may be more efficient to add the tests there.

donna-marijne (approved):
Looks good to me, thank you!

PR #1429, opened 2020-10-07
feat(ui): phase 1 initial support for RTL

Deklin:
#### Checklist

* [ X] The contents of this PR are mine to submit. No third party code or other material is being committed.
* [ ] New third party dependencies (if any) are linked via `package.json`. Third party dependencies are licensed under one of the following: Apache, MIT, BSD, Mozilla Public License, Eclipse Public License, or Oracle Binary Code License.
* [ ] The code in this PR does not contain export-restricted encryption algorithms.

Adding support for RTL for the following components
TODO - Update w/ component list.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_


docs/index.html
```diff
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" dir="rtl">
```
Deklin, line 2: temporary added this for testing.

Deklin:
Some notes:
1. I am not sure we want to always force on RTL, but instead provide as separate RTL style sheet, but curious thought on this?
2. I will likely not have time to do every ux-aspects component, while a lot are CSS changes, some are code changes too which would require a new input to the component rtl=true (like material does) to flip drag positions etc.
3. Looking for general comments and approach for now on this? (do not merge btw as index.html needs to reverT)

Deklin:
@donna-marijne

donna-marijne:
Thank you Steve, it is great to get this contribution and finally make some progress on RTL! I had a quick talk with Chris this morning, who I think has been more involved with this while I was on vacation.

Regarding #1, I believe a separate or add-on stylesheet would be preferable to an all-in-one solution. As you can see from reviewing the source, we are in a bad place regarding the size of our stylesheets because we were not able to leverage component-encapsulated styles alongside our theming approach. If we can deliver RTL without contributing further to that problem, that would be ideal.

For a productized solution, ideally I would like to see a post-processing solution such as RTLCSS handle most of these changes. I understand that you have looked into this so I would be interested in hearing your findings. The reason for preferring this over separate stylesheet rules is primarily how maintainable it will be in the future. If someone is making a style change they will need to be aware of this separate style block that will also need to be updated. If it's handled automatically or via inline directive comments, it's much less likely to be overlooked.

I understand that there will probably need to be some extra style rules specific to RTL support, but I am hoping that such things which cannot be handled by post-processing at all will be a minority of the required changes.

Deklin:
@donna-marijne reach out to chris. originally we were going to do the post processing rtlcss solution but he favored specific implementation. I can go either way. i've spent the time to get rtlcss working in our angular build environment and it works very well. However it won't solve all the problems, some of the components will definitely need updates.

1. resizable table will need code updates and a rtl @input option to enable rtl mode due to left right drag.
2. the picker component needed to swap icons based on RTL so they icons match the direction of where selections will go.
3. I am happy to do an automation update which creates a separate style sheet specific to RTL and can abandon this PR in favor of that, but as stated in #1, code updates are still needed.

donna-marijne:
Great, I definitely understand the need for exceptional CSS and code so that is no problem. There is a script in scripts/inline-less.js where the postprocessor could easily plug in. I also see that there is a [webpack plugin](https://www.npmjs.com/package/rtlcss-webpack-plugin) so we could build it into the docs site for testing as well.

Regarding the code changes that are required, what do you think about using the [CDK bidi API](https://material.angular.io/cdk/bidi/api) to detect the appropriate directionality?

Deklin:
please refer to othe rpr.

PR #1430, opened 2020-10-09
EL-4060 - Update chart.js and ng2-charts

donna-marijne:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4060

* Update chart.js and ng2-charts dependencies to match those used in the doc site.
* Move the dependencies to peerDependencies.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4060-charts-dependency

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4060-charts-dependency~CI/)


src/package.json
```diff
@@ -74,6 +71,9 @@
     "@angular/core": ">=8.0.0 <11.0.0",
     "@angular/forms": ">=8.0.0 <11.0.0",
     "@angular/upgrade": ">=8.0.0 <11.0.0",
-    "angular-tree-component": ">=5.2.0"
+    "@types/chart.js": ">=2.9.3",
```
ashley-hunter, line 74: If I try and install the tgz for this build in a project I get the following error when I try to build:

`ERROR in node_modules/@ux-aspects/ux-aspects/plugins/chartjs/timeline/timeline-chart.module.d.ts(123,27): error TS2503: Cannot find namespace 'Chart'.`

Installing the chart.js and `@types/chart.js` (ng2-charts is fine as we don't depend on it within the library) modules manually fixes the issue as `TimelineChartModule` directly relies on those dependencies. I realise that is the purpose of peer dependencies, you install them yourself with the version you require, but perhaps this would be considered a breaking change and should maybe be done as part of a 3.0 release and perhaps have update schematics? Or is it fine to just add this to the release notes?

If we are happy to go ahead with this we should probably update the getting started page and ng add schematics.

ashley-hunter (changes requested):

src/package.json
donna-marijne: Thank you, yes you are right, I did not intend to make a breaking change so I have moved those two back to dependencies.

ashley-hunter (approved):
Looks great! Thank you!

PR #1431, opened 2020-10-12
EL-4117 - basic-input-controls

cmallon367:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4117

Added a disabled input & applied some disabled styling to the input

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4117-basic-input-controls

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4117-basic-input-controls~CI/)


src/components/slider/slider.component.ts
```diff
@@ -220,36 +228,39 @@ export class SliderComponent implements OnInit, AfterViewInit, DoCheck {
 
     thumbEvent(thumb: SliderThumb, event: SliderThumbEvent): void {
 
-        // get the current thumb state
-        const state = this.getThumbState(thumb);
+        if (!this.disabled) {
```
donna-marijne, line 231: Usually with disabled-type properties, we do:

```ts
if (this.disabled) {
  return;
}
```

at the top of the function. This helps reduce nesting, which can be problematic especially in large functions like this.

src/components/slider/slider.component.ts
```diff
@@ -91,7 +94,7 @@ export class SliderComponent implements OnInit, AfterViewInit, DoCheck {
     ticks: SliderTick[] = [];
     defaultOptions: SliderOptions;
 
-    constructor(colorService: ColorService, private _changeDetectorRef: ChangeDetectorRef) {
+    constructor(colorService: ColorService, private _changeDetectorRef: ChangeDetectorRef, private _changeDetector: ChangeDetectorRef) {
```
DeanHaire, line 97: Why is the change detector injected here twice?

src/components/slider/slider.component.html
```diff
@@ -1,11 +1,12 @@
 <div class="track"
      #track
+     [class.disabled]="disabled"
```
donna-marijne, line 3: This should not be needed since the parent element has a disabled class as well, you can use that in the selector instead.

donna-marijne (changes requested):
Thanks Christina, looks like a good start. Bear in mind the usual acceptance criteria when enhancing any existing component.

1. The new property should be documented.
2. We will need test coverage of the new property.
3. When disabled, the slider thumb still has a "pointer" cursor, implying that interaction is permitted. It should be the default cursor.
4. When disabled, the slider thumb should not be able to receive focus.
5. The slider thumb should probably be grayed out in the same way as the track.

https://plnkr.co/edit/hkebVQjLwyn3CHCM

donna-marijne (changes requested):

donna-marijne (approved):

DeanHaire (approved):

PR #1435, opened 2020-10-14
EL-4108 - Theme-specific documentation examples

donna-marijne:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4108

* Add `theme` property to `BaseDocumentationSection` which allows `ngSwitch` to be used on the main documentation example.
* Update `BaseDocumentationSection` to use theme-specific snippets. Snippets with the active theme ID in their filename will be used in preference to the default snippet. E.g. `app.microfocus2020.html` will be mapped to `appHtml` in the snippets collection while the MicroFocus2020 theme is active.
* Update `IPlaygroundProvider` to optionally support a function instead of an `IPlayground` object. This is needed to support the case when the active theme changes after the documentation section has loaded. When the theme changes, the function will be called to get an updated playground object containing the theme-specific code snippets.
* Update the button documentation section to test the theme switching features. The 2020 snippets feature the new Neutral button, and exclude the deprecated Accent button.
* Update the checkbox section so that it can be shared with the MF site. This verifies a secondary use case for this feature.

Proposed procedure for adding a theme-specific example to a documentation section:
1. Ensure that the parent module imports `CommonModule`.
2. Wrap the existing example code:
```html
<ng-container [ngSwitch]="theme">
    <div *ngSwitchDefault ...>
        Existing example content
    </div>
</ng-container>
```
3. Insert the themed content underneath the default with `*ngSwitchCase="SiteThemeId.MicroFocus2020"`.
4. Inside the `snippets` subdirectory, copy the existing HTML snippet (and any others that need to change) and rename it to include `.microfocus2020` in the filename, e.g. `app.html` -> `app.microfocus2020.html`.
5. Convert the `playground` property into a function which returns the playground object:
```ts
    playground = () => {
        return {
            files: {
                'app.component.html': this.snippets.raw.appHtml,
                'app.component.ts': this.snippets.raw.appTs,
            },
        };
    }
```

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4108-themed-examples

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4108-themed-examples~CI/)


docs/app/components/component-section/component-section.component.ts
```diff
@@ -31,27 +36,58 @@ export class ComponentSectionComponent implements OnInit {
...
+        if (!component) {
             console.warn(`ComponentSectionComponent: ${this.componentName} cannot be resolved - decorate component with @DocumentationSectionComponent.`);
+            return;
         }
 
+        const factory = this._resolverService.resolveComponentFactory(component);
+        const componentRef = this.viewContainer.createComponent<T>(factory);
+        this._documentationSection = componentRef.instance;
+
+        this.updateWithTheme(this._siteThemeService.theme$.getValue());
```
DeanHaire, line 60: I'm not sure if `updateWithTheme` clearly describes the purpose of this. From what I can see so far, it looks like it switches the theme value in `BaseDocumentationSection` and updates the snippets. If in future I were to arrive in this file and read this line of code, the function name doesn't really indicate its purpose and I'll have to read the actual content of the function. Ideally we should be able to clearly understand the purpose of a function just from its name. Could `updateWithTheme` be renamed to possibly better reflect that it's switching to the theme specific snippets?

DeanHaire (changes requested):
Minor suggestion to hopefully improve the readability in this area

docs/app/components/component-section/component-section.component.ts
donna-marijne: It's a function which is called when the theme changes, I'm sure sure what else to call it. `onThemeChange`?

DeanHaire: I'm not sure `onThemeChange` would work either. Since from the perspective of `BaseDocumentationSection` it's actually updating the theme. I think something along the lines of `useSnippetsForTheme(theme)` or `replaceSnippetsForTheme(theme)`. With those names (or something similar you might be able to think of), you have the advantage of being able to know what the purpose of the function is without having to open the function and read/understand its contents. Currently, `updateWithTheme` gives no indication that the snippets are being replaced with their theme specific alternatives.

donna-marijne: Compare it with ngOnInit for example. If we need to make it do something in addition to changing the snippets, `useSnippetsForTheme` would become misleading.

DeanHaire: If we were doing something else other than just updating the snippets then I agree `onThemeChange` would be fine. But since currently all we're doing is replacing the contents of the snippets then I think we could take advantage of the simplicity and readability of having a function named something like `useSnippetsForTheme` (or if you can think of anything better). If we need to add more to that function in the future then we could always rename it to better reflect its purpose at that point.

donna-marijne: Right now it does do multiple things:
```ts
    private updateWithTheme(theme: SiteThemeId): void {
        // Some sections without snippets don't extend BaseDocumentationSection, ignore those
        if (isBaseDocumentationSection(this._documentationSection)) {
            this._documentationSection.updateWithTheme(theme);
            this.updatePlayground();
        }
    }
```

DeanHaire: My bad, I probably could've put this comment on a different line to be more clear. I meant the `updateWithTheme` function in the `BaseDocumentationSection` class which is called on line 79 of this file. Although now that I notice it, it probably isn't good to have two functions both called `updateWithTheme`, with one of them calling the other.

The private function in this component makes sense named as `updateWithTheme` (possibly without the 'with'). A function which is updating a theme in this scenario seems to have two actions:

1. Updating the snippets (via line 79, which with the current naming doesn't indicate that this is happening)
1. Updating the playground

By renaming the `BaseDocumentationSection`'s `updateWithTheme` function, then when you read the private function in this class you'd be able to see these two actions without having to open and read through the functions.

donna-marijne: Thanks, yes they should not share the same name, this is a good point.

The function in `BaseDocumentationSection` also does two separate things, and if a documentation section overrides it to hook into the event, possibly more.

I've done some renaming which should hopefully make the purpose clear.

DeanHaire: Thanks, I'm happy with the updated naming. Definitely seems clearer and easier to follow now 👍

donna-marijne: Great, thank you for the feedback!

DeanHaire (dismissed):
Thanks for the changes, looks good

Sdhilipkumar (changes requested):

docs/app/components/base-documentation-section/base-documentation-section.ts
```diff
@@ -1,34 +1,65 @@
-import { ISnippets } from '../../interfaces/ISnippets';
+import { ISnippets, SnippetCollection } from '../../interfaces/ISnippets';
+import { SiteThemeId } from '../../interfaces/SiteTheme';
 
 export abstract class BaseDocumentationSection {
 
     snippets: ISnippets;
+    theme: SiteThemeId = SiteThemeId.Keppel;
```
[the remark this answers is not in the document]
donna-marijne: The names are the strings which you can see in the dropdown in the top navigation bar. They are separate so that we can adjust those strings (which we have already had to do) without affecting the functionality of the application.

I'm not aware of any convention that says ID values have to be numbers.

Sdhilipkumar (approved):
Looks good.

PR #1444, opened 2020-10-30
EL-4157-Input-Dropdown:input field unexpectedly getting focus when opened programmatically

cmallon367:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4157

The fix is to remove filter focus from the dropdown menu open so that it no longer receives focus when opened programmatically  
#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4157-Input-Dropdown-input-field

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4157-Input-Dropdown-input-field~CI/)


DeanHaire:
I'm wondering would it be possible to test this via an e2e test? Possibly by using `await browser.driver.switchTo().activeElement()` to get the currently focused element you could assert that this element isn't the input field?

Also, could you update the title of the PR to better describe the issue so that if someone is looking through the commits they can get an understanding of the purpose of this PR without needing to open and read through it.

cmallon367:
I'm assuming the reason we didn't use the e2e tests is because there wasn't an issue on our end it became apparent whenever the user was running cypress

donna-marijne:
If it is possible to add test coverage, you definitely should.

DeanHaire:
> I'm assuming the reason we didn't use the e2e tests is because there wasn't an issue on our end it became apparent whenever the user was running cypress

Yes, but it'd be good to cover this case if we can to prevent a regression in future. There's a couple of examples of using the approach I suggested such as `hasFocus` in `column-picker.po.spec.ts` and in the [sliders tests](https://github.com/UXAspects/UXAspects/blob/a14387516089e28b647f3bd451051062810b6fbb/e2e/tests/components/sliders/sliders.e2e-spec.ts#L373) if that's helpful

DeanHaire (approved):

tobias-schreiner:
There are two issues I can see with this approach:

1. The filter field is never focused when the drop-down is opened. This degrades the usability of the component for interactive use.

2. Using only the keyboard, I cannot access the filter field and also the content of the drop-down anymore. So the accessibility of the component is broken.

I have to admit, though, that I have no access to your ticketing system, so I don't know the background for this change.

andyreidz:
image

donna-marijne:
So it seems like `this._focusFilter` should still be called in every case except when the menu has been opened by setting `dropdownOpen` to true. Checking the value of `dropdownOpen` at the start of `onMenuOpen` might do the trick.

src/components/input-dropdown/input-dropdown.component.spec.ts
```diff
@@ -194,6 +194,17 @@ describe('InputDropdownComponent', () => {
...
 
+    it('should not focus the input field when dropdownOpen is programmatically set to true', async () => {
+        component.dropdownOpen = true;
+
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        const input = document.querySelector<HTMLInputElement>('input.form-control');
+
+        expect(document.activeElement).not.toBe(input);
```
donna-marijne, line 205: I reverted your fix below and ran this test and it still passed. Could you look into another way to verify this, and check that it is effective?

src/components/input-dropdown/input-dropdown.component.ts
```diff
@@ -162,7 +161,6 @@ export class InputDropdownComponent<T> implements ControlValueAccessor, AfterVie
     inputFocusHandler(): void {
         if (!this.dropdownOpen) {
             this.dropdownOpen = true;
-            this.dropdownOpenChange.emit(this.dropdownOpen);
```
donna-marijne, line 165: This is also called by a `focus` handler in the template, so I don't think this is the right fix. You might be able to move `this._focusFilter()` down a couple of lines to solve this problem.

donna-marijne (changes requested):
Following Tobias' feedback, please also set up test coverage to make sure that the input field is correctly focused when the dropdown is opened via click.

src/components/input-dropdown/input-dropdown.component.spec.ts
donna-marijne: Ignore this, I made a mistake.

src/components/input-dropdown/input-dropdown.component.spec.ts
```diff
@@ -46,6 +46,19 @@ describe('InputDropdownComponent', () => {
...
+        spyOn(component.filterInputElement.nativeElement, 'focus');
+
+        let trigger = document.querySelector('button.form-control') as HTMLButtonElement;
+
+        trigger.click();
+
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        expect(component.filterInputElement.nativeElement.focus).toHaveBeenCalled();
```
donna-marijne, line 59: This should be `expect(document.activeElement).toBe(input);` I think (the opposite of your check in the other test).

src/components/input-dropdown/input-dropdown.component.spec.ts
```diff
@@ -46,6 +46,19 @@ describe('InputDropdownComponent', () => {
         expect(component.filterInputElement.nativeElement.focus).toHaveBeenCalledWith();
     });
 
+    it('should focus the input field when dropdownOpen is opened with a click', async () => {
+        spyOn(component.filterInputElement.nativeElement, 'focus');
+
+        let trigger = document.querySelector('button.form-control') as HTMLButtonElement;
```
donna-marijne, line 52: Should be `const`.

src/components/input-dropdown/input-dropdown.component.ts
```diff
@@ -138,10 +138,9 @@ export class InputDropdownComponent<T> implements ControlValueAccessor, AfterVie
     }
 
     onMenuOpen(): void {
-        this._focusFilter();
-
         if (this.dropdownOpen !== true) {
             this.dropdownOpen = true;
+            this._focusFilter();
```
donna-marijne, line 143: Move this below the next line as well, I think.

donna-marijne (changes requested):

DeanHaire (approved):

donna-marijne (approved):
Looks good, thank you.

PR #1447, opened 2020-11-04
EL-4145 - Masthead allow navigation items to open in a new window/tab

Sdhilipkumar:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4145

Replace a button with anchor tag to enable options like open in new window/tab

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4145-masthead-allow-navigation

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4145-masthead-allow-navigation~CI/)


DeanHaire:
Should this fix include secondary navigation items? Right clicking on one of the items under 'Showcase' doesn't show an open in new tab option (and navigates on right click too before I'd be given a chance to open in a new tab if it were there)?

Sdhilipkumar:
> Should this fix include secondary navigation items? Right clicking on one of the items under 'Showcase' doesn't show an open in new tab option (and navigates on right click too before I'd be given a chance to open in a new tab if it were there)?

Secondary navigation items are already anchor tags. So I have not done any change on that.whereas the primary navigations items are buttons earlier, now we made it as anchor tags.

src/components/page-header/navigation/navigation-item/navigation-item.component.html
```diff
@@ -74,4 +72,36 @@
...
+        </i>
+
+        <ux-icon *ngIf="_iconType === 'component'"
+                 class="navigation-item-icon"
+                 [name]="_item.icon">
+        </ux-icon>
+
+    </ng-container>
+
+    <span class="navigation-item-label">{{ _item?.title }}</span>
```
donna-marijne, line 104: Instead of repeating this content, could you refactor it out into a template?

src/components/page-header/navigation/navigation-item/navigation-item.component.less
```diff
@@ -1,5 +1,5 @@
 ux-page-header-horizontal-navigation-item {
-    .horizontal-navigation-button {
+    .horizontal-navigation-link {
```
donna-marijne, line 2: I don't think this needed to be renamed, might be simpler just to leave it.

src/components/tabset/tabset.component.html
```diff
@@ -15,7 +15,28 @@
...
 
-        <a class="nav-link"
+       <ng-container *ngIf="tab.routerPath">
+            <a class="nav-link"
+            [id]="tab.id"
+            role="tab"
+            uxTabbableListItem
+            uxFocusIndicator
+            (mousedown)="_tabset.select(tab)"
+            (activated)="_tabset.select(tab)"
```
donna-marijne, line 25: If there's a routerLink attached, I don't think we do any of this. I think it should just be the link.

src/components/tabset/tabset.component.html
```diff
@@ -15,7 +15,28 @@
...
+            [attr.aria-controls]="tab.id"
+            [attr.aria-selected]="(_tabset.activeTab$ | async) === tab"
+            [attr.aria-disabled]="tab.disabled"
+            [routerLink]="tab.routerPath">
+
+            <span *ngIf="!tab.headingRef">{{ tab.heading }}</span>
+
+            <ng-container *ngIf="tab.headingRef" [ngTemplateOutlet]="tab.headingRef"></ng-container>
```
donna-marijne, line 33: This should probably be refactored as well, even though it's only two lines at the moment.

docs/app/pages/components/components-sections/tabs/tabs/tabs.component.html
```diff
@@ -87,6 +87,9 @@ <h4>{{ tab.title }}</h4>
     <tr uxd-api-property name="disabled" type="boolean" defaultValue="false">
         If <code>true</code> the tab can not be selected.
     </tr>
+    <tr uxd-api-property name="routerPath" type="string">
+        Define the tab router path
```
donna-marijne, line 91: I think this needs some clarification, at least a link to the Angular documentation for `routerLink` so that the format is understand, but also a description of what the effect of setting `routerPath` on a tab is.

src/components/tabset/tab/tab.component.ts
```diff
@@ -31,6 +31,9 @@ export class TabComponent implements OnInit, OnDestroy {
     /** Define the tab heading */
     @Input() heading: string;
 
+    /** Define the tab router path */
+    @Input() routerPath: string;
```
donna-marijne, line 35: Having thought about it, I would prefer `route` as the name of this input, as long as it doesn't cause any conflicts.

src/components/page-header/page-header.component.html
```diff
@@ -85,6 +85,7 @@ <h1 class="page-header-title">{{header}}</h1>
         <ux-tabset *ngIf="(selectedRoot$ | async)?.children; let children" [manual]="true">
             <ux-tab *ngFor="let child of children"
                 [heading]="child.title"
+                [routerPath]="child.routerLink"
                 [active]="child === (selected$ | async)"
                 (activated)="select(child)"
```
donna-marijne, line 90: I'm a bit concerned that this is going to cause multiple route events, because the select function calls `navigate`.

https://github.com/UXAspects/UXAspects/blob/master/src/components/page-header/page-header.service.ts#L44

Since you've applied the routerLink directive, this call to `navigate` should no longer be called on a user click/keypress. However, if the tab is activated via the autoselect behavior, it should still navigate.

donna-marijne (changes requested):
Thanks Sasi. This does need some test coverage. I think it might be feasible to do via karma, not sure though.

Indentation has been broken in several places, please check and fix.

src/components/page-header/page-header.component.html
donna-marijne: @ashley-hunter I'd like to get your opinion on this if you get the chance.

ashley-hunter: Yeah this is a good point. 

I was thinking perhaps we could check if the routerLink is active and then skip navigation if that was the case, but I don't even think that would work as there are cases like lazy loaded routes and route guards which are asynchronous so navigation may not have occurred by the time we check.

Perhaps we could add an optional parameter to the select function to skip navigation?

E.g.

```typescript
select(item: PageHeaderNavigationItem, navigate: boolean = true): void {
...
    if (item.routerLink && navigate) {

            // Trigger router navigation
            const routerLink = Array.isArray(item.routerLink) ? item.routerLink : [item.routerLink];
            this._router.navigate(routerLink, item.routerExtras);

        }
```

```html
(activated)="select(child, false)"
```

Or if there may not always be a routerLink then:
```html
(activated)="select(child, !!child.routerLink)"
```

donna-marijne: Yes, I think `select(child, false)` should work. The `activated` event is also emitted when the tab changes programmatically, but I think that only happens with the `secondaryNavigationAutoselect` option and there's code to cover that (as far as I can tell).

src/components/page-header/navigation/navigation-item/navigation-item.component.html
```diff
@@ -1,7 +1,6 @@
 <div *ngIf="_item.children && _item.children.length > 0 && !(secondary$ | async)">
 
-    <button #navigationBtn
-        type="button"
+    <a #navigationBtn
```
donna-marijne, line 3: This doesn't have an href so I don't think it needs to change.

src/components/page-header/navigation/navigation-item/navigation-item.component.html
```diff
@@ -45,33 +32,56 @@
...
+    [class.disabled]="_item.disabled"
+    [class.selected]="_item.selected"
+    [attr.aria-selected]="_item.selected"
+    [routerLink]="_item.routerLink">
+
+    <ng-container
+        [ngTemplateOutlet]="allIconTypes">
+    </ng-container>
+
+    </a>
```
donna-marijne, line 51: Fix indentation.

src/components/page-header/navigation/navigation-item/navigation-item.component.html
```diff
@@ -45,33 +32,56 @@
...
-    (click)="select()"
-    [disabled]="_item.disabled">
+    (click)="select()">
+
+    <ng-container
+        [ngTemplateOutlet]="allIconTypes">
+    </ng-container>
 
-    <!-- Support all icon types -->
+    </a>
```
donna-marijne, line 70: Fix indentation.

src/components/page-header/navigation/navigation-item/navigation-item.component.html
```diff
@@ -45,33 +32,56 @@
...
+</ng-container>
+
+<!-- Support all icon types -->
+<ng-template #allIconTypes>
```
donna-marijne, line 74:
```suggestion
<ng-template #navigationItemContent>
```

src/components/page-header/page-header.component.html
```diff
@@ -85,8 +85,9 @@ <h1 class="page-header-title">{{header}}</h1>
         <ux-tabset *ngIf="(selectedRoot$ | async)?.children; let children" [manual]="true">
             <ux-tab *ngFor="let child of children"
                 [heading]="child.title"
+                [route]="child.routerLink"
                 [active]="child === (selected$ | async)"
-                (activated)="select(child)"
+                (activated)="select(child,false)"
```
donna-marijne, line 90:
```suggestion
                (activated)="select(child, false)"
```

src/components/tabset/tabset.component.html
```diff
@@ -15,7 +15,31 @@
...
+            [attr.aria-controls]="tab.id"
+            [attr.aria-selected]="(_tabset.activeTab$ | async) === tab"
+            [attr.aria-disabled]="tab.disabled"
+            [routerLink]="tab.route">
+
+            <ng-container
+                [ngTemplateOutlet]="tabDetails">
+            </ng-container>
+
+            </a>
```
donna-marijne, line 38: Fix indentation

src/components/tabset/tabset.component.html
```diff
@@ -26,13 +50,13 @@
...
 
-            <span *ngIf="!tab.headingRef">{{ tab.heading }}</span>
-
-            <ng-container *ngIf="tab.headingRef" [ngTemplateOutlet]="tab.headingRef"></ng-container>
-        </a>
+            <ng-container
+                [ngTemplateOutlet]="tabDetails">
+            </ng-container>
 
+            </a>
```
donna-marijne, line 57: Fix indentation

src/components/tabset/tabset.module.ts
```diff
@@ -8,7 +9,8 @@ import { TabsetComponent } from './tabset.component';
 @NgModule({
     imports: [
         AccessibilityModule,
-        CommonModule
+        CommonModule,
+        RouterModule.forChild([])
```
donna-marijne, line 13:
```suggestion
        RouterModule
```

donna-marijne (changes requested):
Looks mostly okay, make sure to test keyboard navigation in all circumstances however.

src/components/page-header/navigation/navigation-item/navigation-item.component.html
```diff
@@ -45,33 +34,62 @@
...
 
-    <!-- Support all icon types -->
+        <ng-container
+            [ngTemplateOutlet]="navigationItemContent">
+        </ng-container>
+    </a>
+</ng-container>
+
+<ng-container *ngIf="!_item.routerLink">
+    <a *ngIf="!_item.children || _item.children.length === 0 || (secondary$ | async)"
```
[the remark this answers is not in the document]
Sdhilipkumar: Validated, keyboard navigation is working fine on the MF site.

donna-marijne (approved):
This looks good to me, thank you. Take care not to merge this until the downstream e2e tests are sorted out however. We'll need to verify that there were no visual regressions.

DeanHaire (approved):

PR #1472, opened 2021-01-15
EL-3991 - Resolve-dragula-build-error

cmallon367:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-3991

Importing dragula the same way ng2-dragula is imported 
#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3991-Resolve-dragula-build-error

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-3991-Resolve-dragula-build-error~CI/)


ashley-hunter:
Hmm, been looking into this, it's a tricky one. The only workaround I can see working is to have a script to modify our JS bundles to add `window.global = window;` at the start of them all. Not difficult, but not really ideal.

Ideally the solution would be to move to use the CDK drag and drop functionality instead, but that is a bigger task than this was intended to be.

Apart from that I am not sure what to suggest.

donna-marijne:
If we set `window.global = window`, is it possible that we would break something that another library is depending on?

ashley-hunter:
Well, we could probably make it a little bit smarter to only patch it if it doesn't exist, something like this perhaps:

```javascript
if (window) { // in case running in a node environment e.g. server side rendering
    window.global = window.global || window; // don't override global if it is already defined.
}
```

donna-marijne:
Hmm, [this comment](https://github.com/valor-software/ng2-dragula/issues/849#issuecomment-405736862) states that they were not able to do it as part of the library because it would happen too late. It's not clear what they actually tried however.

This was meant to be small ticket, so I'm inclined to leave it and spend the time on transitioning to the CDK instead.

ashley-hunter:
Yeah, this issue is only regarding a built warning, it doesn't break anything, so perhaps its best to leave it as is for now and address it properly with the CDK at some point.

PR #1477, opened 2021-01-25
EL-4187 - Dashboard - min col and row spans

DeanHaire:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4187

- Add new inputs (`minColSpan` and `minRowSpan`) to allow users to specify the minimum col/row span of a widget

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4187-dashboard-min-colrow

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4187-dashboard-min-colrow~CI/)


src/components/dashboard/dashboard.service.ts
```diff
@@ -436,6 +440,22 @@ export class DashboardService implements OnDestroy {
             dimensions.height = this.options.minHeight;
         }
 
+        const minWidth = this.getColumnWidth() * action.widget.minColSpan - this.getColumnWidth() / 2 + 1;
```
ashley-hunter, line 443: Could we add some parenthesis here and for the `minHeight` calculation too. We are heavily relying on operator precedence and at a glance it is not obvious what order this is running in, e.g.

```suggestion
        const minWidth = ((this.getColumnWidth() * action.widget.minColSpan) - (this.getColumnWidth() / 2)) + 1;
```

src/components/dashboard/widget/dashboard-widget.component.ts
```diff
@@ -299,6 +319,8 @@ export class DashboardWidgetComponent implements OnInit, AfterViewInit, OnDestro
     static ngAcceptInputType_autoPositioning: boolean | string;
     static ngAcceptInputType_col: number | string;
     static ngAcceptInputType_row: number | string;
+    static ngAcceptInputType_minColSpan: number | string;
+    static ngAcceptInputType_minRowSpan: number | string;
```
ashley-hunter, line 323: We can use the type `NumberInput` from the CDK Coercion library as the type for this. We can also now change the others here, there is also `BooleanInput` as well. We couldn't previously use those because they were only in CDK v9+ but now we don't support Angular 8 we should now be using these provided types.

src/components/dashboard/dashboard.service.ts
```diff
@@ -1255,6 +1275,22 @@ export class DashboardService implements OnDestroy {
             dimensions.height = this.getRowHeight();
         }
 
+        const minWidth = this.getColumnWidth() * widget.minColSpan + 1;
+        const widthSmallerThanAllowed = dimensions.width < minWidth;
```
ashley-hunter, line 1279: Just a suggestion, its often good practice to prefix boolean variables with `is` or `has` to indicate the type, e.g. `hasInvalidWidth` or `isBelowMinWidth` etc..

src/components/dashboard/dashboard.service.ts
```diff
@@ -116,7 +116,9 @@ export class DashboardService implements OnDestroy {
                 col: widget.getColumn(),
```
ashley-hunter, line 116: Not particularly sure what the cause of this issue is, but you can see here the widget on the right is not behaving properly:

[Image from Gyazo](https://gyazo.com/7ad3ae45ae36fdfe53541997a7ff20e3)

https://plnkr.co/edit/H4fXOruNQFsPYAHm

ashley-hunter (changes requested):
Thanks, just noticed one issue when resizing and had a few very minor suggestions.

src/components/dashboard/dashboard.service.ts
DeanHaire: I've created a separate ticket to look into this since it's a pre-existing issue

ashley-hunter (approved):
Thanks for the updates, looks good to me.

cmallon367 (approved):

PR #1484, opened 2021-02-05
reorderable table view engine fix

ashley-hunter:
#### Checklist

#### Ticket / Issue

N/A

Previously we were doing:

```typescript
{ provide: new InjectionToken<CdkDropListGroup<unknown>>('CdkDropListGroup') , useValue: undefined },
```
Injection Tokens are unique, it is the instance used to determine if they are a match, the string is just a description and not really used for anything, so this was never actually going to provide the value for CdkDropListGroup anyway. 

The CdkDropListGroup injectable was not used in v9 and is optional in later versions. We don't rely on any features of it so it is not required to be provided at all, so we can leave it out and it works fine.

I also believe pre-ivy there were issues with having an undefined value for a injectable which is where the compilation error was coming from in view engine builds.

(I also fixed another small issue where we were using `private` on a method that had a HostListener. This should be public and also would break in a pre-ivy build if someone was to copy the sample code).

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_reorderable-view-engine-fix


donna-marijne (approved):
This is great, thank you for making the fix. I'm going to make a patch release in case someone bumps into this issue before the next regular release.

PR #1486, opened 2021-02-11
EL-4178 - Popover - hovering div edges create event loop

DeanHaire:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4178

- div with popover directive requires a higher z-index value than the z-index value of `.cdk-overlay-container` to prevent a mouseenter / mouseleave event loop when hovering the the edge that the div and popover share

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4178-popover-loop

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4178-popover-loop~CI/)

### DIFF
--- a/src/components/popover/popover.directive.ts
+++ b/src/components/popover/popover.directive.ts
@@ -9,7 +9,12 @@ import { PopoverComponent } from './popover.component';
 
 @Directive({
     selector: '[uxPopover]',
-    exportAs: 'ux-popover'
+    exportAs: 'ux-popover',
+    host: {
+        // requires a higher z-index than .cdk-overlay-container to prevent a mouseenter/mouseleave loop
+        // when hovering on the edge that the div and popover share
+        '[style.z-index]': '10102'
+    }
 })
 export class PopoverDirective extends TooltipDirective implements OnInit, OnChanges {
 
--- a/src/components/popover/popover.directive.ts
+++ b/src/components/popover/popover.directive.ts
@@ -11,9 +11,8 @@ import { PopoverComponent } from './popover.component';
     selector: '[uxPopover]',
     exportAs: 'ux-popover',
     host: {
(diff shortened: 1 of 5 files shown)


src/components/popover/popover.directive.ts
```diff
@@ -9,7 +9,11 @@ import { PopoverComponent } from './popover.component';
 
 @Directive({
     selector: '[uxPopover]',
-    exportAs: 'ux-popover'
+    exportAs: 'ux-popover',
+    host: {
+        '[class.ux-popover-host]': 'true',
+        '[class.ux-tooltip-host]': 'false'
```
donna-marijne, line 15: Do uxPopover and uxTooltip need to have different host classes? They're both essentially the same thing, you can probably use a common class for both.

donna-marijne (changes requested):
Looks good, maybe it can be simplified a bit more though.

src/components/popover/popover.directive.ts
DeanHaire: No, I just did it this way so if you're inspecting it you won't see a `ux-tooltip-host` on a popover to avoid confusion. But I could switch it back to using `ux-tooltip-host` on both if you think that's simpler? I suppose it wouldn't matter too much since popover extends tooltip

donna-marijne: Yes I think that's fine, or if you have any ideas for a neutral name. It's not a big deal, just saves a couple of lines of code.

donna-marijne (approved):
Looks good, thanks.

PR #1495, opened 2021-03-01
EL-4194 - Dashboard scroll

donna-marijne:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4194

While dragging, find the nearest scrollable element and scroll while the pointer is outside of its bounds. I was not able to find a way to test this in protractor.

The two different scrolling cases (page and element) are demonstrated with these plunkers:
- https://plnkr.co/edit/eFDyOyQZLNMfQeCi
- https://plnkr.co/edit/8XMufI6gtK3L5tTf (old)
- https://plnkr.co/edit/vp146ci8dlTvONZi (updated with cdkScrollable)

The documentation site has a scrollable body element, which also needs a special case.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4194-dashboard-scroll

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4194-dashboard-scroll~CI/)


ashley-hunter (changes requested):
Thanks! Looks good! Just noticed one minor issue when using the scroll wheel, perhaps this is a separate issue, but if I start dragging and then use the scroll wheel to scroll the widget position becomes incorrect.

[Image from Gyazo](https://gyazo.com/2137c64fe177c5be7fd480f21237413c)

DeanHaire (commented):
One thing I noticed in the page example is that you can't seem to move the widget when you move the mouse outside of the example (see gif). Would this also occur in a normal app if you moved your mouse onto the browser's toolbar at the top? Is there anything we can do to allow it to keep moving?


Also, is it just me or is the scrolling really slow on the page example as well? Moving them on the element with overflow example works well, but moving a widget the whole way from the bottom to top on the page example would take forever

donna-marijne:
> One thing I noticed in the page example is that you can't seem to move the widget when you move the mouse outside of the example (see gif). Would this also occur in a normal app if you moved your mouse onto the browser's toolbar at the top? Is there anything we can do to allow it to keep moving?

It seems like it might be a plunker specific issue in Chrome, it works as expected in the doc site. I checked and there are no drag events being emitted outside of the plunker iframe at all in Chrome, so I don't think there's any workaround available.

ashley-hunter (approved):
Looks good to me! Thank you!

DeanHaire (approved):

PR #1496, opened 2021-03-04
EL-4217 - Toggle Button Group: support .btn-sm

pamadei:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-4217
~~There is a whitespace between the `btn-group` class and the `:not` selector in one of the buttons styles. This is causing the margin-bottom and padding styles to be applied incorrectly to the `btn-group`.~~
Here the plunker which showcases the issue: https://plnkr.co/edit/vxWYlrvBAcAMy78Y.

* Updated approach is to use `.btn-sm` and modify the stylesheet to apply appropriate padding values.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_fix-btn-group-styles

#### Downstream Build(s)

https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~fix-btn-group-styles~CI/


donna-marijne (changes requested):
I think this is a misunderstanding, I'm quite sure that the space is not there by accident. The reason being that `.btn` is always a child of `.btn-group`, so the rule as you have written it won't match anything.

Changing this rule will break layouts, due to removal of the bottom margin, so we cannot make this change in a minor release.

In your plunker you have a custom set of styles which reduce the height of the buttons. This is why the alignment is broken - the stylesheet is designed for specific button height and padding values.

The correct way to make a small set of buttons is to use the `btn-sm` class. To support this we would have to modify this rule, e.g.:

```css
.btn-group :not(.dropdown-toggle:first-child).btn.btn-sm {
  padding: 4px 8px;
}
```

Here's an example. Let me know what you think.

https://plnkr.co/edit/8kO2x8Cq98eyZ3OB

pamadei-mf:
> I think this is a misunderstanding, I'm quite sure that the space is not there by accident. The reason being that `.btn` is always a child of `.btn-group`, so the rule as you have written it won't match anything.
> 
> Changing this rule will break layouts, due to removal of the bottom margin, so we cannot make this change in a minor release.
> 
> In your plunker you have a custom set of styles which reduce the height of the buttons. This is why the alignment is broken - the stylesheet is designed for specific button height and padding values.
> 
> The correct way to make a small set of buttons is to use the `btn-sm` class. To support this we would have to modify this rule, e.g.:
> 
> ```css
> .btn-group :not(.dropdown-toggle:first-child).btn.btn-sm {
>   padding: 4px 8px;
> }
> ```
> 
> Here's an example. Let me know what you think.
> 
> https://plnkr.co/edit/8kO2x8Cq98eyZ3OB

Thank you Donna for the explanation. I understand what you mean. I have added a `padding` to the `btn-sm`, so the small buttons are displayed correctly. Let me know what do you think.

donna-marijne (approved):
Great - thank you!

PR #1503, opened 2021-03-11
EL-4212 - Select dropdirection auto does not work

DeanHaire:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4212

- Ensure dropdown appears in the correct direction when using auto

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4212-dropdirection-auto

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4212-dropdirection-auto~CI/)


Sdhilipkumar (approved):
Looks good to me. thanks

src/services/popover-orientation/popover-orientation.service.ts
```diff
@@ -60,7 +66,7 @@ export class PopoverOrientationListener {
 
     private onScrollOrResize() {
         this._rect = this._elementParent ? this._elementParent.parentElement.getBoundingClientRect() : this._element.parentElement.getBoundingClientRect();
-        const itemHeight = this._element.offsetHeight;
+        const itemHeight = this._element.offsetHeight || this._maxHeight;
```
donna-marijne, line 69: What is the situation that is being handled here? (Comments would be helpful.)

DeanHaire: If the element height is 0, then it will instead use the max height value. This stops the dropdown from briefly appearing in the wrong direction and then switching to the correct one

donna-marijne (changes requested):
There does seem to some kind of issue with certain scroll positions:
[Image from Gyazo](https://gyazo.com/3f2f5946be52ffd0953fe7270451e7cc)

https://plnkr.co/edit/tG14WaXweOAWwnl6

src/services/popover-orientation/popover-orientation.service.ts
donna-marijne: Is this because of the animation? What problem would occur if maxHeight was not available in pixels?

DeanHaire: It seems to be because at this point the element may not actually exist and have a height of 0, so it'll incorrectly assume that there's enough space based on it having a height of 0 pixels. Then when the element is actually created it'll correct this and move to the up direction. I'm not sure how to prevent this if the value is not in pixels

donna-marijne (approved):
Looks good - thanks.

PR #1518, opened 2021-04-09
docs: correct chart.js path

xianshenglu:
#### Checklist

#### Ticket / Issue

[Dashboard demo link to plunker](https://uxaspects.github.io/UXAspects/#/components/dashboard) doesn't work. It can't load `chart.js` because the path is not correct.

image

Correct `chart.js` path will make the demo work as expected.

### DIFF
--- a/docs/app/services/playground/utilities/system-helper.ts
+++ b/docs/app/services/playground/utilities/system-helper.ts
@@ -44,7 +44,7 @@ export class SystemJSHelper {
         { name: 'rxjs', path: 'rxjs@6.5.2' },
         { name: 'ngx-bootstrap', path: `ngx-bootstrap@${NGX_BOOTSTRAP_VERSION}/bundles/ngx-bootstrap.umd.min.js` },
         { name: 'chance', path: 'chance' },
-        { name: 'chart.js', path: 'chart.js/dist/chart.min.js' },
+        { name: 'chart.js', path: 'chart.js@2.9.3/dist/Chart.bundle.min.js' },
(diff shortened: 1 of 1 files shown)


docs/app/services/playground/utilities/system-helper.ts
```diff
@@ -44,7 +44,7 @@ export class SystemJSHelper {
         { name: 'rxjs', path: 'rxjs@6.5.2' },
         { name: 'ngx-bootstrap', path: `ngx-bootstrap@${NGX_BOOTSTRAP_VERSION}/bundles/ngx-bootstrap.umd.min.js` },
         { name: 'chance', path: 'chance' },
-        { name: 'chart.js', path: 'chart.js/dist/Chart.bundle.min.js' },
+        { name: 'chart.js', path: 'chart.js/dist/chart.min.js' },
```
donna-marijne, line 47: Thanks! We haven't actually done any testing with 3.x yet, so probably better to clamp this to the version that the package depends upon.
```suggestion
        { name: 'chart.js', path: 'chart.js@2/dist/Chart.bundle.min.js' },
```

donna-marijne (changes requested):

docs/app/services/playground/utilities/system-helper.ts
xianshenglu: Please check again. Changed to `chart.js@2.9.3` which is the version used in package-lock.json.

donna-marijne (approved):
Thank you.

PR #1519, opened 2021-04-14
EL-4229: Dashboard: refreshLayout not working after layout update

JonathanToddMF:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-4229

1. Added call to update layout$ in the refreshLayout function. Also added a call to shiftWidgetsUp in the removeWidget function
2. Test coverage and updated all dashboard screenshots with new button

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4229-dashboard-refreshLayout

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4229-dashboard-refreshLayout~CI/)


cmallon367 (approved):
Looks good!

donna-marijne:
Make sure to fill out the pull request as requested. There are scripts here to help:
https://github.houston.softwaregrp.net/caf/ux-aspects-team/tree/master/userscripts

src/components/dashboard/dashboard.service.ts
```diff
@@ -1090,6 +1090,9 @@ export class DashboardService implements OnDestroy {
         // check whether or not changes have been made - if so we need to repeat until stable
         let stable = true;
 
+        // update information about the layout
+        this.layout$.next(this.getLayoutData());
```
donna-marijne, line 1094: This is being called before any layout changes are made. `shiftWidgetsUp` is also a recursive function, so this is potentially called multiple times. Is that intended?

`shiftWidgetsUp` is called in multiple contexts, not just as part of `refreshLayout`. Have you checked the other contexts as well to confirm that this is valid?

Perhaps an explanation of how this solves the problem in the PR description would be useful.

donna-marijne (changes requested):
Change seems to work well. I have some questions about the fix which would be good to clarify before merging.

src/components/dashboard/dashboard.service.ts
JonathanToddMF: Yeah good points here, I may be able to move update layout call out and into another function that is called before this.

I'll update the description as well. The problem being that the internal list of widgets isn't updated when one is removed. So when shiftWidgetsUp is called it still thinks that widget is still there and doesn't move it up.

But yeah this being recursive it wouldn't be ideal to have this called multiple times. I will have another play around with it.

donna-marijne: I'm not super familiar with the dashboard so hopefully this isn't misleading. But if the problem is the layout not updating after removing a widget, perhaps this could happen in the `removeWidget` function? That's called whenever a widget is destroyed.

JonathanToddMF:
I attempted to move the layout update call out of the shiftWidgetsUp function however I couldn't get it to work.

@ashley-hunter can you see any knock on issues with keeping `this.layout$.next(this.getLayoutData());` in the `shiftWidgetsUp` function? I haven't managed to find anything myself. However just checking encase.

src/components/dashboard/dashboard.service.ts
```diff
@@ -1124,6 +1124,9 @@ export class DashboardService implements OnDestroy {
             return true;
         }
 
+        // emit information about the layout
+        this.layout$.next(this.getLayoutData());
```
ashley-hunter, line 1128: We should also emit this in the `removeWidget` function as well as it looks like the widget is getting left behind in the layout data after a widget has been removed.

ashley-hunter (changes requested):
Looks good, just noticed one more place where the layout data should be updated

ashley-hunter (approved):
Looks good to me, thanks

cmallon367 (approved):
Looks good thanks!

cmallon367:
QA of Dashboard
Tested across all our supported browsers an everything is working as expected
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4229-dashboard-refreshLayout/#/components/dashboard#dashboard 
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/caf_CI_ux-aspects-micro-focus_EL-4229-dashboard-refreshLayout/#/ui-components/layout#dashboard

The issue with the dashboard shown in the plunker below is that whenever a widget was deleted the layoutChange never notified the widgets and the layout therefore never updated similarly when a widget moved and another got deleted the widget never responsed so the refreshlayout was never updated.
Original plunker- https://plnkr.co/edit/Tu4YjeZV2ZEnaSII

Steps to reproduce the issue:
1. Click on 'Delete User' button.
Alert widget gets shifted above as 'User 2' widget gets deleted. (expected)
2. Refresh the UI and this time first try to rearrange the widget 'User 5' using drag handler and bring it below 'User 3' widget. Now try deleting again. 'User 2' widget gets deleted but the 'Alert' doesn't shift up.

This has now been fixed as seen in the plunker below when the steps above are carried out again. The fix was to add a call to update layout$ in the refreshLayout function and also to shiftWidgetsUp in the removeWidget function to make sure the layout continually updates
Fixed Plunker- https://plnkr.co/edit/AoVNNKszSohCZzPr

donna-marijne (approved):
Looks good, thank you.

Roland-Penny (approved):
Accepted

PR #1535, opened 2021-05-11
EL-3961: Dashboard drag and resize issues

JonathanToddMF:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-3961

1. Adjusted hidden attributes on widget drag handles
2. Removed return causing resizing not to happen when in stacked mode.
3. Updated e2e tests and added additional test coverage for resizing in stacked mode

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3961-dashboard-drag-issues

#### Downstream Build(s)

[ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-3961-dashboard-drag-issues~CI/)


JonathanToddMF:
Got a few issues with adding e2e tests for the stacked mode. 

We don't allow keyboards to resize the widgets in stacked mode, this can be seen in the `onResize` function on line 1224 in dashboard/dashboard.service.ts. This makes writing e2e tests for stacked mode resize ability difficult as simulating mouse dragging and dropping in protractor can be difficult, from my understanding.

Although if we can resize widgets with the mouse in stacked mode, shouldn't the ability to on keyboards be allowed? Any advice on this would be appreciated @donna-marijne @ashley-hunter.

Tests for resizing in regular mode are already present so I have updated the name of one of the tests and added an additional test when resizable is explicitly set to false.

donna-marijne:
It does seem like an omission that stack mode widgets cannot be resized using a keyboard. Normally I would suggest raising a new ticket for that, but since it blocks the ability to add test coverage here, it might have to be done as a pre-requisite of this fix. Let me know what you think.

ashley-hunter:
If I recall we had a discussion at the time about this and thought it might be difficult to discover, as the arrow keys move the widget, it would require something like ctrl+arrowkeys to resize (or some other combination). However this discussion I believe occurred before we added all the accessibility functionality to our components so probably should be there.

donna-marijne:
It is difficult to discover, you are right. I think the dashboard needs an accessibility subheading in the documentation, like Angular Material does.

JonathanToddMF:
Hi both, thank you for the quick replies, just wanted both of your opinions on this to see this best course of action.

Yeah it would be ideal that we allow for keyboard accessibility on stacked mode. There is just a if statement that returns at the function at the top to prevent this, so it might not be a lot of work. I will investigate. But agreed Donna best fix this here to allow for test coverage

cmallon367:
There appears to be something wrong with the drag when the usage analytics is the second widget, also happens with the bottom two widgets  https://i.gyazo.com/726944240f68a6363059635e11f5c875.gif https://plnkr.co/edit/M1K28eoBlXZM3vno

JonathanToddMF:
This appears on master as well. I don't think my changes introduced this, however there is a lot of recent changes to the dashboard lately, perhaps this is a regression. I will investigate it

JonathanToddMF:
I had a look at previous versions of aspects back to 3.4.0 and this happens there too, so I think this is correct behaviour

donna-marijne:
It looks good, however I am having trouble with plunker today so verification will be delayed.

donna-marijne (approved):
Verification:
1. Click Edit in Plunker
2. Shrink the preview pane until the dashboard enters stacked mode.
3. Drag the edges of each widget to verify.
Each example contains only one resizable widget, others should not be resizable.

QA:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-3961-dashboard-drag-issues/#/landing - pass
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/caf_CI_ux-aspects-micro-focus_EL-3961-dashboard-drag-issues/#/ui-components/layout#dashboard - pass

Roland-Penny (approved):
Looks good. Accepted.

PR #1549, opened 2021-06-17
EL-4244: Dashboard: programmatically modifying colSpan/rowSpan does not work

JonathanToddMF:
#### Checklist

#### Ticket / Issue
https://portal.digitalsafe.net/browse/EL-4244

1. Updated colSpan and rowSpan to use get and set.
2. Added function to shift overlapping widgets if the row or col span changes to a position over another widget.
3. Added these function calls to the setter of col and row span.
4. Updated test suite
5. Added e2e tests to cover changing col and row span

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4244-dynamic-rowspan-colspan

#### Downstream Build(s)

[ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf~ux-aspects-micro-focus~EL-4244-dynamic-rowspan-colspan~CI/)


JonathanToddMF:
Waiting on builds

src/components/dashboard/widget/dashboard-widget.component.ts
```diff
@@ -42,10 +42,32 @@ export class DashboardWidgetComponent implements OnInit, AfterViewInit, OnDestro
...
+    @Input() get colSpan() {
+        return this.getColumnSpan();
+    }
+
+    set colSpan(colSpan: number) {
+        if (colSpan !== null && colSpan !== undefined && colSpan >= this.minColSpan) {
+            this.setColumnSpan(coerceNumberProperty(colSpan));
+            this.dashboardService.shiftOverlappingWidgets(this);
+            this.dashboardService.shiftWidgetsUp();
+            this.dashboardService.renderDashboard();
```
donna-marijne, line 54: I guess you could refactor these lines to ngOnChanges to avoid duplication.

src/components/dashboard/widget/dashboard-widget.component.ts
```diff
@@ -42,10 +42,32 @@ export class DashboardWidgetComponent implements OnInit, AfterViewInit, OnDestro
     }
 
     /** Defines the number of columns this widget should occupy. */
-    @Input() colSpan: number = 1;
+    @Input() get colSpan() {
+        return this.getColumnSpan();
+    }
+
+    set colSpan(colSpan: number) {
```
donna-marijne, line 49: You need to add the `ngAcceptInputType_colSpan` property in this component so that the Angular compiler knows that strings are valid values for this property as well.

src/components/dashboard/widget/dashboard-widget.component.ts
```diff
@@ -42,10 +42,32 @@ export class DashboardWidgetComponent implements OnInit, AfterViewInit, OnDestro
...
+        if (colSpan !== null && colSpan !== undefined && colSpan >= this.minColSpan) {
```
donna-marijne, line 50: Should probably coerce `colSpan` before doing the numeric comparison.

donna-marijne (changes requested):
Thanks for the PR, good test coverage and works well except for one issue that I have noted.

https://plnkr.co/edit/Uc3jpZilydOeKIfb

In this scenario where the widget expands by more than one row and column, overlapping occurs.
[Image from Gyazo](https://gyazo.com/0e6d56abe8654848195d3343eaddaee4)

You may want to add test coverage for this scenario. Perhaps a new test page would avoid impacting the existing screenshots, which would make it easier to verify that there are no regressions.

src/components/dashboard/widget/dashboard-widget.component.ts
JonathanToddMF: Thanks for that suggestion, allowed me to remove the isNaN changes from the dashboard.service.ts

JonathanToddMF:
Test updates to follow, will add a new test page as suggested and move all added tests to there and cover above test case.

donna-marijne:
I have some feedback in progress, but I need to think about this change a bit more, so I think it will have to miss the code freeze. Sorry about that.

src/components/dashboard/dashboard.service.ts
```diff
@@ -722,16 +781,15 @@ export class DashboardService implements OnDestroy {
...
+        if (isNaN(colShift) || colShift === 0) {
+            return;
+        }
+
         // find the positions required
         const targetSpaces = this.getOccupiedSpaces().filter(space => space.widget === widget).map(space => {
             return { column: space.column - colShift, row: space.row, widget: space.widget };
         });
 
-        // check if any of the target spaces are out of bounds
```
JonathanToddMF, line 730: This needs re-added not sure why I removed this

donna-marijne (changes requested):
Thanks Jonathan, the technique works well. My only concern was the duplication of code, so I've spent some time doing an example refactor. I did not want to push to your branch so you can see it here:

https://github.com/UXAspects/UXAspects/compare/EL-4244-dynamic-rowspan-colspan-refactor?expand=1#diff-5b6c6448c83a889ed139122b55961345580339e9142c884407551a5957206db6

It's not an easy refactor, and you might have some better ideas about how to go about it. The main motivation is to avoid creating new parallel code paths, since future enhancements or fixes would potentially have to make similar changes in multiple places (some of which may be overlooked, creating new bugs).

Another thing I would like to consider is the potentially infinite recursion/loop that is being introduced. It may be worth limiting them to a certain (fairly large) number of iterations, to avoid the possibility of hanging a user's browser tab if there is an unresolvable condition. Throwing an exception would be preferable in this case.

src/components/dashboard/dashboard.service.ts
```diff
@@ -576,27 +576,67 @@ export class DashboardService implements OnDestroy {
...
+                column: widget.col,
+                rowSpan: widget.rowSpan,
+                columnSpan: widget.colSpan
+            };
+
+            do {
+                // Check for overlapping widgets and move them. This may need several iterations.
+                this.shiftWidgetsFromRegion(widgetRegion, widget);
+
+            } while (this.getOverlappingWidgets(widgetRegion, widget).length > 0);
```
donna-marijne, line 622: I suggested that you consider the unbounded loop here, maybe that was overlooked.

src/components/dashboard/dashboard.service.ts
```diff
@@ -576,27 +576,67 @@ export class DashboardService implements OnDestroy {
...
 
         // remove any duplicates
-        widgetsToMove = widgetsToMove.filter((widget, idx, array) => array.indexOf(widget) === idx);
+        return widgetsToMove.filter((wgt, idx, array) => array.indexOf(wgt) === idx);
+    }
+
+
+    /**
+        * Resolve any overlapping widgets after a widget changes rowSpan/colSpan.
+    */
```
donna-marijne, line 602: Something has happened to the comment formatting here, could you fix it so that it matches the others?

donna-marijne (changes requested):

donna-marijne:
@ashley-hunter I've been messing with this as well, so I think we need your input now, sorry about that.

src/components/dashboard/dashboard.service.ts
```diff
@@ -576,27 +576,76 @@ export class DashboardService implements OnDestroy {
...
+
+                // Check for overlapping widgets and move them. This may need several iterations.
+                this.shiftWidgetsFromRegion(widgetRegion, widget);
+                i++
+
+                if (i === limit) {
+                    throw new Error('Unable to resolve overlapping widgets!');
+                }
+
+            } while (this.getOverlappingWidgets(widgetRegion, widget).length > 0 && i <= limit);
```
JonathanToddMF, line 631: Not sure if my solution for this is the best approach, any feedback on this would be appreciated

src/components/dashboard/dashboard.service.ts
```diff
@@ -576,27 +576,76 @@ export class DashboardService implements OnDestroy {
...
+        return widgetsToMove.filter((wgt, idx, array) => array.indexOf(wgt) === idx);
+    }
+
+
+    /**
+     * Resolve any overlapping widgets after a widget changes rowSpan/colSpan.
+     */
+    resizeWidget(widget: DashboardWidgetComponent): void {
+
+        try {
```
ashley-hunter, line 605: Try/Catch blocks can be slow (https://riptutorial.com/javascript/example/5297/avoid-try-catch-in-performance-critical-functions) and this function gets called a lot while resizing, so it might be better to just do `console.error` to log the error and `break` from the loop which would avoid the need for try/catch all together.

ashley-hunter (changes requested):
Looks good, thanks! Just had one small suggestion

src/components/dashboard/dashboard.service.ts
```diff
@@ -576,27 +576,76 @@ export class DashboardService implements OnDestroy {
...
+
+            const widgetRegion = {
+                row: widget.row,
+                column: widget.col,
+                rowSpan: widget.rowSpan,
+                columnSpan: widget.colSpan
+            };
+
+            let i = 0;
+            let limit = 100;
```
donna-marijne, line 619: I would probably move this outside of the function (probably outside of the component as well) so that it doesn't go on the stack each time this is called. I tend to use all caps for actual constants.
```ts
const ITERATION_LIMIT = 100;
```

src/components/dashboard/dashboard.service.ts
```diff
@@ -576,27 +576,76 @@ export class DashboardService implements OnDestroy {
...
+
+            do {
+
+                // Check for overlapping widgets and move them. This may need several iterations.
+                this.shiftWidgetsFromRegion(widgetRegion, widget);
+                i++;
+
+                if (i === limit) {
+                    throw new Error('Unable to resolve overlapping widgets!');
+                }
```
donna-marijne, line 629: The trouble with this is that you are throwing the error before checking if `shiftWidgetsFromRegion` actually worked. 

```ts
let i = 0;
let done = false;
do {
  this.shiftWidgetsFromRegion(widgetRegion, widget);
  done = this.getOverlappingWidgets(widgetRegion, widget).length === 0;
  i += 1;
} while (!done && i < ITERATION_LIMIT);

if (!done) {
  // handle error
}
```

(This code is not tested.)

donna-marijne (changes requested):
Suggestions below:

donna-marijne (changes requested):
Thanks for the update. I tested a few more scenarios and unfortunately found one which causes an error and breaks the dashboard.
https://plnkr.co/edit/2DhYB8jzfh6SvL2L
1. Click the "Bigger" button.

Error in the console:
```
ERROR InternalError: too much recursion
```

This resize operation does work when resizing with drag and drop, which is more or less the same code, so hopefully this is fixable without too much trouble.

donna-marijne:
If there isn't an easy fix for the above issue, you might consider limiting the recursion or converting it to a bounded loop so that it at least doesn't crash, instead reporting some kind of warning. In that case we could probably release this change next week, and fix the remaining issue in a future release.

JonathanToddMF:
It appears related to how the `distance` is calculated in the `shiftWidgetsFromRegion` function. The alert widget's row is getting set to -1. Then the shiftWidgetsUp function is called and this is breaking it as it can't handle the -1 row. Causing the widgets top css to be set to a weird minus number

I think fixing the distance calculation here should fix the problem.

JonathanToddMF:
Fix didn't take long, it was just a case of not running `moveWidgetDown` if the distance is a negative number.

I have added a test case to cover this

src/components/dashboard/dashboard.service.ts
ashley-hunter: I don't think this comment was ever resolved?

src/components/dashboard/dashboard.service.ts
```diff
@@ -576,27 +578,77 @@ export class DashboardService implements OnDestroy {
...
+
+            const widgetRegion = {
+                row: widget.row,
+                column: widget.col,
+                rowSpan: widget.rowSpan,
+                columnSpan: widget.colSpan
+            };
+
+            let i = 0;
+            let done = false;
```
ashley-hunter, line 621: You can just define the iteration limit variable here, its not used anywhere else.

ashley-hunter (changes requested):
Thanks, just noticed one comment that wasn't addressed

ashley-hunter (approved):
Seems to be working well, thank you

JonathanToddMF:
Awaiting acceptance @Roland-Penny 

Plunker showing ability to dynamically change the colSpan and rowSpan of a dashboard widget - https://plnkr.co/edit/TzBRJCNg3d2fH6bY ✔️

Documentation - https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4244-dynamic-rowspan-colspan/#/components/dashboard#dashboard ✔️

Roland-Penny (approved):
Accepted

PR #1557, opened 2021-07-06
EL-4249 - Resizable Columns: widthChange not emitted when the size of the table changes

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4249

1. Added observable to trigger when the window is resized.
2. Added test cases to cover

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4249-resizable-columns-widthChange

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4249-resizable-columns-widthChange~CI/)


src/components/table/table-column-resize/resizable-table-column.component.ts
```diff
@@ -102,9 +102,26 @@ export class ResizableTableColumnComponent implements AfterViewInit, OnDestroy {
...
+                this.setColumnWidth();
+                this.setColumnFlex();
+
+                const width = this._table.getColumnWidth(this.getCellIndex(), ColumnUnit.Pixel);
+
+                if (!isNaN(width)) {
+                    this.widthChange.emit(width);
+                }
+            });
+
```
ashley-hunter, line 124: There is an `onResize` subscription below, which looks like it is not getting fired, I suggest trying to fix it instead as that should resolve the issue. We can't just rely on window resize to do this as a table could be resized for many reasons, e.g. a panel opening shrinking the available space. This window resize event would not get triggered in that case so this would not fix the issue for that case.

ashley-hunter (changes requested):
Thanks, just had a suggestion

src/components/table/table-column-resize/resizable-table-column.component.ts
```diff
@@ -102,9 +102,12 @@ export class ResizableTableColumnComponent implements AfterViewInit, OnDestroy {
     /** Emit when all observables should be unsubscribed */
     private _onDestroy = new Subject<void>();
 
+    resizeObservable$: Observable<Event>;
```
ashley-hunter, line 105: I don't think this is ever used?

src/components/table/table-column-resize/resizable-table-column.component.html
```diff
@@ -1,4 +1,5 @@
-<div class="ux-resizable-table-column-content">
+<div class="ux-resizable-table-column-content"
+    (uxResize)="onResize()">
```
ashley-hunter, line 2: Could we use the ResizeService on the table element rather than having to do this on each column? This would setup up a resize observer for every column rather than one per table which could slow performance on larger tables.

ashley-hunter (changes requested):
Thanks, Just had a few small suggestions, I also notice the size displayed on the last column in the screenshots has now changed? I wouldn't have expected them to change?

ashley-hunter (commented):
The fix looks good, my only concern is the screenshot changes? They are showing the last column width is now different? Is this expected?

JonathanToddMF:
It should of been reverted let me take a look

donna-marijne (changes requested):
Thanks, it seems to work well. However, the new reference image you have added does not have accurate column width values:
image

I think the test is not working as expected. Perhaps you need to run a change detection cycle. If resizing the viewport cannot be made to work, you could try adding a `style.width` binding to the table instead.

donna-marijne (commented):
It looks like this file:
e2e/screenshots/desktop_chrome/column-resize-window-resize--500x800.png

is unused, is that correct?

JonathanToddMF:
> It looks like this file:
> e2e/screenshots/desktop_chrome/column-resize-window-resize--500x800.png
> 
> is unused, is that correct?

Apologies yes this should of been removed. Removed now, waiting on the build passing

donna-marijne (changes requested):
Thanks Jonathan. During testing I found an issue with the uxResizableExpandingTable.

https://plnkr.co/edit/JXxEDpVXqMoroHT0?preview

Drag the plunker splitter to the left. The column headers and table cells become misaligned.
[Image from Gyazo](https://gyazo.com/ff9234e546829f959d5f705490059ccd)

JonathanToddMF:
The problem I am facing with this is that when resizing the table and emitting the onResize event this causes the columns min-widths to be regenerated meaning the uxResizableExpandingTable functions similar to that of the uxResizableTable.

PR #1560, opened 2021-07-09
EL-4250 - Menu: trigger is not linked to menu container (accessibility)

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4250

- added overlay element id as aria-controls 
- Added role menu to overlay container element
- Updated karma tests to cover this

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4250-menu-accessibility

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4250-menu-accessibility~CI/)

### DIFF
--- a/src/components/menu/menu-trigger/menu-trigger.directive.ts
+++ b/src/components/menu/menu-trigger/menu-trigger.directive.ts
@@ -38,7 +38,7 @@ export class MenuTriggerDirective implements OnInit, OnDestroy {
 
     /** Get the aria controls for accessibility */
     get ariaControls(): string | null {
-        return this.menu?.isMenuOpen ? this.menu?.id + '-menu' : null;
+        return this.menu?.isMenuOpen ? this.menu?._ariaControls : null;
     }
 
     /** Store the instance of the focus indicator */
--- a/src/components/menu/menu-trigger/menu-trigger.directive.ts
+++ b/src/components/menu/menu-trigger/menu-trigger.directive.ts
@@ -16,7 +16,7 @@ import { MenuComponent } from '../menu/menu.component';
         '[attr.disabled]': 'disabled ? true : null',
         '[attr.aria-haspopup]': '!!menu',
         '[attr.aria-expanded]': 'menu?.isMenuOpen',
-        '[attr.aria-controls]': 'menu?.isMenuOpen ? this._overlayRef.overlayElement.id : null'
+        '[attr.aria-controls]': 'ariaControls'
     }
 })
 export class MenuTriggerDirective implements OnInit, OnDestroy {
@@ -34,7 +34,12 @@ export class MenuTriggerDirective implements OnInit, OnDestroy {
     private _portal: TemplatePortal;
 
     /** Store the reference to the overlay */
-    _overlayRef: OverlayRef;
+    private _overlayRef: OverlayRef;
+
+    /** Get the aria controls for accessibility */
+    get ariaControls(): string | null {
+        return this.menu?.isMenuOpen ? this._overlayRef?. overlayElement.id : null;
+    }
 
     /** Store the instance of the focus indicator */
     private _focusIndicator: FocusIndicator;
--- a/src/components/menu/menu-trigger/menu-trigger.directive.ts
+++ b/src/components/menu/menu-trigger/menu-trigger.directive.ts
@@ -15,7 +15,8 @@ import { MenuComponent } from '../menu/menu.component';
     host: {
         '[attr.disabled]': 'disabled ? true : null',
         '[attr.aria-haspopup]': '!!menu',
-        '[attr.aria-expanded]': 'menu?.isMenuOpen'
+        '[attr.aria-expanded]': 'menu?.isMenuOpen',
+        '[attr.aria-controls]': 'menu?.isMenuOpen ? this._overlayRef.overlayElement.id : null'
     }
 })
 export class MenuTriggerDirective implements OnInit, OnDestroy {
@@ -33,7 +34,7 @@ export class MenuTriggerDirective implements OnInit, OnDestroy {
     private _portal: TemplatePortal;
 
     /** Store the reference to the overlay */
-    private _overlayRef: OverlayRef;
+    _overlayRef: OverlayRef;
 
     /** Store the instance of the focus indicator */
     private _focusIndicator: FocusIndicator;
--- a/src/components/menu/menu-trigger/menu-trigger.directive.ts
+++ b/src/components/menu/menu-trigger/menu-trigger.directive.ts
@@ -38,7 +38,7 @@ export class MenuTriggerDirective implements OnInit, OnDestroy {
 
     /** Get the aria controls for accessibility */
     get ariaControls(): string | null {
-        return this.menu?.isMenuOpen ? this.menu?.id : null;
+        return this.menu?.isMenuOpen ? this.menu?.id + '-menu' : null;
     }
 
     /** Store the instance of the focus indicator */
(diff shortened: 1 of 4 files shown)


src/components/menu/menu-trigger/menu-trigger.directive.ts
```diff
@@ -33,7 +34,7 @@ export class MenuTriggerDirective implements OnInit, OnDestroy {
     private _portal: TemplatePortal;
 
     /** Store the reference to the overlay */
-    private _overlayRef: OverlayRef;
+    _overlayRef: OverlayRef;
```
ashley-hunter, line 37: Could we keep this private and add a getter specifically for aria controls?

The main reason is `_overlayRef` really should be a marked as possibly undefined `private _overlayRef?: OverlayRef;`, this might require a few changes to other places in the code, but `this._overlayRef.overlayElement.id` could cause errors if we are trying to access `overlayElement` of undefined. Whereas if we have a dedicated getter for this we can use nullish coelescing:

```typescript
get ariaControls(): string | null {
    return this.menu?.isMenuOpen ? this._overlayRef?. overlayElement.id : null;
}
```

src/components/menu/menu/menu.component.html
```diff
@@ -4,6 +4,7 @@
         class="ux-menu"
         [class.ux-sub-menu]="_isSubMenu"
         [class.ux-menu-placement-top]="placement === 'top'"
+        [attr.role]="'menu'"
```
ashley-hunter, line 7:
```suggestion
        role="menu"
```

No need for a binding here as it is not a dynamic value

ashley-hunter (changes requested):
Thanks, just a few small suggestions

src/components/menu/menu.spec.ts
```diff
@@ -536,4 +550,22 @@ describe('MenuTriggerDestroyTestComponent', () => {
...
+        expect(document.querySelector('.btn').getAttribute('aria-controls')).toBeFalsy();
+
+        // open menu
+        component.trigger.openMenu();
+
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        // expect menu to be open
+        expect(document.querySelectorAll('.ux-menu').length).toBe(1);
```
donna-marijne, line 563: I see there's an existing test with this assertion, so we don't need to duplicate it in the new tests.

src/components/menu/menu-trigger/menu-trigger.directive.ts
```diff
@@ -33,7 +34,12 @@ export class MenuTriggerDirective implements OnInit, OnDestroy {
     private _portal: TemplatePortal;
 
     /** Store the reference to the overlay */
-    private _overlayRef: OverlayRef;
+    private _overlayRef?: OverlayRef;
+
+    /** Get the aria controls for accessibility */
+    get ariaControls(): string | null {
+        return this.menu?.isMenuOpen ? this._overlayRef?.overlayElement.id : null;
```
donna-marijne, line 41: I don't think this is right. According to this reference, the `aria-controls` attribute should refer to the element with `role="menu"`:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/menu-button/menu-button-links.html

So `ux-menu` should accept an `id` input and derive an id for the `role="menu"` element, e.g. `[attr.id]="id + '-menu'"`. `aria-controls` will contains the same value. If the developer has not provided an `id` for `ux-menu`, one should be generated. Check the ux-select component for an example of id generation for child elements.

donna-marijne (changes requested):
Thanks for this, I think the target of `aria-controls` needs an adjustment however.

src/components/menu/menu/menu.component.html
```diff
@@ -2,6 +2,8 @@
     <div
         *ngIf="isMenuOpen"
         class="ux-menu"
+        role="menu"
+        [id]="id"
```
donna-marijne, line 6: I mentioned that this should be derived from the provided `id` input rather than applied directly. The reason for that is to avoid duplicate ids in the DOM:
image

donna-marijne (changes requested):
Thanks Jonathan, I think this just needs one additional adjustment.

src/components/menu/menu-trigger/menu-trigger.directive.ts
```diff
@@ -33,7 +34,12 @@ export class MenuTriggerDirective implements OnInit, OnDestroy {
     private _portal: TemplatePortal;
 
     /** Store the reference to the overlay */
-    private _overlayRef: OverlayRef;
+    private _overlayRef?: OverlayRef;
+
+    /** Get the aria controls for accessibility */
+    get ariaControls(): string | null {
+        return this.menu?.isMenuOpen ? this.menu?.id : null;
```
donna-marijne, line 41: aria-controls needs to refer to the id of the element with `role="menu"`, currently it refer to this empty element:
image

donna-marijne (changes requested):

src/components/menu/menu-trigger/menu-trigger.directive.ts
JonathanToddMF: Thanks for catching that Donna, just as a side note, should the `role="menu"` be removed from that empty ux-menu element?

src/components/menu/menu-trigger/menu-trigger.directive.ts
```diff
@@ -33,7 +34,12 @@ export class MenuTriggerDirective implements OnInit, OnDestroy {
     private _portal: TemplatePortal;
 
     /** Store the reference to the overlay */
-    private _overlayRef: OverlayRef;
+    private _overlayRef?: OverlayRef;
+
+    /** Get the aria controls for accessibility */
+    get ariaControls(): string | null {
+        return this.menu?.isMenuOpen ? this.menu?.id + '-menu' : null;
```
donna-marijne, line 41: I think you should store this inner id in a separate property rather than duplicating this magic string in two places.

donna-marijne (changes requested):

src/components/menu/menu/menu.component.ts
```diff
@@ -247,4 +257,8 @@ export class MenuComponent implements AfterContentInit, OnDestroy, OnChanges {
         this._isFocused$.next(false);
     }
 
+    _getAriaControls(): void {
```
donna-marijne, line 260: I guess this should be removed.

src/components/menu/menu/menu.component.ts
```diff
@@ -94,6 +99,11 @@ export class MenuComponent implements AfterContentInit, OnDestroy, OnChanges {
     /** Automatically unsubscribe when the component is destroyed */
     private readonly _onDestroy$ = new Subject<void>();
 
+    /** Get the aria controls label */
+    get _ariaControls(): string {
```
donna-marijne, line 103: Since this is used outside of the component, it doesn't need the underscore - we use that for properties which are public only for the purposes of the component's template. Also I would suggest something like `innerId` because it is only an `aria-controls` value from the perspective of the other control.

donna-marijne (changes requested):

donna-marijne:
Nice work, thank you. Starting QA now.

donna-marijne (changes requested):
I've run the a11y analysis tool that Thomas was using, and there is still one more error that we should probably clear up. Because you added `role="menu"` to the inner div of ux-menu, you just need to remove that role from the component wrapper itself:
```
    host: {
        'role': 'menu'
    },
```

donna-marijne (approved):
Verification:

Use [axe DevTools](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd) to verify that there are no warnings associated with the menu or its trigger.

https://plnkr.co/edit/nyoNVMRcf5A2yp1z (open source) - :heavy_check_mark:
https://plnkr.co/edit/GCikaNO5LBKczzzv (Micro Focus) - :heavy_check_mark:

PR #1566, opened 2021-08-06
EL-4255 - Dropdown covers the button when there is not enough room to expand

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4255

1. Implemented similar code to that of the angular material tooltip fallback (our tooltip uses a deprecated position strategy) - https://github.com/angular/components/tree/master/src/material/tooltip
2.  Added class to change the animation when menu is positioned to the top
3. Added new fallback test suite similar to tooltips
4. Updated original menu tests and screenshots

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4255-menu-fallback-position

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4255-menu-fallback-position~CI/)


src/components/menu/menu-trigger/menu-trigger.directive.ts
```diff
@@ -362,45 +358,80 @@ export class MenuTriggerDirective implements OnInit, OnDestroy {
...
+    private invertPosition(x: HorizontalConnectionPos, y: VerticalConnectionPos) {
+        if (this.menu.placement === 'top' || this.menu.placement === 'bottom') {
+            if (y === 'top') {
+                y = 'bottom';
+            } else if (y === 'bottom') {
+                y = 'top';
+            }
+        } else {
+            if (x === 'end') {
+                    x = 'start';
```
donna-marijne, line 428: Indentation issue here.

donna-marijne (changes requested):
Thanks Jonathan, looks like it works really well. I only noticed one minor animation issue with the collapsing menu:
[Image from Gyazo](https://gyazo.com/970b7a18c204b25c0be4c0e81a445e46)

Could you consider refactoring this into e.g. a service so that we don't have to maintain two very similar code bases across tooltip and menu?

donna-marijne:
Realistically, a technical debt ticket will never come out of the backlog. So if you introduce technical debt like duplicated code, it's there for potentially a very long time, probably forever. In my view, it's preferable to do the best you can with each commit to master. In the best case of committing this as it is, you just do the same work later instead now (might as well do it now in that case); in the worst case it causes confusion and extra effort to maintain similar but distinct code bases before someone finally gets around to doing the refactor.

donna-marijne:
Thanks, if you run into any trouble just let me know.

JonathanToddMF:
Hi @donna-marijne 

I seem to be having an issue with getting the micro focus build to pass with this. I am getting the following error locally
`Compiling TypeScript sources through NGC ERROR: Maximum call stack size exceeded`

The same error can be seen in the build here - https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4255-menu-fallback-position~CI/13/console @2021-08-13T18:18:43.841963141Z

I was doing some googling about this issue and can't find anything concrete. Any help on this I would be grateful. I have a feeling it could be a circlular dependency somewhere. I have a feeling it could be how I am importing the new service into the `popover.directive.ts`

Thanks, 
Jonathan

src/services/overlay-fallback/overlay-fallback.service.ts
```diff
@@ -0,0 +1,149 @@
+import { ConnectedPosition, FlexibleConnectedPositionStrategy, HorizontalConnectionPos, OriginConnectionPosition, OverlayConnectionPosition, OverlayRef, VerticalConnectionPos } from '@angular/cdk/overlay';
+import { Injectable } from '@angular/core';
+import { AnchorPlacement } from '../../components/tooltip';
```
donna-marijne, line 3: You do have at least one circular dependency here, the fix for this one would be to move the AnchorPlacement type definition to a separate file alongside this service. See if that helps.

JonathanToddMF: Thanks for this, however it didn't fix the issue I am facing

src/components/menu/menu/menu.component.ts
```diff
@@ -3,7 +3,8 @@ import { FocusKeyManager, FocusOrigin } from '@angular/cdk/a11y';
 import { AfterContentInit, ChangeDetectionStrategy, ChangeDetectorRef, Component, EventEmitter, HostBinding, Inject, Input, OnChanges, OnDestroy, Optional, Output, QueryList, SimpleChanges, TemplateRef, ViewChild, ViewRef } from '@angular/core';
 import { BehaviorSubject, merge, Observable, Subject } from 'rxjs';
 import { map, switchMap, takeUntil } from 'rxjs/operators';
-import { AnchorAlignment, AnchorPlacement } from '../../tooltip/index';
+import { AnchorPlacement } from '../../../services/overlay-fallback';
+import { AnchorAlignment } from '../../tooltip/index';
```
donna-marijne, line 7: Another one here.

donna-marijne:
Actually, if you could work on something else for the time being, I'll pull this down for a closer look (probably tomorrow).

JonathanToddMF:
> Actually, if you could work on something else for the time being, I'll pull this down for a closer look (probably tomorrow).

Thanks appreciate it

donna-marijne:
I've made some changes which should hopefully solve the issue. I was going to suggest changing the service to a singleton and some other minor changes, but since I was working on the code I figured it would be easier to made the changes directly. Let me know if there are any problems or build failures.

JonathanToddMF:
> I've made some changes which should hopefully solve the issue. I was going to suggest changing the service to a singleton and some other minor changes, but since I was working on the code I figured it would be easier to made the changes directly. Let me know if there are any problems or build failures.

Thanks for these changes Donna, looks like both builds are building locally.  

I am looking through the changes to see how I can use these going forward. 

Appreciate it,
Jonathan

donna-marijne (approved):
Verification:

https://plnkr.co/edit/JXgqiPccOgzNlPhj (open source) - :heavy_check_mark:
https://plnkr.co/edit/VKF3GsDNYJPMoJPN (Micro Focus) - :heavy_check_mark:

Roland-Penny (approved):
Looks good

PR #1574, opened 2021-08-26
EL-4267 - Column Picker: selectedChange emits old data on reorder

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4267

Added setTimeout when emitting onReorder which fixes the issue.
Added e2e test coverage it was hard to figure out a way to do this in karma tests

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4267-selectedChange-emits-old-data

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4267-selectedChange-emits-old-data~CI/)


src/components/table/column-picker/column-picker.component.ts
```diff
@@ -186,7 +186,9 @@ export class ColumnPickerComponent implements OnChanges {
 
     /** Update when reordering has occurred */
     onReorder(): void {
-        this.selectedChange.emit(this.selected);
+        setTimeout(() => {
```
donna-marijne, line 189: I think using setTimeout here is a bit of a hack. Looking at the source, I guess that the problem is that `reorderEnd` emits before `reorderableModelChange`. Perhaps you can emit `selectedChange` directly after `selected` has been updated instead.

e2e/pages/app/column-picker/column-picker.testpage.component.ts
```diff
@@ -44,10 +44,16 @@ export class ColumnPickerTestPageComponent {
...
     ];
 
+    selectedChange: ReadonlyArray<string>;
+
     groups: ColumnPickerGroup[] = [
         { name: 'Metadata', expanded: true }
     ];
 
+    _selectedChange(change: any) {
+        this.selectedChange = change;
```
donna-marijne, line 54: Establishing a duplicate of `this.selected` to test this issue indicates that there is another issue with the column picker. It looks like the array bound to the `selected` input is being modified in place, which is why it gets the updated value while `selectedChange` returns an incorrect value. That should be fixed instead of making this change to the test.

Also, don't use `any`, and give your functions return types.

donna-marijne (changes requested):

e2e/pages/app/column-picker/column-picker.testpage.component.ts
JonathanToddMF: Hi Donna, thanks for the feedback on this. I have reverted the changes to this test suite. But I am unsure as to how to test the `selectedChange` value without the above function and changes to the test HTML. 

`selectedChange` now returns the correct value expected but I tried writing a Karma test for this, however it doesn't seem possible to reorder the selection from it to trigger the `selectedChange`

Further help on this would be appreciated.

Thanks,
Jonathan

donna-marijne: The array bound to the `selected` input is being modified in place (it is meant to be readonly). Once that is fixed, no additional test coverage is needed. `[(selected)]="selected"` is equivalent to `[selected]="selected" (selectedChange)="selected = $event"`, so you already have `selectedChange` hooked up in this test.

JonathanToddMF: Thanks for that Donna, sorry was in training most of today. 
So from this recommendation the selected input cannot change and can only be ReadonlyArray and therefore there would need to be another variable to store the augmented and changed state of the selected items when it is reordered or items are removed and added?

Sorry if I am completely missing misunderstanding this one.

donna-marijne: Yes, if the array is to be modified you need to create a whole new array which is normally done with the spread operator.
```
this.selected = [...this.selected];
```

JonathanToddMF:
Hi Donna I have had a look through the component and can't find a situation where `this.selected` was being changed place. All instances seemed to be using the correct spread operator you mentioned. I did however see `this._selectedSelection` being changed in place and changed it to being a spread operator.

donna-marijne:
You can check whether the array is being mutated by removing the output binding and rearranging some columns, e.g.:
https://plnkr.co/edit/XqKA0qLunbifmnuR

So the original reference is still being improperly modified.

[Here](https://github.com/UXAspects/UXAspects/blob/master/src/components/table/column-picker/column-picker.component.html#L141) you can see the `selected` array is being passed into the `uxReorderable` directive. The type of the `reorderableModel` input is `Array<T>` so that is where you are losing the typescript safety of `ReadonlyArray`. [This](https://github.com/UXAspects/UXAspects/blob/master/src/directives/reorderable/reorderable.directive.ts#L72) is most likely where your array is actually getting mutated. So you need to shallow copy the array before it gets to this directive.

After that's fixed, the existing test should catch the original problem without needing modification.

JonathanToddMF:
> You can check whether the array is being mutated by removing the output binding and rearranging some columns, e.g.:
> https://plnkr.co/edit/XqKA0qLunbifmnuR
> 
> So the original reference is still being improperly modified.
> 
> [Here](https://github.com/UXAspects/UXAspects/blob/master/src/components/table/column-picker/column-picker.component.html#L141) you can see the `selected` array is being passed into the `uxReorderable` directive. The type of the `reorderableModel` input is `Array<T>` so that is where you are losing the typescript safety of `ReadonlyArray`. [This](https://github.com/UXAspects/UXAspects/blob/master/src/directives/reorderable/reorderable.directive.ts#L72) is most likely where your array is actually getting mutated. So you need to shallow copy the array before it gets to this directive.
> 
> After that's fixed, the existing test should catch the original problem without needing modification.

I appreciate this Donna. I will take a look! Thanks!

donna-marijne (approved):
Verification:

https://plnkr.co/edit/BosPZZo3xnQUprI3 - :heavy_check_mark:
https://plnkr.co/edit/f4ESzpUVAoO2pR3e - :heavy_check_mark:

Roland-Penny (approved):
Accepted based on dev QA feedback

PR #1579, opened 2021-09-15
EL-4265 - Select Component emits null when programmatically changing value

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4265

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4265-select-emits-null

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4265-select-emits-null~CI/)

### DIFF
--- a/src/components/select/select.component.ts
+++ b/src/components/select/select.component.ts
@@ -255,9 +255,11 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
             filter(() => this.allowNull),
             filter(value => !this.multiple && value !== this.getDisplay(this.value)),
             takeUntil(this._onDestroy)
-        ).subscribe(() => {
+        ).subscribe((value) => {
+            if (value === '') {
+                this.valueChange.next(null);
+            }
             this.value = null;
-            this.valueChange.next(null),
             this._onChange(null);
         });
 


src/components/select/select.component.ts
```diff
@@ -377,6 +376,15 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
             this._userInput = true;
             this._dropdownOpen = true;
         }
+
+        setTimeout(() => {
+            let target = event.target as HTMLInputElement;
+            if (event.keyCode === DELETE && target.value === '' && this.allowNull) {
+                this.valueChange.next(null);
+            }
+        }, 10);
```
JonathanToddMF, line 385: I know this probably isn't an ideal solution for this (it will fire if you keep pressing DELETE for example), the problem I am having is getting the current value of the text in the input as the _input$ observables fire after this is called. Like you mentioned, perhaps the way this component handles this input would need changed?

donna-marijne: According to the provided plunker, the issue is that null is emitted when the `value` input is set programmatically, so it's not really anything to do with key handling. If you trace what happens when the setter for `value` is called you will find that it ultimately calls `valueChange`. What you need to do is eliminate any calls of valueChange which come from changes to the component's inputs. It should only be called when the user does something (keydown, click, blur, etc).

I think this will probably be a complicated task. Since the component relies heavily on observables, you may need to change them to carry a structured object which includes a flag indicating whether the change from from a user action or not.

donna-marijne: Correction to the above comment, I do see what you are getting at with this fix. I'm not sure if it can be made to work without the kind of systematic change I mentioned above however. If you carry a flag along with the string value of input you can decide whether to emit the event. Hopefully that makes sense.

JonathanToddMF:
Think there is on going build issues, however this UX Aspects build seems to be passing

src/components/select/select.component.ts
```diff
@@ -46,8 +46,8 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
 
     /** The text in the input area. This is used to filter the options dropdown. */
     @Input()
-    set input(value: string) {
-        this._input$.next(value);
+    set input(value: InputValue) {
```
ashley-hunter, line 49: I don't think we can change the type of `input` and `inputChange`, this would introduce a breaking change for anyone relying on these

ashley-hunter (changes requested):
Thanks, unfortunately I don't think we can make this exact change as this would introduce a breaking change.

You might be able to make the change on the internal properties only (e.g. _input$) but we would have to retain the same type for the inputs and outputs

src/components/select/select.component.ts
```diff
@@ -381,7 +383,15 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
...
     /** This gets called whenever the user types in the input */
     onInputChange(input: string): void {
-        this.inputChange.emit(input);
+
+        this.input = input;
+
+        this._input$.next({
+            value: this.input,
+            userInteraction: true
+        });
```
ashley-hunter, line 392: The `this.input` setter emits _input$, so this code looks like it would be emitting the observable twice? Is that expected?

I think this could probably just be this, but I haven't checked it:

```suggestion
        this._input$.next({
            value: input,
            userInteraction: true
        });
```

ashley-hunter (changes requested):
Thanks, I just have one query about an observable possibly being fired too many times

src/components/select/select.component.ts
```diff
@@ -381,7 +383,13 @@ export class SelectComponent<T> implements OnInit, OnChanges, OnDestroy, Control
...
     /** This gets called whenever the user types in the input */
     onInputChange(input: string): void {
-        this.inputChange.emit(input);
+
+        this._input$.next({
+            value: input,
+            userInteraction: true
+        });
+
+        this.inputChange.next(this.input);
```
ashley-hunter, line 392:
```suggestion
        this.inputChange.emit(this.input);

```

ashley-hunter (changes requested):
Thanks looks good, just notice one very minor issue.

ashley-hunter (changes requested):
I notice some behaviour I don't think is correct:

https://plnkr.co/edit/6RLUzXi3odxdKIVl

If I click in the select box, then click out the input value gets cleared:

[Image from Gyazo](https://gyazo.com/664ec74394355d45ed4848d482148579)

ashley-hunter (approved):
Thanks, looks good to me.

donna-marijne (approved):
Verification:

https://plnkr.co/edit/aKm1uecv9RRA51Df (open source) - :heavy_check_mark:
https://plnkr.co/edit/WdC7KHynlLvsYjcG (Micro Focus) - :heavy_check_mark:

PR #1581, opened 2021-09-21
EL-4269 - Number Picker: add readonly mode

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4269

1. Added readonly input, changed styling to remove grey background and text.
2. Added new test suite, tests are written with browser.actions because already existing helper functions return an invalid element state as the input is now readonly.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4269-number-picker-readonly-mode

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4269-number-picker-readonly-mode~CI/)


e2e/tests/components/number-picker/readonly/number-picker-read-only.e2e-spec.ts
```diff
@@ -0,0 +1,93 @@
...
+            expect(await page.emittedValue.getText()).toBe('{ "readonly": 6 }');
+        });
+
+        it('should allow changes to the value by hitting the down arrow on the keyboard', async () => {
+            await browser.actions().mouseMove(page.numberPickerReadOnly).click().perform();
+            await browser.actions().sendKeys(Key.ARROW_DOWN).perform();
+
+            expect(await page.emittedValue.getText()).toBe('{ "readonly": 4 }');
+        });
+    });
```
donna-marijne, line 92: I would hope that these are covered by the existing e2e tests for number picker, so it should be safe to omit them. Bear in mind that the e2e tests are extremely slow to run, so we only want to add tests which definitely add value.

src/components/number-picker/number-picker.component.ts
```diff
@@ -45,6 +45,9 @@ export class NumberPickerComponent implements ControlValueAccessor, OnDestroy, O
     /** Specified if this is a required input. */
     @Input() required: boolean;
 
+    /** Specified if this is a readonly input. */
+    @Input() readonly: boolean = false;
```
donna-marijne, line 49: As usual, you should coerce boolean inputs.

e2e/tests/components/number-picker/readonly/number-picker-read-only.e2e-spec.ts
```diff
@@ -0,0 +1,93 @@
...
+
+            expect(await page.emittedValue.getText()).toBe('{ "readonly": 5 }');
+        });
+
+        it('should prevent changes to the value by hitting the down arrow on the keyboard', async () => {
+            await browser.actions().mouseMove(page.numberPickerReadOnly).click().perform();
+            await browser.actions().sendKeys(Key.ARROW_DOWN).perform();
+
+            expect(await page.emittedValue.getText()).toBe('{ "readonly": 5 }');
+        });
```
donna-marijne, line 48: The missing test case here is the scroll wheel.

e2e/tests/components/number-picker/readonly/number-picker-read-only.po.spec.ts
```diff
@@ -0,0 +1,25 @@
+import { $, browser, by, element, ElementFinder } from 'protractor';
+import { NumberPickerPage } from '../number-picker.po.spec';
+
+export class NumberPickerUpdateOnPage extends NumberPickerPage {
```
donna-marijne, line 4:
```suggestion
export class NumberPickerReadOnlyPage extends NumberPickerPage {
```

donna-marijne (changes requested):
It looks good, I have some feedback.

e2e/pages/app/number-picker/readonly/number-picker-read-only.testpage.component.html
```diff
@@ -0,0 +1,16 @@
+<a id="top-focus" href="javascript:void(0)">Top focus target</a>
+
+<div id="root">
+
+    <form [formGroup]="form">
+        <p>readonly:</p>
+        <ux-number-picker id="read-only-number-picker" class="number-picker-component-demo" min="0" max="10" [readonly]="readonly" formControlName="readonly"></ux-number-picker>
```
donna-marijne, line 7: Since you are visually testing, you should also include `ux-number-picker-inline`. It has a separate appearance in the Micro Focus theme.

I also notice that you have set min and max both in the markup and in the form builder, with conflicting values. This test does not cover min/max so you can remove all of that. Best to keep tests simple.

donna-marijne (changes requested):

e2e/tests/components/number-picker/readonly/number-picker-read-only.e2e-spec.ts
JonathanToddMF: It appears that protractor has a hard time simulating scroll events, it doesn't have any built in support. I tried using `browser.executeScript('window.scrollBy(0,100);');` on the element but it doesn't seem to change the value when readonly is set to false. 
Is there any e2e tests that I could use for help? I had a look but couldn't find anything.

donna-marijne: I believe you will have to do it using a karma test, there is a specific event for the mousewheel which is not invoked just by scrolling.
https://developer.mozilla.org/en-US/docs/Web/API/Element/wheel_event

JonathanToddMF: Awesome thanks!

ashley-hunter (changes requested):
Looks good, I just noticed one issue, if I set readonly="false" it acts the same as if readonly were true where I can't edit the input field:

https://plnkr.co/edit/LMUViCSCDkYkjgA5

JonathanToddMF:
> Looks good, I just noticed one issue, if I set readonly="false" it acts the same as if readonly were true where I can't edit the input field:
> 
> https://plnkr.co/edit/LMUViCSCDkYkjgA5

Weird this, I think the other build may be out of sync. Running it again to check

ashley-hunter (approved):
Looks good, thanks

JonathanToddMF:
Ready for verification @Roland-Penny 

The readonly status is applied to the number picker components in the following plunkers

Aspects - https://plnkr.co/edit/naz9WjnKrxngjcC3 
Micro Focus - https://plnkr.co/edit/5WLVi4qkdQY2yfrJ

Roland-Penny:
Looks good - accepted

PR #1596, opened 2021-10-29
EL-4277 - Combobox: Issue with combox controls and custom validators

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4277

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4277-combobox-validators

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4277-combobox-validators~CI/)


src/components/select/select.component.spec.ts
```diff
@@ -943,3 +943,81 @@ describe('Select with recent options', () => {
...
+                [disabled]="disabled"
+                [required]="true"
+                [options]="options">
+            </ux-select>
+        </form>
+
+        <span class="info">
+            <span>select value: <span id="select-value">{{ form.get('select')?.value }}</span></span><br>
+            <span>select errors: <span id="select-errors">{{ (form.get('select')?.errors | json) }}</span></span>
+        </span>
```
donna-marijne, line 962: This is a unit test, so you don't need to render and and extract values from the DOM to test.

src/components/select/select.component.spec.ts
```diff
@@ -943,3 +943,81 @@ describe('Select with recent options', () => {
...
+    it('should remove required error once option has been selected', async () => {
+        component.disabled = false;
+        fixture.detectChanges();
+        await fixture.whenStable();
+
+        expect(selectValue.innerText).toBe('');
+        expect(selectErrors.innerText).toBe('{ "required": true }');
+
+        const input = nativeElement.querySelector<HTMLInputElement>('input');
+        console.log('file: select.component.spec.ts ~ line 1013 ~ fit ~ input', input);
```
donna-marijne, line 1011: This should be removed.

src/components/select/select.component.spec.ts
```diff
@@ -943,3 +943,81 @@ describe('Select with recent options', () => {
...
+            </ux-select>
+        </form>
+
+        <span class="info">
+            <span>select value: <span id="select-value">{{ form.get('select')?.value }}</span></span><br>
+            <span>select errors: <span id="select-errors">{{ (form.get('select')?.errors | json) }}</span></span>
+        </span>
+    `
+})
+export class SelectNullValueTestComponent {
```
donna-marijne, line 965: This issue is about the required validator, maybe a more accurate name can be chosen.

donna-marijne (changes requested):
Based on testing the given repro example, this does not seem to resolve the problem.
[Image from Gyazo](https://gyazo.com/c31633da65ee903ff4a37641fe472665)

Additionally as seen above, the value appears briefly in red, which is not desirable.

JonathanToddMF:
> Based on testing the given repro example, this does not seem to resolve the problem. [Image from Gyazo](https://gyazo.com/c31633da65ee903ff4a37641fe472665)
> 
> Additionally as seen above, the value appears briefly in red, which is not desirable.

Hey Donna, I noticed something about their example plunker which isn't correct. I have detailed the issue in a response to Vadim here - https://teams.microsoft.com/l/message/19:f9c3e360e0454784a9fb04c49f04ea65@thread.skype/1635870242326?tenantId=856b813c-16e5-49a5-85ec-6f081e13b527&groupId=51fc2a42-4497-4231-baca-19348d1ef830&parentMessageId=1634891721057&teamName=%23ux-aspects&channelName=General&createdTime=1635870242326

The colour appearing briefly in red is still an issue I am working to resolve.

JonathanToddMF:
Hi @donna-marijne, I have been looking at the issue regarding the red colour and can't seem to find a way to resolve it. There seems to be a delay before the ng-invalid class is removed. There doesn't seem to be any timers or delays in the code which would introduce this. Unless I am missing something? 

Any help would be appreciated!

donna-marijne (changes requested):
Documentation resolution:
- Add a note to the `disabled` and `required` inputs that they should not be used within a form.
- Add some example code under a new header showing how to use the `required` validator within a form. There is an example of form validation in the [Text Input](https://pages.github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/#/ui-components/input-controls#raised-text-input) documentation.

JonathanToddMF:
Closing this as the branch isn't the same name as the ux-aspects-micro-focus branch

PR #1598, opened 2021-11-10
EL-4288 - Angular 13

JonathanToddMF:
#### Checklist

- Migrated project to use Angular CLI. Custom webpack configs were difficult to maintain and updating them was problematic with many issues such as the build not updating when we changed files, and I spent days trying to figure it out with no success, so moving to the CLI was the easier approach. It should also greatly help with future updates and make moving to Nx easy should we ever choose to.
- Had to keep plunkers using Angular 12, Angular 13 removes view engine which requires a compilation step in the browser which plunker does not support.
- TSLint is deprecated and removed from Angular, so using ESLint instead

#### Ticket / Issue

https://portal.digitalsafe.net/browse/El-4288

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_El-4288-angular-13

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~El-4288-angular-13~CI/)


JonathanToddMF:
Hi @donna-marijne, I think I have updated most of the current dependencies and when running npm run start I am getting a few errors of "One or more import cycles would need to be created to compile this component, which is not supported by the current compiler configuration."

I am not sure if this is something that needs updated as part of the angular 13 update or is something I have missed. Any advice on this would be great, maybe running npm run start at this stage isn't the correct way to test at this stage?

ashley-hunter:
Thanks Jonathan, I am getting quite a few errors when I try to run `npm start` which is to be expected. As of Angular 13 View Engine is no longer supported, so I don't believe there is a way for us to easily continue support Angular 11 either which required libraries to be shipped in the View Engine format. As of recently Angular's approach is now deprecate in one major version and remove in the next major version, which means we likely will only be able to support 2 versions of Angular going forward rather than the 3 we previously did. This aligns with how we do it in Universal too.

To fix the errors you need to tell Angular to build the library in the Ivy `partial` format. In the `tsconfig-build.json` file you need to remove `"enableIvy": false` and add `"compilationMode": "partial"` instead.

You will also likely get some errors complaining about `.mjs` files, so you will need to update the webpack configs to add that as a possible extension (note we should also now be able to remove the `rxAlias` that was in the resolve section before):

```javascript
resolve: {
        extensions: ['.ts', '.js', '.mjs'],
},
```

And then update the `rules` section to include the following rule:

```javascript
{
    test: /\.m?js/,
    resolve: {
        fullySpecified: false
    }
},
```

Angular 13 has also dropped support for IE 11. While this shouldn't effect people consuming our library (if they are using Angular 12), we can probably now remove the `// Downlevel Angular Packages` section from our webpack config as this was required to support IE11, but if we are using Angular 13 we can't support IE 11 anyway.

There may be additional Webpack config changes required, I haven't checked, but for reference you can see the configuration the Angular CLI uses here which may be useful as a reference:

https://github.com/angular/angular-cli/blob/master/packages/angular_devkit/build_angular/src/webpack/configs/common.ts

ashley-hunter:
Hey @JonathanToddMF, I pushed a fix for all the NG3003 errors that were appearing on serve. Hopefully that will help you out a bit.

JonathanToddMF:
Thanks @ashley-hunter!

donna-marijne (changes requested):
Initial verification results:

This page hangs the browser tab (Firefox and Chrome on Windows):
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_El-4288-angular-13/#/showcase/visualizations/partition-map

Hierarchy bar icons are missing (also affects org chart example):
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_El-4288-angular-13/#/components/hierarchy-bar

Otherwise works well. Code review to follow.

ashley-hunter:
Thanks Donna, the partition map showcase issue is a pre-existing issue. I will look into the missing icons, thanks!

src/styles/ux-aspects.less
```diff
@@ -1,6 +1,8 @@
 // Third Party Imports
-@import '~@angular/cdk/overlay-prebuilt.css'; // CDK Classes for using overlays (ie. tooltips, menus, popovers etc..)
-@import '~@angular/cdk/text-field-prebuilt.css'; // CDK Classes for using AutoFillMonitor
+@import '../../node_modules/@angular/cdk/overlay-prebuilt.css';
+// CDK Classes for using overlays (ie. tooltips, menus, popovers etc..)
+@import '../../node_modules/@angular/cdk/text-field-prebuilt.css';
+// CDK Classes for using AutoFillMonitor
```
donna-marijne, line 5: Looks like these comments got moved in a misleading way.

donna-marijne (changes requested):
Thanks for the intense work to revise the build for this project. I have added some comments. Not quite clear on the removal of IE11 at this point. If someone has set up an Angular 12 app for IE11 I would expect them to be able to deploy UX Aspects.

I noticed that some auto formatting has been applied to affected files, but I was not able to determine which tool has been configured for this. If we do apply formatting, I would prefer that it uses the same rules that we use in Universal.

ashley-hunter:
Thanks for the feedback Donna, I'll restore the IE content, that was an oversight on my part.

As for the formatting changes, that was probably due to my editor setup using universal. You mention you'd like it to be consistent with Universal, would you like me to bring in prettier with the same config? I realise we didn't do this before because it can make git blame harder, but if you think now is an acceptable time to set it up I can easily add it in?

donna-marijne:
Thanks Ashley. If it's easy enough to do, you could bring in a prettier config to make sure that formatting is consistent when it is done. Not sure about adding the git hook though, maybe we can leave it as a manual step for a while and see how it goes.

JonathanToddMF:
Just to add to the above, I have spent a good portion of the day going through most of the common commands that I use when making changes to the repo and can't find anything other than the missing icons mentioned above. I will look through the code base for anything I notice.

Thanks again Ashley for the work put into this, should make things a lot easier going forward.

ashley-hunter:
@donna-marijne I have updated the package.json to make chart.js and ng-2charts peer dependencies with a restriction so the version is <3. This should prevent the issue you were seeing from arising.

donna-marijne (approved):
Tested documentation and in standalone Angular 12 and 13 projects, all looks good to me.

PR #1610, opened 2022-01-17
EL-4297 - EL-4297-organizational-chart-example

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://portal.digitalsafe.net/browse/EL-4297

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_EL-4297-organizational-chart-example

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~EL-4297-organizational-chart-example~CI/)

### DIFF
--- a/src/components/organization-chart/organization-chart.component.ts
+++ b/src/components/organization-chart/organization-chart.component.ts
@@ -339,6 +339,7 @@ export class OrganizationChartComponent<T> implements AfterViewInit, OnChanges, 
 
         // reset the connector changed status
         this._hasConnectorChanged = false;
+        this._isTransitioning = false;
     }
 
     /** Select a specified node */


JonathanToddMF:
Hi @donna-marijne ,

The issue with this seems to be be that `_isTransitioning` isn't getting set to false when on the inner source side. 

Once you click on a node the `onClick` function fires and and runs the `toggle` function which gets to line 400 - https://github.com/UXAspects/UXAspects/blob/master/src/components/organization-chart/organization-chart.component.ts#L397, an if statement returning if `_isTransitioning` is true and stops the code.

I've added a line of code to the end of the render method that sets transitioning to false explicitly. This does not fix the issue unless stepped through using a debugger. This leads me me to believe that there is something being lazy loaded or a race condition? Is there anything on the inner source side that could cause this?

I am hesitant to make changes to the component as this surely must be an issue with the documentation on the inner source side? 

Any help would be appreciated.

PR #1618, opened 2022-01-27
382091 - Dropdown Menu: Tabbing issues

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=382091

- Added tab listener to menu item to close menu when pressed.
- Removed detachments listener as it was causing the close menu function to be called more times than necessary - This is to fix the issue of the focus flicking between the next tab item and the menu item.
- Added test coverage

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_382091-dropdown-menu-tabbing

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~382091-dropdown-menu-tabbing~CI/)


src/components/menu/menu/menu.component.ts
```diff
@@ -241,6 +242,10 @@ export class MenuComponent implements AfterContentInit, OnDestroy, OnChanges {
 
         // emit the keydown event
         this._onKeydown$.next(event);
+
+        if (event.keyCode === TAB) {
```
donna-marijne, line 246: Use the blur event instead of listening for a tab key. Some screen readers use other keys as well as tab to change focus.

donna-marijne (changes requested):
This only seems to solve one of the points noted in the ticket. Focus does not progress to the next item in the tab ring.
menu-focus

JonathanToddMF:
Hey @donna-marijne I have done some refactoring on this. Just one thing I have noticed about the above gif you included. Is this using the same menu for each of the buttons? I believe this is what is causing the issue. Removing the same menu trigger for the other buttons seems to fix the issue. I think the same instance of the trigger is causing it to go to the next tab item which would be the plunker iframe in this example. 

If you have another button that isn't using the same menu trigger it fixes the issue. I don't believe there would be a scenario in which there would be two buttons opening the same menu beside each other? Here is a plunker to demonstrate - https://plnkr.co/edit/nAYn3LKv3Z2OuuA7

donna-marijne:
I believe it was a slightly modified version of our documentation example.

JonathanToddMF:
I did some playing around with this. I managed to solve the issue.

src/components/menu/menu/menu.component.ts
```diff
@@ -167,6 +168,7 @@ export class MenuComponent implements AfterContentInit, OnDestroy, OnChanges {
         this._activeItem$.complete();
         this._items$.complete();
         this._placement$.complete();
+        this._keyManager.tabOut.complete();
```
donna-marijne, line 171: I don't think this is appropriate, you didn't create this observable so you should not terminate it.

donna-marijne (changes requested):
Thanks, it seems to work well. Could you revert the changes to the existing e2e tests, since the changes alter the nature of the tests.

JonathanToddMF:
Hi @donna-marijne with regards to the e2e test changes. These changes mixed with the space bar fix fail the original tests.

Originally these were to test when the menu trigger loses focus when the menu was open, but because the space bar now focuses the first menu item and tabbing away from the menu it will close it, this breaks one of the tests. I changed the tests to a click as it doesn't focus the first item in the list.

Although, looking at the documentation for closeOnBlur - https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_382091-dropdown-menu-tabbing/#/components/buttons#dropdowns - we have it listed for when the menu loses focus, although the closeOnBlur doesn't cover the requirements in this ticket, the code changes in this PR work better than the original closeOnBlur implementation. 

Perhaps these changes should replace the current closeOnBlur functionality?

donna-marijne:
Hi Jonathan, just want to be a bit careful about this so that we don't revert a fix that was done relatively recently for someone.

I think this is the relevant ticket:
https://internal.almoctane.com/ui/?p=131002/7002#/entity-navigation?entityType=work_item&id=373244

Important to note that in the described scenario of EL-4252, tab is not moving focus away from the trigger button but rather the menu itself. This is because pressing space on the trigger moves focus to the first menu item. The title of the ticket is misleading, which is my fault.

This current request and EL-4252 seem to be complimentary: the menu should optionally close when tabbing out (obeying closeOnBlur), and the next focus target should be the focusable element that comes after the trigger in the normal tab sequence.

In terms of implementation, it's fine to change the inner workings of existing behavior, as long as the e2e tests which are already in place continue to pass without modification.

src/components/menu/menu/menu.component.ts
```diff
@@ -169,6 +169,12 @@ export class MenuComponent implements AfterContentInit, OnDestroy, OnChanges {
         this._placement$.complete();
     }
 
+    setCloseOnBlur(closeOnBlur: boolean): void {
```
donna-marijne, line 172: This doesn't need a parameter, you are checking the value of closeOnBlur twice.
```
        if (this.closeOnBlur) {
            this.menu.setCloseOnBlur(this.closeOnBlur);
        }
```

donna-marijne (changes requested):
Thanks, make sure to update relevant comments so that they remain accurate when you make changes.

donna-marijne (approved):
Verification:
https://plnkr.co/edit/qicO2cPmqiywgll2 - :heavy_check_mark:
https://plnkr.co/edit/q6Skm4pXdSE4FhHH - :heavy_check_mark:

PR #1634, opened 2022-02-23
396461 - ng2-charts v3 upgrade

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=396461

Timeline chart plugin works with v2 and v3
All documentation examples to use v3 of chart.js and ng2-charts

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_396461-ng2-charts-v3-upgrade

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~396461-ng2-charts-v3-upgrade~CI/)


src/package.json
```diff
@@ -76,7 +75,7 @@
     "@angular/compiler": ">=12.0.0 <14.0.0",
     "@angular/core": ">=12.0.0 <14.0.0",
     "@angular/forms": ">=12.0.0 <14.0.0",
-    "chart.js": ">2.9.0 <3.0.0",
-    "ng2-charts": ">=2.3.0 <3.0.0"
+    "chart.js": ">3.0.0 <4.0.0",
+    "ng2-charts": ">3.0.0 <4.0.0"
```
ashley-hunter, line 79: As it stands this is a breaking change which means this is unsuitable to be merged until UX Aspects v6 (likely May time to align when Angular 14).

donna-marijne: Is there any way to continue supporting 2.9?

JonathanToddMF: I don't think this would be possible, v3 has a lot of breaking changes. In their migration guide they mentioned it was necessary to break backwards compatibility to achieve this. https://www.chartjs.org/docs/latest/getting-started/v3-migration.html

Although that said, I can't foresee many other changes being introduced to charts in the mean time. So these changes could sit and wait till Angular 14?

ashley-hunter: Possibly, we would need to remove any reliance on types that are not part of both versions, there aren't many so in some cases, e.g. `Color` we could create our own type that has the same structure.

Also anywhere that we have had to make changes in the plugin we would need to abstract it away, we could do this with a set of utility functions (or potentially an adapter class with two implementations, one for each version), for example, something like this:

```ts
function registerPlugin(instance?: unknown): void {
    // if pluginService exists then we are in v2
    if (window.Chart.pluginService) {
       (window as any).Chart.pluginService.register(new TimelineChartPlugin());
    else {
            instance.register(new TimelineChartPlugin());
    }
}
```

But how possible this is for every case I'm not sure

donna-marijne: Yes, typescript makes it difficult but I would think that we could probe for capabilities. I would suggest evaluating this approach. Judging from the amount of time invested in this ticket, I anticipate some resistance in moving teams over to v3 even if we do introduce it in a major version increment. Since it's a hard dependency, it will force a team to upgrade their entire application to ng2-charts 3 even if they don't have any need to do so.

ashley-hunter: We may even be able to just have two different versions of the plugin the old one, and the one with the v3 changes, and the appropriate one gets registered based on detecting the version. We would have to strip out a lot of the types, but for this one file I think that shouldn't be a problem

JonathanToddMF: Thanks you both for the discussion on this. I shall make an attempt at this, there isn't a massive amount of changes made to the plugin so it may not a massive issue.

Apologies if this is obvious but in terms of the documentation examples, would it make sense to revert them back to v2, it would make reviewing these changes a lot simpler.
From my understanding the plugin should support v2 and v3 but the documentation site would have the v2 installed and it would be up to users to update their charts to v3 if they wish to use it?

donna-marijne: No, you can keep the v3 examples, it's only example code so changing it doesn't affect anyone who has already used them. The only part which really matters is the plugin.

JonathanToddMF: No worries, apologies if I added more work to this than was necessary.

src/plugins/chartjs/timeline/timeline-chart.module.ts
```diff
@@ -402,7 +410,10 @@ export class TimelineChartPlugin {
     private handleMouseMove(chart: TimelineChart, event: any): void {
         const mousePosition = this.isWithinHandle(chart, event);
 
-        const timelineOptions = chart.config.options as TimelineChartOptions;
+        let timelineOptions = chart.options as TimelineChartOptions;
+        if (this.version3Check()) {
+            timelineOptions = chart.config.options;
+        }
```
ashley-hunter, line 416: Again we can simplify this:

```ts
const timelineOptions = (this.isVersion3() ? chart.config.options : chart.options) as TimelineChartOptions;
```

src/plugins/chartjs/timeline/timeline-chart.module.ts
```diff
@@ -772,12 +781,29 @@ export class TimelineChartPlugin {
...
+
+            // get the range on the x-axis
+            minimum = data.x.min;
+            maximum = data.x.max;
+        } else {
+
+            // get the current data
+            const { data } = chart.getDatasetMeta(0);
+
+            console.log(data);
```
ashley-hunter: We can remove this

ashley-hunter (changes requested):
Thanks, looks good, just had a few small suggestions

src/package.json
```diff
@@ -76,7 +75,7 @@
     "@angular/compiler": ">=12.0.0 <14.0.0",
     "@angular/core": ">=12.0.0 <14.0.0",
     "@angular/forms": ">=12.0.0 <14.0.0",
-    "chart.js": ">2.9.0 <3.0.0",
-    "ng2-charts": ">=2.3.0 <3.0.0"
+    "chart.js": ">2.0.0 <4.0.0",
+    "ng2-charts": ">2.0.0 <4.0.0"
```
donna-marijne, line 79: Was something changed to support older versions of these dependencies as well? If not, I would suggest maintaining the previous lower bounds of these ranges.

donna-marijne (changes requested):
Thanks for the PR. I have tried the updated timeline chart example in a standalone application and there are a few compilation errors:
```
Error: src/app/app.component.html:5:6 - error TS2322: Type 'ChartDataset<keyof ChartTypeRegistry, (number | ScatterDataPoint | BubbleDataPoint | null)[]>[]' is not assignable to type 'ChartDataset<"line", (number | ScatterDataPoint | BubbleDataPoint | null)[]>[]'.

5     [datasets]="lineChartData"
       ~~~~~~~~

  src/app/app.component.ts:15:16
    15   templateUrl: './app.component.html',
                      ~~~~~~~~~~~~~~~~~~~~~~
    Error occurs in the template of component AppComponent.


Error: src/app/app.component.ts:21:3 - error TS2564: Property 'baseChart' has no initializer and is not definitely assigned in the constructor.

21   baseChart: BaseChartDirective;
     ~~~~~~~~~


Error: src/app/app.component.ts:25:3 - error TS2564: Property 'lineChartLabels' has no initializer and is not definitely assigned in the constructor.

25   lineChartLabels: string[];
     ~~~~~~~~~~~~~~~


Error: src/app/app.component.ts:28:3 - error TS2564: Property 'timelineChartLabels' has no initializer and is not definitely assigned in the constructor.

28   timelineChartLabels: string[];
     ~~~~~~~~~~~~~~~~~~~


Error: src/app/timeline-chart.service.ts:8:21 - error TS2503: Cannot find namespace 'Chart'.

8   private _dataset: Chart.ChartPoint[] = [];
                      ~~~~~


Error: src/app/timeline-chart.service.ts:40:50 - error TS2454: Variable 'value' is used before being assigned.

40       this._dataset.push({ x: new Date(date), y: value });
                                                    ~~~~~


Error: src/app/timeline-chart.service.ts:45:17 - error TS2503: Cannot find namespace 'Chart'.

45   getDataset(): Chart.ChartPoint[] {
                   ~~~~~
```

Might be worth checking the other updated examples as well, if you haven't already.

docs/app/services/playground/utilities/system-helper.ts
```diff
@@ -59,9 +59,12 @@ export class SystemJSHelper {
         { name: 'rxjs', path: 'rxjs@6.6.0' },
         { name: 'ngx-bootstrap', path: `ngx-bootstrap@${ NGX_BOOTSTRAP_VERSION }/bundles/ngx-bootstrap.umd.min.js` },
         { name: 'chance', path: 'chance' },
-        { name: 'chart.js', path: 'chart.js@2.9.3/dist/Chart.bundle.min.js' },
-        { name: 'ng2-charts', path: 'ng2-charts@2.3.0/bundles/ng2-charts.umd.js' },
+        { name: 'chart.js', path: 'chart.js@3.7.1/dist/chart.min.js' },
+        { name: 'ng2-charts', path: 'ng2-charts@3.0.8/fesm2020/ng2-charts.mjs' },
         { name: 'lodash', path: 'lodash/lodash.js' },
+        { name: 'lodash-es', path: 'lodash-es@4.17.21/lodash.js' },
```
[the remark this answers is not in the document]
JonathanToddMF: I tried removing the lodash on line 64 to see. It looks like the plunkers continue to work.

Would this also mean removing the package from the package.json aswell?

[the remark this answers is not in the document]
donna-marijne: Yes, I thought we stopped using lodash a while back. I've had a quick look at it seems like we're only using the types package, which is probably useless on its own and should be removed.
https://github.com/UXAspects/UXAspects/blob/master/package.json#L93

Regarding `lodash-es`, is this being used in one of the updated plunker examples?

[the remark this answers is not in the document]
ashley-hunter: Chart.js 3 has a dependency on lodash-es which is why it was added. If we don't need lodoash then we can remove it, but this is only a list of where to find the files if a dependency is requested so it won't be loaded unless it is imported by something, so if it is safer to keep lodash in there is won't have any negative impact on load times.

donna-marijne (changes requested):
Thanks, it looks good, there is some use of the definite assignment operator here which I consider to be generally harmful.

donna-marijne (approved):
Verification:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_396461-ng2-charts-v3-upgrade/#/charts/timeline-chart#timeline-chart - :heavy_check_mark:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/caf_CI_ux-aspects-micro-focus_396461-ng2-charts-v3-upgrade/#/ui-components/charts#timeline-chart - :heavy_check_mark:

Also verified in a standalone app with ng2-charts@2.

ashley-hunter (approved):
Thanks, looks good to me too!

Roland-Penny (approved):
Looks good

PR #1645, opened 2022-04-01
410022 - codesandbox

donna-marijne:
#### Checklist

#### Ticket / Issue
https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=410022

- rework the playground service to build and launch a codesandbox instance.
  - it is template based so whenever a new version of Angular comes out we can download updated starter templates from codesandbox and paste them into the source tree 
  - there are transformers similar to the typescript API to modify the template files according to each playground
  - the codesandbox projects now have some metadata including the active theme and the branch they were taken from, this should help to avoid using the wrong code when doing QA or sharing examples on the channel.
  - most of the examples use Angular 13 in codesandbox but codesandbox can't currently process ViewEngine dependencies with Angular 13, so in a few cases the project has to be downgraded to Angular 12.
- update dev build process to bundle and serve a tgz package of the library instead of the plunker assets.
  - there is a watch script to repackage whenever the code is changed.
  - added some color/formatting since there is a huge wall of logging output from npm start now.
- update the production build to copy in a tgz package of the library into the assets dir
  - in future we should use Artifactory for this, it would be ideal since it already hosts the tgz file and would also allow consumers to create codesandbox examples for past releases.
  - the plunker assets remain so that existing plunkers will still work after this change
- removed `IconModule`, `AccessibilityModule`, and `ColorServiceModule` from many playground definitions since these are loaded in every example by default.
- fixed a number of missing assets and other issues in various playgrounds.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_410022-codesandbox

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~410022-codesandbox~CI/)


JonathanToddMF (approved):
Changes work well. Tested all sections to ensure they run. Played around with new scripts in the package.json. Everything works as expected.

Roland-Penny (approved):
Looks good

JonathanToddMF:
Verification

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_410022-codesandbox/#/components - ✔️
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/caf_CI_ux-aspects-micro-focus_410022-codesandbox/#/ui-components - ✔️

PR #1646, opened 2022-04-07
373587 - Select: empty string does not work as an option

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=373587

- Added explicit null checks to singleOptionSelected and to the addTag functions.
- Added height to typeahead list element, this was preventing empty options from showing on the outer source, not an issue in the inner source.
- Added styling to center icons with no string in outer source, not an issue in the inner source
- Added e2e tests as the issue couldn't be tested programmatically through karma tests

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_373587-empty-string-select

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~373587-empty-string-select~CI/)


src/components/tag-input/tag-input.component.html
```diff
@@ -5,6 +5,7 @@
     (click)="toggle()">
 
     <li *ngFor="let tag of _tags; let i = index" class="ux-tag"
+        [class.ux-tag-empty]="!tag"
```
donna-marijne, line 8: This class makes the delete button a bit wonky when compared to normal tags.

I do not believe that teams will actually use an empty string as the display value, since this would be very confusing to the user in both single and multiple select modes. I expect it will rather be something like this:
https://plnkr.co/edit/vSnRsqRFARUd9kBT

So no special handling is required for "empty tags" on our end, in my opinion.

donna-marijne (changes requested):
Thanks for the PR, I think it looks good. I just had one suggestion.

donna-marijne (approved):
Verification:

I have tested with a function that renders a placeholder value to the user when an empty string is selected, because there are significant usability issues around showing a value which has no description at all.

https://plnkr.co/edit/vSnRsqRFARUd9kBT - :heavy_check_mark:
https://plnkr.co/edit/s8LFCQGHctP1dzVY - :heavy_check_mark:

PR #1650, opened 2022-05-03
436153 - ngx bootstrap 8

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=436153

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_436153-ngx-bootstrap-8

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~436153-ngx-bootstrap-8~CI/)


donna-marijne (changes requested):
I see that you have updated the CSS for the documentation example, which presumably means that consumers will have to copy and paste this rule every time. I think it would be better to incorporate this into our library stylesheet, unless there is some reason not to.

Also, we must not use `!important` or recommend consumers to use it. It causes an arms race if someone needs to override it down the line. Always use a selector of proper specificity to override according to the cascade rules.

donna-marijne:
Thanks Jonathan. I think the third option is preferable and more sustainable for the future. You can clean it up a bit:
https://codesandbox.io/s/color-picker-ux-aspects-forked-63rqiy?file=/src/app/app.component.ts

I expect that we will be able to replace the dropdownMenu directive with ux-menu in the other documentation examples as well. No point continuing to use a problematic 3rd party component when we have our own fully functional equivalent.

e2e/tests/components/color-picker/color-picker.e2e-spec.ts
```diff
@@ -0,0 +1,16 @@
+import { imageCompare } from '../common/image-compare';
+import { ColorPickerPage } from './color-picker.po.spec';
+
+describe('Color Picker Tests', () => {
```
donna-marijne, line 4:
```suggestion
describe('Color Picker', () => {
```

donna-marijne (changes requested):
Thanks for the update. I have some feedback.

Would you avoid reformatting existing source files unless they are significantly malformatted? It makes it hard to identify relevant changes when so many lines have been pointlessly rewrapped. On top of this, it makes git's [blame feature](https://github.com/UXAspects/UXAspects/blame/50aae292c1c9dcca3d859a912d227d098fa67656/docs/app/pages/components/components-sections/tables/sorting/sorting.component.html) essentially unusable, and we do rely on this from time to time. I'm not sure if VS Code formats on save by default now; if it does, please make sure to disable it for your UX Aspects workspace. Lastly, some of the formatting changes here have negatively impacted the readability of the code, so please review and revert those unnecessary changes.

In testing I noticed that the "Actions" card no longer works in the Drag & Drop Cards documentation example.
image

Tested with Firefox on Windows.

docs/app/pages/components/components-sections/color/color-picker/color-picker.component.html
```diff
@@ -1,53 +1,42 @@
...
 
     <span class="m-l-lg">
-        Selected: <code>{{selected.hex}}</code> &mdash; <code>{{selected.rgba}}</code>
+        Selected: <code>{{ selected.hex }}</code> &mdash;
+        <code>{{ selected.rgba }}</code>
     </span>
 </div>
 
+<ux-menu #dropdownMenu id="color-picker-dropdown">
+    <div uxMenuItem [closeOnSelect]="false">
```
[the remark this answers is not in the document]
JonathanToddMF: I tried using this but in the content of the application there are some issues with tabbing into the dropdown. It seems if there are other elements in the tabindex it will prioritise those over the dropdown - https://codesandbox.io/s/color-picker-ux-aspects-forked-8fdz3l?file=/src/app/app.component.html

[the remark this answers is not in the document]
donna-marijne: Sure, but this is a hack - you can see that the background color changes, and the pointer cursor is active on the entire picker. If you're going to use ux-menu here then it needs to be an adequate replacement without degrading the user experience. Maybe you need to modify ux-menu so that it works properly with custom content as well as with menu items, for example.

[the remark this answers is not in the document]
JonathanToddMF: Agreed, I had a play around with this using the `uxMenuTabbableItem` directive. It appears to work quite well as it can be applied to any element. The only thing about it is the extra tab step but it doesn't have the background or clickable background issue when using the `uxMenuItem` directive. 

https://codesandbox.io/s/color-picker-ux-aspects-forked-8fdz3l?file=/src/app/app.component.html

Just thought I should get an opinion on this before making any changes to the ux-menu.

src/components/color-picker/color-picker.component.ts
```diff
@@ -106,6 +106,10 @@ export class ColorPickerComponent implements OnInit, OnDestroy {
     @Input()
     inputAriaLabel: (mode: ColorPickerInputMode) => string = this.getInputAriaLabel;
 
+    /** Whether or not to forward focus to the selected or first button in the color picker. */
+    @Input()
+    forwardFocusToButton: boolean = false;
```
donna-marijne, line 111: I think this should be the default behavior, no need for configuration. I don't see any need for landing focus on the wrapper.

docs/app/pages/components/components-sections/color/color-picker/color-picker.component.html
```diff
@@ -1,53 +1,44 @@
...
-        Selected: <code>{{selected.hex}}</code> &mdash; <code>{{selected.rgba}}</code>
+        Selected: <code>{{ selected.hex }}</code> &mdash;
+        <code>{{ selected.rgba }}</code>
     </span>
 </div>
 
+<ux-menu #dropdownMenu id="color-picker-dropdown">
+    <ux-color-picker
+        uxMenuTabbableItem
+        [type]="menuItemType"
```
donna-marijne, line 24: What is this input? It's not documented as far as I can see. Can we do without it?

src/components/menu/menu-tabbable-item/menu-tabbable-item.directive.ts
```diff
@@ -15,7 +15,7 @@ export class MenuTabbableItemDirective implements OnInit, OnDestroy, FocusableOp
     @Input() disabled: boolean = false;
 
     /** Indicate the type of the menu item */
-    readonly type: MenuItemType = MenuItemType.Default;
+    @Input() type: MenuItemType = MenuItemType.Default;
```
donna-marijne, line 18: Avoid adding directive inputs with the same name as an HTML attribute.

donna-marijne (changes requested):
Thanks for the update, one thing I noticed is that the focus disappears when opening the color picker using the keyboard.
[Image from Gyazo](https://gyazo.com/3563f2ad334e0a29ac18669e0e9a676e)

Also I think we just need a very simple convenience directive for placing first focus when the menu opens, instead of trying to wedge it in to an existing directive that we would otherwise not be using. E.g.
```html
<ux-menu #dropdownMenu>
  <ux-color-picker uxMenuInitialFocus ...>
</ux-menu>
```

That would be a lot more self-explanatory than this `type` input.

@ashley-hunter just wanted to check if you have any suggestions on this, in case we've overlooked something that already exists.

ashley-hunter:
> @ashley-hunter just wanted to check if you have any suggestions on this, in case we've overlooked something that already exists.

The only thing I think we currently have is a `focusIf` directive, so potentially that may be an option. The only other thing is the CDK comes with a `cdkFocusInitial` directive which might be possible to use: https://material.angular.io/cdk/a11y/overview#regions

But alternatively your suggestion of a menu directive seems good too.

donna-marijne:
I pushed the solution I've been suggesting. It's simple and saves the consumer having to wire up a whole bunch of inputs and outputs.

src/components/menu/menu-initial-focus/menu-initial-focus.directive.ts
```diff
@@ -0,0 +1,29 @@
...
+        private readonly _menu: MenuComponent,
+        private readonly _elementRef: ElementRef<HTMLElement>,
+        private readonly _renderer: Renderer2
+    ) {}
+
+    ngOnInit(): void {
+        this.ensureFocusable();
+
+        // Focus the host element when the parent menu is opened.
+        this._menu.opened.pipe().subscribe(() => {
```
ashley-hunter, line 16: Think we need to unsubscribe from this observable 😊

JonathanToddMF: I can make this change, thanks Ashley

donna-marijne (commented):
I pushed up some code review feedback.

JonathanToddMF:
Had a look through the changes, everything looks in order

donna-marijne (approved):
Verification:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_436153-ngx-bootstrap-8/#/components/color#color-picker - :heavy_check_mark:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_436153-ngx-bootstrap-8/#/components/drag-and-drop#drag-and-drop-cards - :heavy_check_mark:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_436153-ngx-bootstrap-8/#/components/notifications#notifications - :heavy_check_mark:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_436153-ngx-bootstrap-8/#/components/page-header#page-header - :heavy_check_mark:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_436153-ngx-bootstrap-8/#/components/tables#custom-filters - :heavy_check_mark:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_436153-ngx-bootstrap-8/#/components/tables#sorting - :heavy_check_mark:

Roland-Penny (approved):
Accepted based on dev QA feedback

PR #1652, opened 2022-05-05
374510 - Radio Button - Inaccessible by keyboard navigation with no option selected

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=374510

- Added 0 tabindex to inputs this reverts the functionality to the same as html radio buttons
- Updated e2e tests to cover keyboard accessibility with no selection
- Updated screenshots with new button

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_374510-radio-button-keyboard

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~374510-radio-button-keyboard~CI/)


src/components/radiobutton/radiobutton.component.html
```diff
@@ -1,19 +1,20 @@
...
-        <input class="ux-radio-button-input"
+        <input
+            class="ux-radio-button-input"
             uxFocusIndicator
             type="radio"
             [id]="(id || _radioButtonId) + '-input'"
             [checked]="value === option"
             [disabled]="disabled"
-            [tabindex]="tabindex || value === option ? 0 : -1"
+            [tabindex]="0"
```
donna-marijne, line 17: Is the `tabindex` input used for anything now that it is not part of this expression?

donna-marijne (changes requested):
It looks like the second part of this problem still exists (when the selected option is disabled).
radio-button-focus

Also, we are using managed focus for radio buttons, which means that only one button in the group should be focusable.

https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/examples/radio/radio.html

src/components/radiobutton/radiobutton.component.html
JonathanToddMF: @donna-marijne Thanks for the feedback. Good point I will remove this. I also think explicitly setting the tabIndex here isn't required and without it the tabIndex of our radio buttons will perform as specified in the above w3 link. Only one button is focusable at the start or end depending on the direction the tab comes from with each option being selectable via the arrow keys.

Apologies on the second issue I missed it. With regards to it. I played around with this for awhile and it seems to be an issue (or correct behaviour) of all radio buttons. I checked with default html and Angular Material's implementation and they all have this same issue, which would lead me to believe this is correct behaviour.

As a workaround I added a readonly input which allows this functionality to work. Although there may be a slightly hacky solution that I could investigate further

donna-marijne: Well, we can't remove an input without deprecating it first and giving proper notice.

@ashley-hunter Do you remember why a tabindex input was added?

donna-marijne:
If there's no solution for the second issue then we will need to document a workaround. It's certainly unacceptable to have a control which can only be operated with a mouse. One option could be to remove the selected state from a radio before disabling it. This would not be built-in functionality (this would be a breaking change), but rather something that we show how to do in the documentation example.

src/components/radiobutton/radiobutton.component.html
```diff
@@ -1,19 +1,20 @@
...
-        <input class="ux-radio-button-input"
+        <input
+            class="ux-radio-button-input"
             uxFocusIndicator
             type="radio"
             [id]="(id || _radioButtonId) + '-input'"
             [checked]="value === option"
             [disabled]="disabled"
-            [tabindex]="tabindex || value === option ? 0 : -1"
+            [readonly]="readonly"
```
donna-marijne, line 17: `readonly` is not a valid attribute on `input type="radio"`.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly

donna-marijne (changes requested):

src/components/radiobutton/radiobutton.component.html
ashley-hunter: No I don't recall why it was added. It looks like it was an attempt to add roving focus support, but I believe the browser handles that in radio groups anyway so not required.

donna-marijne: As discussed, let's mark the tabindex input `@deprecated` so that it can be removed in the next major release.

ashley-hunter: If you find there is absolutely no behaviour effect of tabindex (e.g. even without the name). It would not be a breaking change to remove the input as tabindex is a property available on all html elements, so it would not cause a build to fail by removing the input. You would just check there is no unexpected behaviour change.

docs/app/pages/components/components-sections/input-controls/radio-button/radio-button.component.html
```diff
@@ -98,10 +101,36 @@
...
+</p>
+
+<h4>Keyboard Accessibility</h4>
+<p>
+    Due to an issue within native HTML it is impossible to focus any of the
+    <code>ux-radio-buttons</code> while one is <code>selected</code> and <code>disabled</code> when
+    using the keyboard. As a workaround we recommend removing the <code>selected</code> state from
+    the <code>ux-radio-button</code> when disabling it.
 </p>
+<p>Below is a code example:</p>
```
donna-marijne, line 116: Sorry for the misunderstanding, if you have a workaround could you integrate it into the working example in our docs? I don't want any of our examples to be keyboard inaccessible.

donna-marijne (changes requested):

docs/app/pages/components/components-sections/input-controls/radio-button/radio-button.component.ts
```diff
@@ -42,4 +42,11 @@ export class ComponentsRadioButtonComponent extends BaseDocumentationSection imp
         super(require.context('./snippets/', false, /\.(html|css|js|ts)$/));
     }
 
+    disable(radio: number | string | object): void {
```
donna-marijne, line 45:
```suggestion
    toggleDisabled(radio: number | string | object): void {
```

donna-marijne (changes requested):
Thanks for the update, this works well except for one issue. When no radio is checked, all of the radios are focusable. Focus should be managed within the radio group, so if none are checked only the first non-disabled radio should be focusable.

src/components/radiobutton/radio-button-group/radio-button-group.directive.ts
```diff
@@ -95,10 +95,28 @@ export class RadioButtonGroupDirective<T = any> implements ControlValueAccessor,
...
+    /** Determine and set the correct tabindex */
+    determineAndSetTabIndex() {
+
+        const firstEnabled = this._radioButtons.find(radio => {
+            return radio.disabled === false;
+        });
+
+        this._radioButtons.forEach(radio => {
+
+            if (this._value) {
```
donna-marijne, line 107: This isn't a good way to check for an undefined user-provided value, someone might need to use zero or null or an empty string as a radio option. This same issue was reported very recently for the select component, so make sure to keep a note of recurring problems like this.

src/components/radiobutton/radiobutton.component.html
```diff
@@ -1,19 +1,20 @@
...
-        <input class="ux-radio-button-input"
+        <input
+            class="ux-radio-button-input"
             uxFocusIndicator
             type="radio"
             [id]="(id || _radioButtonId) + '-input'"
             [checked]="value === option"
             [disabled]="disabled"
-            [tabindex]="tabindex || value === option ? 0 : -1"
+            [tabindex]="tabindex || internalTabindex"
```
donna-marijne, line 17:
```suggestion
            [tabindex]="tabindex ?? internalTabindex"
```

Probably shouldn't override the tabindex if it is explicitly set to 0.

src/components/radiobutton/radiobutton.component.ts
```diff
@@ -33,8 +33,11 @@ export class RadioButtonComponent<T = any> implements ControlValueAccessor {
     /** Specify if this is a required input */
     @Input() required: boolean;
 
-    /** Specify the tabindex */
-    @Input() tabindex: number = 0;
+    /**
+     * Specify the tabindex
+     * @deprecated This input is deprecated and will be removed in the next major release.
+     */
```
donna-marijne, line 39: I noticed that you have reinstated `tabindex` in the template. Is it still appropriate to remove this input?

docs/app/pages/components/components-sections/input-controls/radio-button/radio-button.component.html
```diff
@@ -102,6 +102,15 @@
     the radio button with a <code>name</code> or <code>id</code> if required.
 </p>
 
+<h4>Keyboard Accessibility</h4>
+<p>
+    Due to an issue within native HTML it is impossible to focus any of the
```
donna-marijne, line 107: This sentence is a bit confusing, also HTML is a spec so it's more on the browsers. You can just say that it's not possible to land keyboard focus on a radio button group when the selection option is disabled.

e2e/tests/components/radiobuttons/radiobuttons.e2e-spec.ts
```diff
@@ -124,4 +124,14 @@ describe('RadioButton Tests', () => {
         expect(await page.text1.getText()).toBe('string');
 
     });
+
+    it('should allow selection via the keyboard when no radio button is selected', async () => {
```
donna-marijne, line 128: There were two issues reported in the ticket, and QA feedback during the review process. Make sure you are capturing these use cases as tests!

src/components/radiobutton/radiobutton.component.ts
```diff
@@ -126,4 +138,9 @@ export class RadioButtonComponent<T = any> implements ControlValueAccessor {
         this.disabled = isDisabled;
         this._changeDetector.markForCheck();
     }
+
+    setTabIndex(tabIndex): void {
```
donna-marijne, line 142:
```suggestion
    setInternalTabindex(tabIndex): void {
```
Or something else, since this component now has two tabindex properties. Also, add descriptions to your functions.

donna-marijne (changes requested):
Thanks for the update.

src/components/radiobutton/radiobutton.component.ts
JonathanToddMF: Yeah, ideally the tabindex will be handled by the `_internalTabIndex` but if anyone wishes to use this input for now it can remain deprecated but it really isn't that useful, now the original issues have been resolved.

donna-marijne (approved):
Verification:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_374510-radio-button-keyboard/#/components/input-controls#radio-button - :heavy_check_mark:
https://pages.github.houston.softwaregrp.net/sepg-docs-qa/caf_CI_ux-aspects-micro-focus_374510-radio-button-keyboard/#/ui-components/input-controls#radio-button - :heavy_check_mark:

Roland-Penny (approved):
Looks good

PR #1654, opened 2022-05-11
372601 - Slider: Cannot change values when low and high values are the same max value

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=372601

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_372601-slider-low-high

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~372601-slider-low-high~CI/)

### DIFF
--- a/e2e/tests/components/sliders/sliders.e2e-spec.ts
+++ b/e2e/tests/components/sliders/sliders.e2e-spec.ts
@@ -344,6 +344,20 @@ describe('Sliders', () => {
 
     });
 
+    it('should allow dragging of the lower thumb when both values are set to the max', async () => {
+
+        page.input2.clear();
+        page.input2.sendKeys('100' + Key.ENTER);
+
+        page.input1.clear();
+        page.input1.sendKeys('100' + Key.ENTER);
+
+        page.dragAndDropHandle(page.rangeWithTextInputs, 'lower', { x: -30, y: 0 });
+        expect(await page.getTooltipValue(page.rangeWithTextInputs, 'lower')).toEqual('95');
+        expect(await page.getInputValue(page.input1)).toEqual('95');
+
+    });
+
     describe('in disabled state', () => {
         beforeEach(async () => {
             await page.disabledButton.click();
--- a/src/components/slider/slider.component.ts
+++ b/src/components/slider/slider.component.ts
@@ -273,6 +273,7 @@ export class SliderComponent implements OnInit, AfterViewInit, DoCheck {
 
         // update the thumb state
         this.setThumbState(thumb, state.hover, state.drag);
+        this.updateOrderAtMax();
 
     }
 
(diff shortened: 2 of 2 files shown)


donna-marijne (changes requested):
Thanks for the PR. Looks like there are still some issues when the handles are very close to each other but not the same.
https://codesandbox.io/s/sliders-ux-aspects-forked-f4ulhh?file=/src/app/app.component.html

You might check with @Roland-Penny about how to solve this UX issue.

JonathanToddMF:
Checked with @Roland-Penny he is happy with the solution implemented, ready for review

src/components/slider/slider.component.ts
```diff
@@ -745,6 +746,19 @@ export class SliderComponent implements OnInit, AfterViewInit, DoCheck {
         return destination;
     }
 
+    private updateOrderAtMax(): void {
+        const lowerValue = this.getThumbValue(this.sliderThumb.Lower);
+        const upperValue = this.getThumbValue(this.sliderThumb.Upper);
+        const max = this._options?.track?.max;
+        const maxBuffer = max / 100 * 95;
```
donna-marijne, line 753: This doesn't consider the min value, so you can get stuck at the lower bound as well. Check the last example here:
https://codesandbox.io/s/sliders-ux-aspects-forked-f4ulhh?file=/src/app/app.component.html

e2e/tests/components/sliders/sliders.e2e-spec.ts
```diff
@@ -344,6 +344,20 @@ describe('Sliders', () => {
 
     });
 
+    it('should allow dragging of the lower thumb when both values are set to the max', async () => {
```
donna-marijne, line 347: We're going to need more test coverage for this. I linked a sandbox previously with some test cases, that should be a minimum. You can set up a new test page with more numerical inputs to configure min and max and so on.

donna-marijne (changes requested):
Thanks, this works pretty well for most cases, however at smaller widths it can still be a bit difficult.
[Image from Gyazo](https://gyazo.com/079185b12bcb15a63942a19bb7127fac)

Instead of picking a magic percentage, consider making the breakpoint the mid-point of the slider instead.
- If the upper slider is below the 50% mark of the track, it goes on top.
- If the lower slider is above the 50% mark of the track, it goes on top.

src/components/slider/slider.component.ts
```diff
@@ -453,14 +455,39 @@ export class SliderComponent implements OnInit, AfterViewInit, DoCheck {
 
     }
 
-    private updateOrder(thumb: SliderThumb): void {
+    private updateOrder(thumb?: SliderThumb): void {
```
donna-marijne, line 458: You've packed another function inside this one, controlled by whether the parameter is provided. These are called from different places in the component, so there is no need for the function to figure out which of two completely distinct code paths to take. It's cleaner and more maintainable to have functions do one thing. An easy way to spot this pattern is when the function body is almost entirely inside an if/else statement.

donna-marijne (changes requested):
Seems to work well, and the code coverage is improved. I have noticed one code hygiene issue.

donna-marijne (approved):
Verification:
https://codesandbox.io/s/sliders-ux-aspects-forked-f4ulhh?file=/src/app/app.component.html - :heavy_check_mark:
https://codesandbox.io/s/slider-ux-aspects-micro-focus-forked-xnwk0g?file=/src/app/app.component.ts - :heavy_check_mark:

Roland-Penny (approved):
Looks good to me

PR #1658, opened 2022-05-30
464101 - Tag Input - Console error when enforceTagLimits is at maxTag limit when tabbing through tags

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=464101

1. ISSUE: There is a console error when tabbing through the component when there are max tags and enforce tag limits is true. The input is removed when this parameter is met and therefore this.tagInput is undefined in the keyHandler function.
2. FIX: Making problem parameters optional.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_464101-tag-input-console-error

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~464101-tag-input-console-error~CI/)

### DIFF
--- a/src/components/tag-input/tag-input.component.ts
+++ b/src/components/tag-input/tag-input.component.ts
@@ -209,7 +209,7 @@ export class TagInputComponent<T = any> implements AfterContentInit, OnChanges, 
 
     @ContentChildren(TypeaheadComponent) typeaheadQuery: QueryList<TypeaheadComponent>;
 
-    @ViewChild('tagInput', { static: false }) tagInput: ElementRef;
+    @ViewChild('tagInput', { static: false }) tagInput: ElementRef<HTMLInputElement>;
 
     selectedIndex: number = -1;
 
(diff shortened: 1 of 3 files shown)


src/components/tag-input/tag-input.component.ts
```diff
@@ -336,10 +336,10 @@ export class TagInputComponent<T = any> implements AfterContentInit, OnChanges,
         }
 
         // Get the input field cursor location
-        const inputCursorPos = this.tagInput.nativeElement.selectionStart;
+        const inputCursorPos = this.tagInput?.nativeElement.selectionStart;
```
ashley-hunter, line 339: Could we update the ViewChild statement to include the correct generic type. Currently typescript is inferring `inputCursorPos` as `any` rather than `number`.

Also, right now whenever there is no input `inputCursorPos` is `undefined`, however later we are assuming it is a number, e.g.:

```ts
inputCursorPos <= 0
```

This seems like it might be a bit unsafe, it might be good to supply a default value for `inputCursorPos` whenever the input element is undefined. (I'm not sure what the correct value should be, but something like this):

```ts
const inputCursorPos = this.tagInput?.nativeElement.selectionStart ?? 0;
```

Alternatively we could update `canNavigateLeft` and `canNavigateRight` to check if `inptuCursorPos` is `undefined` before using it.

ashley-hunter (changes requested):
Thanks! Works well! I just had one suggestion on how we can improve some type safety to help prevent any future similar issues.

Roland-Penny (approved):
Accepted on dev QA feedback

PR #1663, opened 2022-06-14
Angular 14

donna-marijne:
#### Checklist

#### Ticket / Issue
https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=474061

- update to Angular 14 and run migrations
- switch to Ashley's custom builder to continue building the library with Angular 13 tools, this is because libraries created with Angular 14 are not backward compatible
- remove Angular 12 support
- correct some outdated information in CONTRIBUTING
- update to the new build scripts, this exposes the ability to package @ux-aspects/ux-aspects-docs without scripts
- remove documentation references to Angular 12 and Internet Explorer
- fix a media player type error
- remove some leftover code from protractor-retry

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_474061-angular-14

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~474061-angular-14~CI/)


JonathanToddMF:
Thanks for these changes Donna, all is looking good so far I haven't encountered any problems bar one. 

I think the peerDependencies need to be bumped for v13 and v14 the library

JonathanToddMF (dismissed):
Hi Donna, everything looks great with this I tested it out locally and everything seems in order.

Just one thing I noticed, NGX-Bootstrap 8 doesn't officially support Angular 14, doesn't seem to cause any issues but just though we should be aware- https://valor-software.com/ngx-bootstrap/#/documentation#compatibility

@ashley-hunter I don't know if you want to take a look too!

ashley-hunter:
Looks fine to me, the only thing we may want to change all the uses of Untyped form controls to the type safe FormControl. Angular 13 doesn't have any UntypedForm control types, so any sample code that uses them will not work in Angular 13.

donna-marijne:
> Looks fine to me, the only thing we may want to change all the uses of Untyped form controls to the type safe FormControl. Angular 13 doesn't have any UntypedForm control types, so any sample code that uses them will not work in Angular 13.

I might be wrong, but the Angular 13 types don't seem to have a generic type parameter. I think this feature was pushed out entirely within Angular 14. So I'm not sure how to update e.g. the number picker example with example code that would work in both.

ashley-hunter:
Hmm, good point. I think its probably fine then. Its easy to explain if someone has any issues.

donna-marijne:
I'll update the examples which strictly use strings, should reduce the chance of someone having trouble.

donna-marijne:
Seems like they could all be updated, the only ambiguous usage of FormControl was in the unit tests.

ashley-hunter (approved):
Thanks Donna, looks good to me!

JonathanToddMF (approved):
Everything looks good to me!

JonathanToddMF:
Just double checking package on ng13 project before merging

PR #1667, opened 2022-07-13
507155 - Accessibility part 1

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=507155

1. Date Time Picker: switched unsuitable use of aria-selected to aria-pressed. Seems other date pickers such as Angular materials use this approach
2. Removed problem uses of aria-label. I could not find a suitable role for these elements. Open to any feedback.
3. Date Time Picker: Noticed issue with aria-valuenow in the timezone time view, this expects a numerical value as opposed to a string. valuetext can be used here instead.
4. Tested each component in a standalone application to check AXE reports

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_507155-accessibility-part-1

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~507155-accessibility-part-1~CI/)

### DIFF
--- a/docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
+++ b/docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
@@ -113,10 +113,10 @@
         Aria label of the filter field. If not specified, the placeholder will be used.
     </tr>
     <tr uxd-api-property name="searchFilterButtonAriaLabel" type="string">
-        Aria label of the Search filter field icon.
+        Aria label of the filter field search icon.
     </tr>
     <tr uxd-api-property name="clearFilterButtonAriaLabel" type="string">
-        Aria label of the Clear filter field icon.
+        Aria label of the filter field clear icon.
     </tr>
     <tr uxd-api-property name="dropdownOpen" type="boolean">
         If set to <code>true</code> this will programmatically show the typeahead dropdown.
--- a/docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
+++ b/docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
@@ -9,7 +9,8 @@
             [(filter)]="filter"
             [allowNull]="allowNull"
             [maxHeight]="maxHeight"
-            [placeholder]="placeholder">
+            [placeholder]="placeholder"
+            [aria-label]="'Filter input'">
 
             <ng-template #displayContent>
                 <b>Selection:</b> {{ selected ? selected.name : '(none)' }}
--- a/docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
+++ b/docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
@@ -5,12 +5,14 @@
             [(selected)]="selected"
             [dropdownOpen]="dropdownOpen"
             (dropdownOpenChange)="dropdownOpenChange($event)"
+            (filterChange)="setFilterIconAriaLabel($event)"
             [disabled]="disabled"
             [(filter)]="filter"
             [allowNull]="allowNull"
             [maxHeight]="maxHeight"
             [placeholder]="placeholder"
-            [aria-label]="'Filter input'">
+            [ariaLabelIcon]="ariaLabelIcon"
+            aria-label="Filter input">
 
             <ng-template #displayContent>
                 <b>Selection:</b> {{ selected ? selected.name : '(none)' }}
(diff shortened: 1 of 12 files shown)


src/components/date-time-picker/day-view/day-view.component.html
```diff
@@ -19,7 +19,7 @@
                         [class.range-end]="isRangeEndDate(item.date)"
                         [focusIf]="shouldFocus(item)"
                         [attr.aria-label]="item.date | date"
-                        [attr.aria-selected]="isItemActive(item.date, item.isActive)"
```
donna-marijne, line 22: `aria-selected` is fine to use on an element with role="gridcell". I guess the issue here is that the role is on the parent element (`td`) here.

https://www.w3.org/TR/wai-aria/#aria-selected

The date picker buttons are not toggle buttons, so I don't think `aria-pressed` is correct.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -9,8 +9,7 @@
             [(filter)]="filter"
             [allowNull]="allowNull"
             [maxHeight]="maxHeight"
-            [placeholder]="placeholder"
-            aria-label="Filter input">
```
donna-marijne, line 13: I would say that the `aria-label` attribute should be forwarded to the `button` element of `ux-input-dropdown` along with `aria-labelledby`. Buttons can certainly have an aria-label.

donna-marijne (changes requested):
Thanks for the update.

I saw this note in the issue, possibly overlooked:

> Input Dropdown
>    Filter button needs aria-label

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
JonathanToddMF: Is there a way a forward this without creating a new input? There is already an aria-label input but that is for the filter input element.

Unless I create a second input called `ariaLabelButton` and document it? What are you thoughts on this.

donna-marijne: In that case, you'll still need to fix this issue rather than remove the aria-label from the example (the issue still exists, you've just hidden it). It is certainly wrong that the aria-label is forwarded to the filter input, but we can't change that without breaking existing installations.

docs/app/pages/components/components-sections/input-controls/input-dropdown/input-dropdown.component.html
```diff
@@ -10,7 +10,7 @@
             [allowNull]="allowNull"
             [maxHeight]="maxHeight"
             [placeholder]="placeholder"
-            aria-label="Filter input">
+            [aria-label]="'Filter input'">
```
donna-marijne, line 13:
```suggestion
            aria-label="Filter input">
```
You can use the following technique to remove the attribute from the host wrapper:
https://stackblitz.com/edit/angular-ivy-luwfpc?file=src%2Fapp%2Fhello.component.ts,src%2Fapp%2Fapp.component.html

src/components/input-dropdown/input-dropdown.component.html
```diff
@@ -44,6 +44,7 @@
 
             <button type="button"
                     class="btn btn-flat filter-button"
+                    [attr.aria-label]="filter.length === 0 ? 'Search' : 'Close'"
```
donna-marijne, line 47: Unfortunately we can't hard code aria-label values because these are human facing and need to be localizable. (The time picker has hard coded aria-labels but this is a pre-existing fault.)

donna-marijne (changes requested):

src/components/date-time-picker/day-view/day-view.component.html
```diff
@@ -8,12 +8,13 @@
...
         <tr role="row" *ngFor="let row of dayService.grid$ | async; trackBy: trackWeekByFn">
 
-            <td *ngFor="let item of row; trackBy: trackDayByFn" class="date-cell" role="gridcell">
+            <td *ngFor="let item of row; trackBy: trackDayByFn" class="date-cell">
 
                 <button type="button"
                         uxFocusIndicator
                         uxFocusIndicatorOrigin
                         class="date-button"
+                        role="gridcell"
```
donna-marijne, line 17: I'm seeing more errors out of the axe tool with this placement, so I think the path of least resistance is to remove `role="grid"`, `role="row"`, and `role="gridcell"` altogether.

donna-marijne (changes requested):

donna-marijne (approved):
Looks good to me, thank you!

PR #1681, opened 2022-10-13
D3 Upgrade

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/issues/908

upgraded d3 to version 7
updated organisational chart
updated nested donut chart

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_issue-908-d3-upgrade

#### Downstream Build(s)

https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~issue-908-d3-upgrade~CI/


protractor.conf.js
```diff
@@ -28,7 +28,7 @@ const config = {
     baseUrl: `http://${e2eHostAddress}:4000/#/`,
     allScriptsTimeout: 11000,
     specs: [
-        './e2e/**/*.e2e-spec.ts'
+        './e2e/tests/components/organization-chart/*.e2e-spec.ts'
```
donna-marijne, line 31: I guess this was not intended to be committed.

donna-marijne (changes requested):
Thanks Jonathan, upgrade looks good. However this appears to be a breaking change, so it needs to be part of our next major version increment.

ashley-hunter:
> Thanks Jonathan, upgrade looks good. However this appears to be a breaking change, so it needs to be part of our next major version increment.

I don't think this is a breaking change, unlike most third party libraries we use, d3 is listed under `dependencies` and not `peerDependencies`, so we are in control of the version that our components use. The root application can install a different version of D3 should they wish but our components will still use 7.x.x. This could result in some code duplication in the applications bundle as it may be bundling two versions of some D3 functions, but it means it is not a breaking change.

donna-marijne:
Hmm, I'm seeing some changed callback signatures, presumably that's going to behave differently depending on which version of d3 the application has installed.
image

(Unless we do some detection and handle them differently depending on which version of d3 we are dealing with of course, which I believe is the plan.)

ashley-hunter:
> Hmm, I'm seeing some changed callback signatures, presumably that's going to behave differently depending on which version of d3 the application has installed.

That would be the case if D3 was listed as a peer dependency, but it is a dependency so it doesn't matter what version the application has installed (the application may not even have it installed).

If an application is currently using D3 v6 node_modules would look like this:

- node_modules
  + d3 (v6)
  + @ux-aspects/ux-aspects
      + d3 (v7)

If the application has d3 v7 installed or no version of D3 installed the node_modules would look like this:

- node_modules
  + d3 (v7)
  + @ux-aspects/ux-aspects

In either case, our components are going to use D3 v7 and the application will use whatever version is in the top level node_modules folder. So different signatures will have no effect on the application as it can use a different version of D3 than our components.

ashley-hunter:
> Hmm, I'm seeing some changed callback signatures, presumably that's going to behave differently depending on which version of d3 the application has installed.

That would be the case if D3 was listed as a peer dependency, but it is a dependency so it doesn't matter what version the application has installed (the application may not even have it installed).

If an application is currently using D3 v6 node_modules would look like this:

- node_modules
  + d3 (v6)
  + @ux-aspects/ux-aspects
      + d3 (v7)

If the application has d3 v7 installed or no version of D3 installed the node_modules would look like this:

- node_modules
  + d3 (v7)
  + @ux-aspects/ux-aspects

In either case, our components are going to use D3 v7 and the application will use whatever version is in the top level node_modules folder. So different signatures will have no effect on the application as it can use a different version of D3 than our components.

ashley-hunter:
> Never mind, I misread your comment. It's worth checking for sure. As long as d3 doesn't have any globals like some of those old libraries tend to.

Very true, worth checking!

JonathanToddMF:
@donna-marijne and @ashley-hunter had a look through v5/6 of D3 and there aren't any global assignments

JonathanToddMF:
Put together a quick example in codeSandBox looks good to me. 

Using d3 version 5 to create a simple bar chart below the organization chart - https://codesandbox.io/s/organization-chart-ux-aspects-micro-focus-forked-o1t5ns?file=/src/app/app.component.ts:5567-5572

donna-marijne (approved):
Looks good, thank you for checking.

PR #1683, opened 2022-10-31
fix(tabs) - add ngOnChanges 

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/issues/901

add on changes to tab component,
force tabset to rerender once changes detected

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_issue-901-tabs

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~issue-901-tabs~CI/)


ashley-hunter (changes requested):
Thanks, seems to work well. I had a suggestion on how we can simplify this, let me know if there are any issues.

src/components/tabset/tab/tab.component.ts
```diff
@@ -1,9 +1,11 @@
-import { ChangeDetectionStrategy, ChangeDetectorRef, Component, ContentChild, EventEmitter, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges, TemplateRef } from '@angular/core';
+import { ChangeDetectionStrategy, ChangeDetectorRef, Component, ContentChild, EventEmitter, Inject, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges, TemplateRef } from '@angular/core';
 import { NavigationExtras } from '@angular/router';
 import { Subject } from 'rxjs';
 import { distinctUntilChanged, takeUntil } from 'rxjs/operators';
+import type { TabsetComponent } from '../../..';
```
ashley-hunter, line 5: Can we use a path with a filename here, sometimes this can cause issues with the Angular compiler to default to index.ts

src/components/tabset/tabset.token.ts
```diff
@@ -0,0 +1,6 @@
+import { InjectionToken } from '@angular/core';
+
+/**
+ * This token is used to avoid circular dependency
+ */
+export const TabsetToken = new InjectionToken('Tabset');
```
ashley-hunter, line 6: This is something we should update in Universal, but injection tokens should ideally contain the type information of the thing it will provide. This is not so important when using constructor injection, but as of Angular 14 you can use an `inject` function instead so this provides type inference. E.g.

```ts
readonly tabset = inject(TabsetToken);
```

We will likely be changing our components to use this approach once we drop Angular 13 support as it allows us to more easily extend components as we don't need to forward all constructor parameters to parent component classes.

```suggestion
import type { TabsetComponent } from './tabset.component';

export const TabsetToken = new InjectionToken<TabsetComponent>('Tabset');
```

ashley-hunter (changes requested):
Thanks, looks great, glad that approach worked, just noticed two small things, thanks

ashley-hunter (approved):
Looks good to me, thanks

Roland-Penny (approved):
Accepted based on dev feedback.

PR #1686, opened 2022-11-23
Angular 15

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://github.houston.softwaregrp.net/caf/ux-aspects-micro-focus/issues/920

- Upgrade to Angular 15
- Update documentation examples to use `import.meta.webpackContext`
- Remove Angular 13 support
- Remove documentation references to Angular 13
- Update breaking changes
- move src/ files to use inject

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_angular-15

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~angular-15~CI/)


JonathanToddMF:
> Thanks, had a quick look and looks good! Will do a proper review later. Just wanted to check if we could update `angular-split` dependency. We are using v4 which is using view engine and the latest is v14, so we are rather out of date. I know the Apollo team were having some issues will ngcc last week and the only reason ngcc was being run was because of this dependency (which they weren't even using) so ideally we could update it and it would help them out also.

Thanks Ashley! Sorry forgot to add this to the PR. I have the angular split changes in a separate pr here - https://github.com/UXAspects/UXAspects/pull/1688

Just thought this PR was getting rather cumbersome! I aim to have it for review once these changes are merged.

ashley-hunter (changes requested):
Thanks, looks good! Had one or two minor suggestions.
I also noticed linting is failing.

Additionally, I notice that some injected properties are not readonly when they could be (likely an issue with my script 😜), but an easy way to mass fix this would be to add: "@typescript-eslint/prefer-readonly": "error", to our eslint config and then run `ng lint ux-aspects --fix` and that will add in readonly everywhere that safely can be be for you.

src/components/icon/icon.service.ts
```diff
@@ -8,6 +8,9 @@ import { uxIconset } from './iconsets/ux-iconset';
 
 @Injectable()
 export class IconService {
+    readonly options = inject<IconModuleOptions>(ICON_OPTIONS_TOKEN, { optional: true });
```
ashley-hunter, line 11: In this case we shouldn't need the generic here:

```suggestion
    readonly options = inject(ICON_OPTIONS_TOKEN, { optional: true });
```

The correct approach would be to update the token to specify the type:

```ts
export const ICON_OPTIONS_TOKEN = new InjectionToken<IconModuleOptions>('ICON_OPTIONS_TOKEN');
```

This should be done for all our tokens.

src/components/virtual-scroll/virtual-for/virtual-for.directive.ts
```diff
@@ -15,6 +15,17 @@ import { VirtualForRange, VirtualForService } from './virtual-for.service';
...
+
+    private readonly _templateRef = inject<TemplateRef<VirtualForOfContext<T>>>(TemplateRef);
+
+    private readonly _differs = inject(IterableDiffers);
+
+    private readonly _renderer = inject(Renderer2);
+
+    private readonly _changeDetector = inject(ChangeDetectorRef);
+
+    private readonly _virtualScroll = inject<VirtualForService<T>>(VirtualForService);
```
ashley-hunter, line 28: This was previously optional

ashley-hunter (changes requested):
Thanks, just noticed a dependency that used to be optionally that is now no longer, and noticed we should update our injection tokens to specify their types rather than relying on casting the result from the `inject` function

ashley-hunter (dismissed):
Looks good as far as I can see, thanks!

JonathanToddMF:
Just to note the marquee wizard codeSandBox will be fixed in this pr - https://github.com/UXAspects/UXAspects/pull/1688

src/directives/accessibility/focus-indicator/focus-indicator.directive.ts
```diff
@@ -11,7 +11,6 @@ import { FocusIndicatorService } from './focus-indicator.service';
 @Directive({
     selector: '[uxFocusIndicator]',
     exportAs: 'ux-focus-indicator',
-    providers: [LocalFocusIndicatorOptions]
```
ashley-hunter: Can I just clarify why this was removed? Is it not needed any more?

JonathanToddMF: Yeah I had incorrectly added this in a previous commit - https://github.com/UXAspects/UXAspects/pull/1686/commits/f2e758e581ad7a60a41ce8c59a0113d6a518c320

ashley-hunter: Perfect, thank you just wanted to make sure this wasn't unintentional

ashley-hunter (approved):
Looks good, thanks

PR #1688, opened 2023-01-09
Upgrade angular split + ng2-file-upload

JonathanToddMF:
#### Checklist

#### Ticket / Issue

n/a

TODO: Merge after Angular 15
- Update angular split to 14
- Updated screenshot that was incorrectly showing the value as 25 when the correct value is 100
- Updated tests to account for new arrow movement. Right/Left keys move the splitter more than it did previously. Up/Down keys move the splitter the same as before. As such, I have updated the previous tests to use Up/Down and added additional coverage for the new left/right movement.
- Upgraded ng2-file-upload to 3.0.0
- Removed view engine transformer

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_upgrade-angular-split

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~upgrade-angular-split~CI/)


docs/app/services/playground/transformers/view-engine-downgrade.ts
```diff
@@ -26,7 +26,6 @@ function isViewEngineModule(playgroundModule: IPlaygroundModule): boolean {
         ? playgroundModule.imports
         : [playgroundModule.imports];
     return (
-        playgroundModule.library === 'angular-split' ||
         playgroundModule.library === 'ng2-file-upload' ||
         imports.findIndex(_import => _import === 'MarqueeWizardModule') >= 0
```
ashley-hunter, line 30: Would it be possible to update ng2-file-upload as well if it is a view engine library? I think that should be fairly trivial as its just an example, not a dependency or peer dependency of our library.

And I assume MarqueeWizardModule is only here because it relies on angular split?

If we are able to update ng2-file-upload then I believe we can remove the whole view engine transformer all together.

ashley-hunter (changes requested):
Thanks, the angular-split update looks good, but I think we can remove our reliance on View Engine all together fairly easily, probably good to do that if possible, thanks.

ashley-hunter (approved):

ashley-hunter (approved):
Looks good to me, thank you

docs/app/pages/components/components-sections/file-upload/file-upload/file-upload.component.ts
```diff
@@ -48,5 +48,7 @@ export class ComponentsFileUploadComponent
         );
 
         this.uploader.onCompleteAll = () => announcer.announce('All files have been uploaded.');
+
+        // this.uploader.on
```
ashley-hunter, line 52: Think this was accidentally committed

ashley-hunter (approved):
Looks good to me, thanks

PR #1692, opened 2023-02-07
412134 - Accessibility fixes

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=412134

Navigation
- removed aria-expanded when there is no children in the nav

Pagination
- Moved pagination to use buttons instead of anchor tags
- Added some additional styling to bring buttons inline with anchor tags
- Updated tests to use button instead of anchor

Column picker
- Added new input for aria-label
- Moved aria-selected to treeitem
- Updated selection item to allow disabling of aria selected

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_412134-accessibility

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~412134-accessibility~CI/)


JonathanToddMF:
Pagination 
https://cw0g7v.csb.app/

Navigation
https://6tr0zs.csb.app/

Column Picker 
https://g15x32.csb.app/

src/components/table/column-picker/column-picker.component.html
```diff
@@ -22,19 +22,22 @@
...
-                [ngClass]="'column-picker-tree-node-level-' + node.level">
-
-                <ng-container *ngIf="!deselectedTemplate">{{ node.name }}</ng-container>
-
-                <ng-container
-                    *ngIf="deselectedTemplate"
-                    [ngTemplateOutlet]="deselectedTemplate"
-                    [ngTemplateOutletContext]="{ $implicit: node.name }">
-                </ng-container>
+            <div role="group">
```
donna-marijne, line 25: This group will only ever contain one item. Since this is set up as a flat tree, which you can observe by looking at the DOM:
image

you are not able to use `role="group"` as intended. With a screen reader, every item will be `item one of one`.

You also added `role="treeitem"` which duplicates the role of the parent element.

I suspect that the fix is to move `aria-selected` to the cdk-tree-node element.

src/components/table/column-picker/column-picker.component.html
```diff
@@ -120,7 +123,7 @@
         </ng-container>
     </div>
 
-    <div class="column-picker-list">
+    <div class="column-picker-list" role="listbox" [attr.aria-label]="columnPickerListAriaLabel">
```
donna-marijne, line 126: Why does the right hand column need a label when the left hand column doesn't have one? Surely both or neither would make most sense.

Also there is descriptive text already available at the top of each column:
image

You can link existing text in the UI as a label for another element using `aria-labelledby`.

src/components/table/column-picker/column-picker.component.ts
```diff
@@ -69,6 +69,9 @@ export class ColumnPickerComponent implements OnChanges {
     /** Define a comparator function used for sorting the deselected columns. */
     @Input() sort: (a: ColumnPickerGroupItem, b: ColumnPickerGroupItem) => number;
 
+    /** Define a aria label for the column picker list. */
+    @Input() columnPickerListAriaLabel: string = 'Column Picker aria label';
```
donna-marijne, line 73: I wouldn't blame a blind person for wanting to throw their headphones out the window if they are trying to navigate an interface and hear the phrase 'Column Picker aria label.' ARIA attributes are as much a part of the user interface as any widget you click on, and usability needs to be considered just as carefully.

donna-marijne (changes requested):
Thank you for the PR. The axe tool is happy, but I have some feedback.

This remark from the reporter:
> aria-label attribute cannot be used on a a with no valid role attribute.

is actually nonsense. I know you already did the work, but in future it's a good idea to reality check the output of these tools.

Other feedback below. If you are in doubt about anything, just turn on a screen reader and see how it sounds.

docs/app/pages/components/components-sections/tables/selection/selection.component.html
```diff
@@ -151,6 +151,9 @@
     <tr uxd-api-property name="selected" type="boolean">
         Defines whether or not this item is currently selected.
     </tr>
+    <tr uxd-api-property name="preventAriaSelected" type="boolean" defaultValue="false">
+        Defines whether or not aria-selected is added to the host element.
```
donna-marijne, line 155: 1. Use monospace for attribute names.
2. "Defines" is redundant here.
3. "Whether" implies the possibility of "not."
```suggestion
        Whether <code>aria-selected</code> is added to the host element.

```

src/directives/selection/selection-item.directive.ts
```diff
@@ -52,6 +54,9 @@ export class SelectionItemDirective<T> implements OnInit, OnChanges, OnDestroy {
         this._isDisabled = isDisabled;
     }
 
+    /** Defines whether or not aria-selected is added to the host element. */
+    @Input() preventAriaSelected: boolean = false;
```
donna-marijne, line 58: Coerce boolean property values.

donna-marijne (changes requested):
Thanks for the update. A few pretty basic mistakes here.

donna-marijne (approved):
Changes look good, thank you.

PR #1713, opened 2023-07-27
979 - Page Header accessibility

JonathanToddMF:
#### Checklist

#### Ticket / Issue

https://internal.almoctane.com/ui/entity-navigation?p=131002/7002&entityType=work_item&id=979

Removed role="menubar" from the page header navigation when there are no child menuitems

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_979-page-header-accessibility

#### Downstream Build(s)

* [caf/ux-aspects-micro-focus](https://jenkins.swinfra.net/job/SEPG/job/caf/job/caf~ux-aspects-micro-focus~979-page-header-accessibility~CI/)

### DIFF
--- a/src/components/page-header/navigation/navigation.component.ts
+++ b/src/components/page-header/navigation/navigation.component.ts
@@ -1,4 +1,4 @@
-import { AfterViewInit, Component, ElementRef, inject, OnDestroy, QueryList, ViewChildren } from '@angular/core';
+import { AfterViewInit, ChangeDetectorRef, Component, ElementRef, inject, OnDestroy, QueryList, ViewChildren } from '@angular/core';
 import { NavigationExtras } from '@angular/router';
 import { BehaviorSubject, Subject } from 'rxjs';
 import { distinctUntilChanged, takeUntil } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


src/components/page-header/navigation/navigation.component.ts
```diff
@@ -33,6 +35,11 @@ export class PageHeaderNavigationComponent implements AfterViewInit, OnDestroy {
 
     private readonly _onDestroy = new Subject();
 
+    get role(): string {
+        this._changeDetectorRef.detectChanges();
```
ashley-hunter, line 39: We shouldn't run change detection inside a getter, if we think about what change detection does, it checks bindings in a template (and host bindings). 

In this case, the host binding calls the role getter to get the value and check if it has changed, which will trigger change detection, which checks the binding, which triggers change detection etc... and so on..

If we need to trigger change detection, it should be done in response to something, in this case it is in response to the menuItems changing, so we could so:

```ts
this.menuItems.changes.pipe(takeUntil(...)).subscribe(() => this._changeDetectorRef.detectChanges());
```

ashley-hunter (changes requested):
Thanks, just noticed one issue which could potentially lead to an infinite loop

ashley-hunter (approved):
All good, thank you

PR #1725, opened 2024-01-30
fix: partition map performance issues

ashley-hunter:
#### Checklist

#### Ticket / Issue

N/A - reported by Andrew Jolly, but no ticket raised

When displaying large numbers of items in the partition map, performance can become incredibly slow. There were two key issues I discovered:

1. We recompute the x position and width of segments many times as this runs on change detection and it is calculated multiple times within each change detection cycle. These calculations are reasonably expensive, so I have added a caching mechanism to only calculate the values once, until they may have changed giving a large speed boost.
2. The tooltip (which is often used in conjunction with the partition map) when `hide` is called it will run change detection. This happens regardless of whether the tooltip is actually open or not. Adding a check to only run the hide logic if the tooltip is actually visible.

#### Documentation CI URL

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_performance-improvements

https://pages.github.houston.softwaregrp.net/sepg-docs-qa/UXAspects_CI_UXAspects_performance-improvements/#/charts/partition-map#partition-map

#### Downstream Build(s)

https://jenkins.swinfra.net/job/SEPG/job/caf/view/Developer/job/caf%7Eux-aspects-micro-focus%7Eperformance-improvements%7ECI/

### DIFF
--- a/src/components/partition-map/partition-map.component.ts
+++ b/src/components/partition-map/partition-map.component.ts
@@ -7,6 +7,11 @@ import { ContrastService, FocusIndicatorOriginService } from '../../directives/a
 import { ResizeService } from '../../directives/resize/index';
 import { ColorService, ThemeColor } from '../../services/color/index';
 
+interface SegmentCacheData {
+    x?: number;
+    width?: number;
+}
+
 @Component({
     selector: 'ux-partition-map',
     templateUrl: './partition-map.component.html',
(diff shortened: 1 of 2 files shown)


DeanHaire:
Thanks Ashley, I've tested this on a large dataset and it works perfectly now 👍

JonathanToddMF (approved):

ashley-hunter:
@DeanHaire thank you very much for the feedback! Glad that resolved the issue!

